PR #492, opened 2015-01-19
Change exec to spawn Fix #486

### LINKED ISSUE #486: code injection vulnerability in ungit
screen shot 2015-01-04 at 2 44 58 pm

Let's assume an user enters inputs such as above on an ungit server that is hosted on unix flavored OS.  Ungit will run the below command ultimately.

`git -c color.ui=false -c core.quotepath=false -c core.pager=cat remote add test6 github.com/codingtwinky/ungit && rm package.json`

Although above command is rather harmless but I'm sure you can see that this is a classic code injection vulnerability that can be potentially critical.

Will be adding a method to sanitize user inputs via checking for & char but should I be checking for any other characters?

codingtwinky:
As discussed in #486, child_process.exec has a remote code injection vulnerability so it was replaced with spawn.  

This was the most horrifying fun I ever had as it broke almost every single tests on initial solution which caused by several subtle differences between spawn and exec.  In fact it is possible that some of the newer or untested feature may have bugs so I think further testing would be wise.

Notes:
* spawn give you stdout and stderr as stream so call back may get called multiple times whereas exec's callback will get called only once when it is done.  Solved via q library.
* Just splitting git commands via `.split(' ');` obviously cause problems so each individual git calls were evaluated.
* Git really didn't like unnecessary double quote wrapped around some of the arguments.
* No tests script were altered, except where pure `/git` api that is used in some tests as it needed to be changed to array argument not string argument.


source/git-api.js
```diff
@@ -128,7 +128,11 @@ exports.registerApi = function(env) {
   });
 
   app.post(exports.pathPrefix + '/init', ensureAuthenticated, ensurePathExists, function(req, res) {
-    git('init' + (req.param('bare') ? ' --bare --shared' : ''), req.param('path'))
+    var arg = ['init'];
+    if (req.param('bare')) {
+      arg = arg.concat(['--bare', '--shared']);
```
campersau: You can just use `arg.push('--bare', '--shared');`. No need to create two new arrays.

codingtwinky: Thanks!

Ajedi32:
Wow, that required way more work than I was expecting it to. Good job! :+1:

FredrikNoren:
Wow! This is awesome!

Only little comment I've got is maybe we should use bluebird instead of Q? It seems to be getting a bit more traction than Q these days, but has got a very similar interface. It's just a suggestion though, if you want to stick with Q I'm fine with that.

codingtwinky:
Bluebird is a really cool library, `Promise.promisifyAll(require("fs"));` is a really freaking cewl concept that I would have to dig the source code later.  I like it better then Q based on my first glance and I thank you for suggestion!

But honestly, I think the usage of promise library for this is little over jealous.  It's a very small and minor case and I don't know if I can justify adding an additional dependency for something that can be easily done as this.

FredrikNoren:
@codingtwinky Makes sense to me, merging this now. Awesome work once again! I think this is a great step towards making ungit safer on hosted environments.

PR #493, opened 2015-01-24
Dependency bump

codingtwinky:
This upgrades all outstanding outdated dependencies except for socket.io.

Note:
* express's `param()` function is now deprecated, which we use it all over the places especially in git-api.  So both runtime and test log output will be noisy with all the deprecated warnings.  Will be addressing this issue later maybe in this pull.
* I think it's super test that now streams content of file when the file type is binary instead of sending the text.  Which we use to test image file diffs.  Test cases are updated to adjust to this issue.


FredrikNoren:
@codingtwinky Just to make sure; are you aware of bumpdependencies in the grunt file? It's a little script I wrote to help automatically bump dependencies. It's also holding some of them back because of different issues (detailed in the Gruntfiles.js). See https://github.com/FredrikNoren/ungit/blob/master/Gruntfile.js#L332 for more details

codingtwinky:
Had no idea about that handy function.

* forever monitor - I did the quick test with a script that is under a directory with space and it worked fine, I think it was fixed.
* imagemin - I'm not sure what the issue is but it seems to be working...
* supertest - issue was that when file type is set as binary, super test no longer read stream as a text anymore.  It output text as byte stream and it is fixed in the new script.

FredrikNoren:
@codingtwinky Sweet!

Issue #432, opened 2014-08-27
Really slow to handles lot of new filesto commit (~4000)


eucalyp:
I changed the name of a wordpress folder. So it's about 4000 files to commit and ungit is really slow (2mn to generate the diff) and even crash to create the diff or to commit (once the diff succeed). I don't know what is the bottleneck. For information, I have a ssd installed on my computer, so the hard drive is not the problem.

campersau:
Should have been closed by https://github.com/FredrikNoren/ungit/pull/508

PR #497, opened 2015-02-08
Rolling load when too many upstaged files. fix #485

### LINKED ISSUE #485: Ungit hanging when handling a lot of (unadded) files
This maybe overlapping with Issue [#432] ("https://github.com/FredrikNoren/ungit/issues/432) but more specifically it's about handling  *unadded* files.

First I added a whole magento folder (~13.000) files, which instantly led to hanging (on OSX 10.9.5 + Chromium 38.0.2125.122 / Firefox 35.0).
I then started cutting down on added files, but even if I only add a single file, ungit keeps hanging until crash.

I then reinitialised the repo in a subfolder with only a few files, which works fine - so I assume it must be the unadded files causing the hang.

Is there any possibility keeping ungit from scanning those unadded files? (I'm gonna have a look myself - any chance you could point me to the file(s) where this could be happening?)

Love the project, keep it up!

codingtwinky:
This fixes #485 and #432.

(Might need to change the name of gitErrors module to notification or something more appropriate since it's used for more then git error notification now.)

### DIFF
--- a/components/gitErrors/gitErrors.html
+++ b/components/gitErrors/gitErrors.html
@@ -10,13 +10,21 @@
   <p data-bind="visible: showEnableBugtracking">
     <button class="btn btn-primary" data-bind="click: enableBugtrackingAndStatistics">Enable bugtracking + usage statistics</button> to automatically report bugs, and/or <a href="https://github.com/FredrikNoren/ungit/issues" target="_blank">file an issue on the ungit issue tracker</a>.
   </p>
-  <h4>Command</h4>
-  <pre data-bind="text: command"></pre>
-  <h4>Error</h4>
-  <pre data-bind="text: error"></pre>
-  <h4>Stderr</h4>
-  <pre data-bind="text: stderr"></pre>
-  <h4>Stdout</h4>
-  <pre data-bind="text: stdout"></pre>
+  <div data-bind="visible: command">
+    <h4>Command</h4>
+    <pre data-bind="text: command"></pre>
+  </div>
+  <div data-bind="visible: error">
+    <h4>Error</h4>
+    <pre data-bind="text: error"></pre>
+  </div>
+  <div data-bind="visible: stderr">
+    <h4>Stderr</h4>
+    <pre data-bind="text: stderr"></pre>
+  </div>
+  <div data-bind="visible: stdout">
+    <h4>Stdout</h4>
+    <pre data-bind="text: stdout"></pre>
+  </div>
 </div>
 <!-- /ko -->
--- a/components/gitErrors/gitErrors.html
+++ b/components/gitErrors/gitErrors.html
@@ -1,29 +1,22 @@
 
 <!-- ko foreach: gitErrors -->
-<div class="alert" data-bind="css: {'alert-danger': type() === 'error', 'alert-warning': type() === 'warning'}">
-  <button type="button" class="close" data-bind="click: dismiss, visible: closeable">&times;</button>
-  <h3><span class="glyphicon glyphicon-warning-sign"></span><span data-bind="text: title"></span></h3>
-    <span data-bind="visible: tip, text: tip"> </span><span data-bind="visible: bugReportWasSent">An automatic bug report was sent.</span>
+<div class="alert alert-danger">
+  <button type="button" class="close" data-bind="click: dismiss">&times;</button>
+  <h3><span class="glyphicon glyphicon-warning-sign"></span> Unhandled git error!</h3>
+  <p>
+    Ungit tried to run a git command that resulted in an unhandled error. <span data-bind="visible: bugReportWasSent">An automatic bug report was sent.</span>
   </p>
+  <p data-bind="visible: tip, text: tip"></p>
   <p data-bind="visible: showEnableBugtracking">
     <button class="btn btn-primary" data-bind="click: enableBugtrackingAndStatistics">Enable bugtracking + usage statistics</button> to automatically report bugs, and/or <a href="https://github.com/FredrikNoren/ungit/issues" target="_blank">file an issue on the ungit issue tracker</a>.
   </p>
-  <br/>
-  <div data-bind="visible: command">
-    <h4>Command</h4>
-    <pre data-bind="text: command"></pre>
-  </div>
-  <div data-bind="visible: error">
-    <h4>Error</h4>
-    <pre data-bind="text: error"></pre>
-  </div>
-  <div data-bind="visible: stderr">
-    <h4>Stderr</h4>
-    <pre data-bind="text: stderr"></pre>
-  </div>
-  <div data-bind="visible: stdout">
-    <h4>Stdout</h4>
-    <pre data-bind="text: stdout"></pre>
-  </div>
+  <h4>Command</h4>
+  <pre data-bind="text: command"></pre>
+  <h4>Error</h4>
+  <pre data-bind="text: error"></pre>
+  <h4>Stderr</h4>
+  <pre data-bind="text: stderr"></pre>
+  <h4>Stdout</h4>
+  <pre data-bind="text: stdout"></pre>
 </div>
 <!-- /ko -->
--- a/components/gitErrors/gitErrors.html
+++ b/components/gitErrors/gitErrors.html
@@ -1,15 +1,14 @@
 
 <!-- ko foreach: gitErrors -->
-<div class="alert alert-danger">
-  <button type="button" class="close" data-bind="click: dismiss">&times;</button>
-  <h3><span class="glyphicon glyphicon-warning-sign"></span> Unhandled git error!</h3>
-  <p>
-    Ungit tried to run a git command that resulted in an unhandled error. <span data-bind="visible: bugReportWasSent">An automatic bug report was sent.</span>
+<div class="alert" data-bind="css: {'alert-danger': type() === 'error', 'alert-warning': type() === 'warning'}">
+  <button type="button" class="close" data-bind="click: dismiss, visible: closeable">&times;</button>
+  <h3><span class="glyphicon glyphicon-warning-sign"></span><span data-bind="text: title"></span></h3>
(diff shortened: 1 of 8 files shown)


codingtwinky:
screen shot 2015-02-06 at 6 58 12 pm

FredrikNoren:
@codingtwinky hm why show the warning at all? If this adds a "load more" button isn't that enough?

codingtwinky:
As a daytime server side grunt, anytime I have to find something informative I would have to dig through lines and lines of logs.  So when informative warning such as this is displayed nicely, it makes me happy...
I agree that with rolling load warning maybe an overkill but at the same time relying solely on "load more" button maybe too subtle.  

I can take it out and revert the gitErrors component change later.

FredrikNoren:
@codingtwinky I see. Well, if the load more button is too subtle we can always make it a little less subtle (put a box around it with a different background or something). Also I think that if we do want warning dialogs like this it's probably better to implement them as new dialogs (just like how the other warning dialogs, such as "new version" and "enable bugtracking", are implemented) since the git-error dialog is designed specifically for git errors.

Ajedi32:
I tend to agree. A bit, obtrusive warning message isn't necessary for something which really shouldn't require any action or thought on the user's part. Just put a large, obvious "load more" button at the end of the list, and ensure that buttons like "select all" work as expected even when not all the files are displayed. The user shouldn't have to care about whether some of the changes are hidden for performance reasons.

codingtwinky:
Updated below

screen shot 2015-02-13 at 4 50 31 pm

Ajedi32:
That would do it. I still like the idea of putting the button at the end of the list a bit better (assuming that the check all button still "checks" items which aren't loaded), but this will work fine too. It's sufficiently subtle, but not so subtle that the user wouldn't realize that it's there.

FredrikNoren:
I agree with @Ajedi32 that it would probably make more sense to put it at the bottom. I think it's what you would expect (you scroll to the bottom to find what you are looking for and once there you click a button to show some more) and it's also how it works in the rest of the UI right?

codingtwinky:
Sorry missed that, moved the load more button to bottom of the staging.

Also, few things to mention:

* "Discard All" will discard all even not displayed in staging.
* "Stash All" will stash all even not displayed in staging.
* "Toggle All" will only toggle displayed files, which might need to update the wording...

screen shot 2015-02-16 at 12 41 33 pm

Ajedi32:
> "Toggle All" will only toggle displayed files, which might need to update the wording...

Yeah, I figured that might be an issue, which is why I said

> I still like the idea of putting the button at the end of the list a bit better **(assuming that the check all button still "checks" items which aren't loaded)**

I like putting the button at the end of the list, because normally the user shouldn't have to care about whether or not all the files are loaded unless the user needs to view those files (at which point he would already have scrolled to the bottom of the list).

The problem though is that if the Toggle All button doesn't check items which aren't loaded, then the user actually *does* have to care about whether all files are loaded. If we're not going to fix that problem at the root yet, then we need to warn the user about that behavior somehow, since we're doing something which violates the principle of least surprise.

On a related note, is that indicator at the top "100 files, 100 to be committed" actually accurate with this implementation? We wouldn't want to give the user the impression that clicking "commit" will commit everything if that's actually not the case.

Ajedi32:
For the "Toggle All" button specifically, I'd recommend changing it to "Toggle Visible" when there are some files which aren't loaded yet. (But leaving it as "Toggle All" when all files are visible.)

codingtwinky:
"100 files, 100 to be committed" is accurate in a sense that it 100 is displayed and 100 will be committed.  Now this is inaccurate in a sense that there are more then 100 staged files that are not showing.

I personally like the load more button at the top so user can immediately know that there are more files then what is visible without scrolling down.

I see these options below, but feel free to add more options.

1. Add load more at the top and leave other functionality as is but fix "toggle all" to "toggle visible" when there are outstanding files.
2. Add load more at the bottom and have more accurate staged file count.  so it would show "43234 files 100 to be committed"
3. Add a separate commit all button that would commit what is not visible as well.  I'm saying this because changing "toggle all" to include what is not visible is some what deceiving to me.

I get the impression that @Ajedi32 is for option 1, now that toggle all is not checking for non visible files, which is what I like as well,.

Ajedi32:
Actually, I'd say all three of those options combined would probably be best. Or at least, they'd be the "next best thing" to actually fixing the toggle all button so that it toggles files that aren't loaded, in addition to the ones that are.

## If we don't fix the toggle all button immediately

* Keeping the button on the top is good, since it makes it more obvious to the user that not all files are loaded, and that actually is something the user has to care about, since clicking "commit" won't actually commit everything, and "Toggle All" only toggles visible items.
* Switching "toggle all" to "toggle visible" when there are outstanding files is good, since I think that "All" would mislead users into believing that they can check files which aren't visible in the UI. My only concern is that "Toggle Visible" sounds a bit like it's toggling the visibility of something, when it's really only toggling the checkbox of visible items. Not sure if there's a better way of wording that.
* Adding a load more button at the bottom is good, since that's where the user will usually be looking anyway when he wants to see more files
* A more accurate staged file count is good. "100 files" to me implies that there are only 100 uncommitted files total, not only 100 uncommitted files currently visible in the UI
* A separate commit all button would be good, since I suspect that's probably going to be the typical use case the user will follow when he has a large number of uncommitted files in the working directory. In the absence of such a button, a user wanting to commit everything would have to either switch to the command line, load all the files into the view before committing, or use multiple commits to commit the files. None of those options sound optimal to me.

## If we do fix the toggle all button immediately

* Keeping the button on the top is unnecessary. The user doesn't need to know that some files aren't displayed unless he actually needs to see those files, at which point he'll be scrolled to the bottom of the page, not the top.
* "Toggle all" would actually toggle all items, so no need to change the wording to "toggle visible"
* Adding a load more button at the bottom is still good, since, as I said, that's where the user will usually be looking anyway when he wants to see more files.
* A more accurate staged file count is still a good idea, for the same reasons as what I listed in the section above.
* A separate commit all button would be unnecessary. That would be the default behavior of the commit button

Ajedi32:
@codingtwinky 

> I'm saying this because changing "toggle all" to include what is not visible is some what deceiving to me.

How so? That seems like the ideal option in my opinion, assuming it's feasible to implement. The fact that not all items are displayed is just due to performance restrictions on the view, so IMO those restrictions should affect the UX as little as possible.

The way I see it, the user shouldn't have to know or care whether all items are displayed or not unless he wants to scroll down to the bottom and check, uncheck, or view those items. At that point, a "Load More" button at the bottom would let the user know that not all files are visible, and give him the option to view them if he wishes. Until the user sees that button though, the user shouldn't have to know or care that not all files are visible. Making the toggle all button select invisible items makes that possible, whereas making the toggle all button only select visible items forces the user to care about what items are visible, even if he doesn't want to interact with any of those invisible items individually.

codingtwinky:
First bullet for "if we don't fix toggle all" says "keeping the button at the top is good" but third bullet says adding a load more button at the bottom is good.  You mean to have two different buttons for "load more"?

And if we do fix toggle all button how are we going to indicate the state where we want to select all visible only?  Is user supposed to un-toggle all and individually select each visible files?  If we were to do that ungit view would be visually indistinguishable between "all selected" and "all visible selected".  Or are you envisioning a dummy file entry to represent all invisible files?


Only reason why I said it is because to me I focused on what is visible and my working scope is what is visible, which is just my preference.  I'm more of the fan where users needs to be informed about all items are displayed or not.  Because I would like to know scope of my actions explicitly, especially in dev tools.  Perhaps this maybe why I don't work in UX team.   I'm not as passionate about this subject, so I will fetch a prototype based on the input and hopefully I will get it right this time.

Ajedi32:
> First bullet for "if we don't fix toggle all" says "keeping the button at the top is good" but third bullet says adding a load more button at the bottom is good. You mean to have two different buttons for "load more"?

Yes that's correct. The button on top makes it easier for the user to see up front that not all the items are loaded, and the button on the bottom makes it possible for the user to load more entries after he has already scrolled down.

> And if we do fix toggle all button how are we going to indicate the state where we want to select all visible only? Is user supposed to un-toggle all and individually select each visible files? 

That seems fine to me, after all isn't that what the user would have to do anyway if all of the files were visible?

> If we were to do that ungit view would be visually indistinguishable between "all selected" and "all visible selected". Or are you envisioning a dummy file entry to represent all invisible files?

Hmm, that actually sounds like a really good idea. You could have a dummy entry that says "and 55 more files" or something like that. Then checking that file would select all the invisible files, and clicking that file would load more files into the view.

FredrikNoren:
Hm this is becoming incredibly complex for something which is arguably kind of an edge case (you don't regularly add a huge number of files to repositories). Not only is there a lot of "cost" (in the sense that we can spend the time doing something that will have larger impact for our users) building it right now, but also maintaining it in the future.

For that reason, and since I know Ungit handles a _fairly_ large amount of files (I think I've seen at least hundreds on the screen with no problem), I would suggest a completely different solution:

When there are more than a certain number of files, say 1000, hide all the normal controls in the staging area and just show a message that says "Ungit will perform badly with this many files, it's recommended to use the command line instead. Here's how to do it using the command line: XXXX" and a single button "Load all files anyway" which would just load all the files even though it would probably destroy the performance.

What do you guys think?

codingtwinky:
That sounds similar to what I had very first, with big warning sign except it was with rolling load and functional staging.  I do agree that this new solution will be much simpler and easier to maintain for a very rare case.  

Will be doing below:
* up the loading limit to 1000
* remove rolling load
* Display warning sign when over the limit with two options (I'm thinking dialog with modal-backdrop)
  * "Don't load" - navigate user back to ungit home page
  * "load anyway" - Load any way

Ajedi32:
Yeah, if you think this is going to be a problem to maintain, then by all means go with a plain warning message; it's not often that I'd have to commit 1000+ files with Ungit.

Might want to open a separate pull request for that and close this one, since that seems like an altogether different fix from what is described here.

FredrikNoren:
+1 to a separate PR if possible, it should be a pretty small change

PR #501, opened 2015-02-14
Refactor clicktests into several files

FredrikNoren:
cc @codingtwinky 

Clicktests in several files, and it's possible to only run one of the suites by running `./node_modules/.bin/phantomjs clicktests/test.generic.js`.


FredrikNoren:
@anbuf you might be interested in this too, as it should be a little easier writing clicktests with this refactoring

PR #502, opened 2015-02-15
Fix refresh button

FredrikNoren:
cc @codingtwinky Seems like the refresh button disappeared in a recent patch, restoring it with this fix.


codingtwinky:
Yes, I noticed that and was hoping to send you the pull request before you noticed it...  :P

FredrikNoren:
Hehe, too slow ;) I hope that with the new click tests infrastructure will
make it easier to write tests and thus catch things like this earlier (in
the PR preferably)

codingtwinky:
Yes will be adding the click test right now!

PR #503, opened 2015-02-15
Moving refresh button out of staging for no header mode

codingtwinky:
Extension of #502.  Moving refresh button out of staging for no header mode so one can refresh even when there are no files in staging.

Regular: 
screen shot 2015-02-15 at 3 12 11 pm

No header mode:
screen shot 2015-02-15 at 3 06 59 pm

### DIFF
--- a/clicktests/test.all.js
+++ b/clicktests/test.all.js
@@ -5,3 +5,4 @@ require('./test.remotes');
 require('./test.screens');
 require('./test.submodules');
 require('./test.generic');
+require('./test.noheader');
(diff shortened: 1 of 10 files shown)


codingtwinky:
hmmm, failed on the submodule init clicktets and I will be looking at it later...

FredrikNoren:
I would recommend setting up a "noheader" clicktest file for this, which would launch ungit in the noheader mode and make sure things like this works there.

FredrikNoren:
Nice!

PR #505, opened 2015-02-18
Using observable array

codingtwinky:
Because it's new, efficient and more feature rich!

http://knockoutjs.com/documentation/observableArrays.html

### DIFF
--- a/components/app/app.js
+++ b/components/app/app.js
@@ -16,7 +16,7 @@ var AppViewModel = function(appContainer, server) {
     this.header = components.create('header', { app: this });
   this.dialog = ko.observable(null);
 
-  this.repoList = ko.observable(JSON.parse(localStorage.getItem('repositories') || localStorage.getItem('visitedRepositories') || '[]')); // visitedRepositories is legacy, remove in the next version
+  this.repoList = ko.observableArray(JSON.parse(localStorage.getItem('repositories') || localStorage.getItem('visitedRepositories') || '[]')); // visitedRepositories is legacy, remove in the next version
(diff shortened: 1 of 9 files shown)


FredrikNoren:
Well, there's actually a reason it's not using observable arrays in many places :)

The reason is that if you set up an observable array, and then push items into it one by one, the UI updates for each of them (at least this was the behaviour of knockout when I wrote Ungit), so in those cases it's actually much better to just use an observable and then put a list in it, so that it's a single UI update.

Now I don't know if knockout has changed since then, but I'd be hesitant to change this unless we're fairly certain we won't get trouble with this (i.e. the knockout implementation has changed) or there's some kind of benefit to changing.

codingtwinky:
Hmm...  I didn't know it was intentional.  I just thought it was a new feature that was added to ko after ungit.  I just thought it would be better because it is explicitly for observableArray and ungit usually doesn't manipulate each array elements, rather it swaps out entire array.

Should I do something like below or is this too much?

```
// Ensure it notifies about changes no more than once per 50-millisecond period
myViewModel.myObservableArray.extend({ rateLimit: 50 });
```

FredrikNoren:
Since Ungit usually swaps the whole array I don't think there's a strong
reason to change it . The rateLimit is just making it even more complicated
imo (and you might get weird states from that). If there's a case where
something can be gained by switching I'd be open to changing in that case
but otherwise let's leave it like it is now

codingtwinky:
I agree, I do think observable array is applicable in these remaining locations though.  Let me know if you still think this is unfitting, I will close it.

PR #508, opened 2015-02-24
Show warning message when too many staging files fix: #485

### LINKED ISSUE #485: Ungit hanging when handling a lot of (unadded) files

codingtwinky:
This fixes #485 and #432.

(Might need to change the name of gitErrors module to notification or something more appropriate since it's used for more then git error notification now.)

screen shot 2015-02-23 at 7 15 55 pm

### DIFF
--- a/components/dialogs/dialogs.js
+++ b/components/dialogs/dialogs.js
@@ -120,8 +120,8 @@ function TooManyFilesDialogViewModel(title, details) {
   this.taDialogName('yes-no-dialog');
   this.result = ko.observable(false);
   this.alternatives([
-    { label: "Don't load", primary: true, taId: 'noLoad', click: function() { self.result(true); self.close(); } },
-    { label: 'Load anyway', primary: false, taId: 'load', click: function() { self.result(false); self.close(); } },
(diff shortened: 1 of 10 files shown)


FredrikNoren:
Cool! There's a merge conflict block me from merging it though.

I'm also wondering if it would be possible to add a test for this? This is definitely the type of thing that will just creep back after a while if we don't have a test that ensures it doesn't

PR #510, opened 2015-02-27
Updating dep bump script

codingtwinky:


FredrikNoren:
Do we need to bump these dependencies? If not I'd rather keep them where they are since we know that it works as it is

codingtwinky:
As a person who is so fascinated about fast evolving world of js, I just love bumping up dependencies and how easy it is.  

So I guess I got carried away...

PR #514, opened 2015-03-08
Diff2html

codingtwinky:
With this pull request, ungit now solely uses diff2html for both normal diff and side by side diff view.

Few things to highlight

* Rolling load per diff, #376, is removed due to parser change.
* The way new and deleted file handled currently doesn't result in diff output.  This may require fix in ungit and/or diff2html.  I will continue to work on this...

screen shot 2015-03-04 at 7 15 58 pm
 
I don't think this should me merged until above short comings are addressed, at the very least 2nd point needs to be addressed. 

fixes: #545 #541

This is feature complete and ready for merge.


FredrikNoren:
Very cool!  -267 loc :+1: What's the status on the two issues you mentioned in the beginning?

codingtwinky:
Shoot, I forgot to link to here.  Problem where delete diff was not generating html diff was fixed here:  https://github.com/rtfpessoa/diff2html-nodejs/pull/2.  

However problem with new file diff not being generated is more on the Ungit side. 
Git diff doesn't generate any diff to parse as new file is not staged.  I think there are two options: stage the new files when they are not staged, which may affect other operations as they may or may not expect new files to be upstaged, or just fake the diff output.

I'm leaning towards the faking the git diff output but let me know.

FredrikNoren:
Yeah faking sounds like the better path, would be weird if just running ungit got files staged

Ajedi32:
Some of the answers here might be helpful: http://stackoverflow.com/q/855767/1157054 (Obviously `git add -N` isn't what we want, but there are a number of other solutions there that look promising.)

codingtwinky:
So this now handles both new and delete file diffs. 

* it is comparing /dev/null and new files to generate git diff.  Which for some reason has return code of `1` instead of `0`.  Which I had to add a variable to represent allowed return code per git commands to handle this.
* `/dev/null` doesn't work for windows and for windows I'm comparing to its equivalent `NUL`.  Which I have no idea it works or not as I don't have M$ machines due to religious reasons.
* Rolling load is not address in this pull...

FredrikNoren:
@codingtwinky Hm I'm a little hesitant to merge this if there's no upgrade for the end user experience. It's nice to get rid of code but if we lose the rolling loads then I feel like it's a small step backwards right? Maybe better to wait until diff2html supports rolling loads (if I understood it correctly that was the reason why this didn't support rolling loads?)

codingtwinky:
Although I think missing rolling load feature is rather minor, I absolutely agree that this one may need more time and mature.  

Will work on getting the rolling load for both side by side and default view shortly...

codingtwinky:
Added rolling load and many more refactoring to diff.

codingtwinky:
Please note that this is not tested in windows.

`/dev/null` is being used at [this](https://github.com/FredrikNoren/ungit/pull/514/files#diff-3d97d6a0bc43fe2abcbb34e038981c44R221) line and for windows `NUL` is being used based on search result that tells me it is the equivalent of that.  But not sure if it works.

campersau:
@codingtwinky I have tested it with `NUL` on windows and it works.

components/commit/commit.html
```diff
@@ -28,11 +28,11 @@
         <div class="body" data-bind="text: body"></div>
         <div class="diff-wrapper" data-bind="visible: showCommitDiff, style: diffStyle, click: stopClickPropagation">
           <div class="btn-group btn-group-xs" data-bind="visible: selected">
-            <button class="btn btn-default bootstrap-tooltip pull-right" data-ta-clickable="graphDefaultView" data-bind="click: showSideBySideDiff, css: {active: textDiffTypeIndex() === 1}" data-toggle="tooltip" data-placement="bottom" data-original-title="Show side by side diff view" data-delay='{"show":"2000", "hide":"0"}'>
-              <span data-bind="text: textDiffOptions[1].name"></span>
+            <button class="btn btn-default bootstrap-tooltip pull-right" data-ta-clickable="commit-sideBySideDiff" data-bind="click: textDiffTypeChange.bind($data, 'sidebysidediff'), css: {active: textDiffType() === 'sidebysidediff'}" data-toggle="tooltip" data-placement="bottom" data-original-title="Show side by side diff view" data-delay='{"show":"2000", "hide":"0"}'>
+              <span data-bind="text: 'Siede by Side'"></span>
```
campersau: I think the text should be `Side by Side`.

components/staging/staging.html
```diff
@@ -56,22 +56,22 @@
             Stash all
             <!-- ko component: stashProgressBar --><!-- /ko -->
           </button>
-          <button class="btn btn-default bootstrap-tooltip pull-right" data-bind="click: viewTypeChangeClick.bind($data, 1), css: {active: textDiffTypeIndex() === 1}" data-toggle="tooltip" data-placement="bottom" data-original-title="Show side by side diff view" data-delay='{"show":"1000", "hide":"0"}'>
-            <span data-bind="text: textDiffOptions[1].name"></span>
+          <button class="btn btn-default bootstrap-tooltip pull-right" data-ta-clickable="stage-sideBySideDiff" data-bind="click: textDiffTypeChange.bind($data, 'sidebysidediff'), css: {active: textDiffType() === 'sidebysidediff'}" data-toggle="tooltip" data-placement="bottom" data-original-title="Show side by side diff view" data-delay='{"show":"2000", "hide":"0"}'>
+            <span data-bind="text: 'Siede by Side'"></span>
```
campersau: I think the text should be `Side by Side`.

codingtwinky:
Thanks @campersau !

FredrikNoren:
@codingtwinky Can you run me through what the current state of this is? (Ready to be merged? Same functionality as the home spun previous solution? Tests ok? Any work left? etc. etc.)

campersau:
The ["load more"](https://github.com/FredrikNoren/ungit/pull/376) feature is now implemented on the [client side](https://github.com/codingtwinky/ungit/blob/diff2html/components/textdiff/textdiff.js#L72-L96) which might cause a higher memory usage for the browser. But I haven't checked that yet.

campersau:
I added syntax highlighting [here](https://github.com/codingtwinky/ungit/pull/3) but this should not be a requirement for merging.

codingtwinky:
fixes: #545 #541 

This is feature complete and ready for merge.

FredrikNoren:
Alright merging this one now, let's fix any outstanding issues (if there are any) in separate PRs

FredrikNoren:
Also want to say: Amazing work on this one! Really sweet with losing ~300 lines of code!!

Issue #513, opened 2015-03-06
Remote editing/removal


mesosteros:
Is there currently anyway to either edit or fully remove a remote url?
I've been using git-gui for that but obviously I would like to do it directly in ungit.

PR #518, opened 2015-03-22
Delete submodule via UI

codingtwinky:
I was working on #513 and got side tracked and did it for submodule first....

Will do same thing for remotes in similar fashion.

screen shot 2015-03-22 at 2 00 57 pm

### DIFF
--- a/components/submodules/submodules.js
+++ b/components/submodules/submodules.js
@@ -18,7 +18,7 @@ function SubmodulesViewModel(server, repoPath) {
 }
 
 SubmodulesViewModel.prototype.onProgramEvent = function(event) {
-  if (event.event == 'submodule-added') this.fetchSubmodules();
+  if (event.event == 'submodule-fetch') this.fetchSubmodules();
 }
 
 SubmodulesViewModel.prototype.updateNode = function(parentElement) {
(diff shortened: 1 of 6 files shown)


codingtwinky:
I thought I added test but it's missing so will update it later...

components/submodules/submodules.js
```diff
@@ -83,3 +85,23 @@ SubmodulesViewModel.prototype.submoduleLinkClick = function(submodule) {
...
+  diag.closed.add(function() {
+    if (diag.result()) {
+      self.addProgressBar.start();
+      self.server.del('/submodules', { path: self.repoPath, submodulePath: submodule.path, submoduleName: submodule.name }, function(err, result) {
+        if (err) {
+          console.log(err);
+          return;
+        }
+
+        programEvents.dispatch({ event: 'submodule-added' });
```
campersau: I think the event should be `submodule-removed`. You can also add it here: https://github.com/codingtwinky/ungit/blob/513/components/submodules/submodules.js#L21

codingtwinky:
@campersau  "submodule-added" is an event that really doesn't suit it's name as all it does is fetch.  I just changed the name to "submodule-fetch" so it is more fitting.

FredrikNoren:
GJ!

PR #520, opened 2015-03-28
fix long file path name: fix #317

### LINKED ISSUE #317: Handle long file paths in the staging area in a slightly better way
When ungit has files stages where the file paths are quite long, it handles them, but it might make more sense to handle them in a better way (e.g. reducing font size, breaking up files by directory, breaking lines at intelligent places, etc) to make the files more easily readable. (See the image below for what I mean)

screen shot 2014-02-26 at 5 14 09 pm

Also, it would be nice if files next to each other in the staging list had alternating light/dark backgrounds. That way, it would be easier to tell which name goes with which checkmark when indicating whether or not to add them to the commit.

codingtwinky:
screen shot 2015-03-26 at 9 40 40 pm

### DIFF
--- a/components/staging/staging.less
+++ b/components/staging/staging.less
@@ -109,6 +109,13 @@
         border: 0px;
         border-radius: 3px;
         color: rgba(255, 255, 255, 0.8);
+        overflow: hidden;
+        word-wrap: break-word;
+        width: ~"calc(100% - 60px)";
+        white-space: nowrap;
+        text-overflow: ellipsis;
+        direction: rtl;
+        text-align: left;
       }
       padding: 0.3em;
 


FredrikNoren:
Just noticed that this didn't work on my screen:

screen shot 2015-04-04 at 2 36 43 pm

Reverting this commit

codingtwinky:
What is the text supposed to be?  And are you using chrome?  I've noticed that it is little flacky in chrome.  Need to do more testing/experiments

FredrikNoren:
@codingtwinky the "Next i x" is on a new line, but it's supposed to be directly after the "wdf"

PR #523, opened 2015-04-01
Add IPv6 test for localhost ip

bensagardoy:
Adding IPv6 localhost address to the test for OS X Yosemite that appears to have made IPv6 default over IPv4 (confirmed on Yosemite 10.10.2 with the use of the node api OS -> os.networkInterfaces).


FredrikNoren:
@bensagardoy Awesome! Thanks!

PR #526, opened 2015-04-05
Put confirm box to discard changes button fix #524

### LINKED ISSUE #524: Put confirm box to discard changes button
ekran resmi 2015-04-05 00 41 48

I accidentally discarded my changes , by clicking this **X** button. It would be great if you put a confirmation box to discard changes. It will pretend losing datas.

erayalakese:
fix for issue #524

### DIFF
--- a/clicktests/test.generic.js
+++ b/clicktests/test.generic.js
@@ -112,6 +112,7 @@ suite.test('Should be possible to discard a created file', function(done) {
     if (err) return done(err);
     helpers.waitForElement(page, '[data-ta-container="staging-file"]', function() {
       helpers.click(page, '[data-ta-clickable="discard-file"]');
+      helpers.click(page, '[data-ta-clickable="yes"]');
       helpers.waitForNotElement(page, '[data-ta-container="staging-file"]', function() {
         done();
       });
--- a/components/staging/staging.js
+++ b/components/staging/staging.js
@@ -329,7 +329,12 @@ FileViewModel.prototype.toggleStaged = function() {
   this.staged(!this.staged());
 }
 FileViewModel.prototype.discardChanges = function() {
-  this.server.post('/discardchanges', { path: this.staging.repoPath, file: this.name() });
+  var self = this;
+  var diag = components.create('yesnodialog', { title: 'Are you sure you want to discard these changes?', details: 'This operation cannot be undone.'});
+  diag.closed.add(function() {
+    if (diag.result()) self.server.post('/discardchanges', { path: self.staging.repoPath, file: self.name() });
+  });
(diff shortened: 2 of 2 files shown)


erayalakese:
master branch failing in Travis tests, too.

FredrikNoren:
@erayalakese Thanks for the patch! However I wonder if we can fix this without having to have another dialog? (It gets old real fast when you want to discard a bunch of files). One idea I've been thinking about is to commit only the file you want to discard and then immediately move the ref back one commit, so that if you mess up you can find the discarded file in the reflog, but from a user perspective it would still look like today. What do you think?

erayalakese:
@FredrikNoren  I think 'commiting and moving back' is a slower process and cause unwanted commits. Also i think it will cause problem while sync with server (I mean pull/push) . I think dialog box is *oldie but goldie* :) 

But you are right it will be a problem if user wants to discard a bunch of files, so we can put a **Discard selected files** button next to **Discard all buttons** ?

Ajedi32:
I think @FredrikNoren's point is that undo is usually better than confirm, and should be used in place of confirm whenever possible, and I agree.

There might be cases though where committing large files would  be a performance issue, and we should definitely be careful not to mess up the staging area, but in the general case I agree that undo should be prefered over confirmation dialogs whenever possible.

Why? Because probably about 90% of the time that the user clicks the delete button, they really mean it. That means that with a confirm message you'd be annoying the user and slowing him down 90% of the time, while simultaneously training the user to click the "Confirm" button without even thinking about it, increasing the likelihood of accidental data loss in the future. Undo has neither of those problems, and still allows the user to recover lost data if he accidentally clicks delete.

Ajedi32:
On the other hand, a confirm message is way easier to implement than an undo button, and IMO a confirm message is still far better than no recovery option at all for an operation which causes permanent (short of a disc recovery tool) data loss.

erayalakese:
I've sent this pull request because on hour befor this request I've lost my 2 hours of work by clicking to *discard* button accidentally :)

But I agree with you about *training them to click the "Confirm"* .

You can implement confirm message immediately for now and implement *undo button* in next releases ?

FredrikNoren:
Hm alright, I really wish we could do this with undo but I honestly can't think of any easy way (other than the commit - move back, but I can see your arguments that that might have hidden implications too). Sure lets go with the dialog for now, thanks for the PR @erayalakese

erayalakese:
You are welcome, and thank you for this great project. I hope I'll find another things to contribute Ungit :)

lusk:
I wish you would also include a way to disable the dialog for a certain amount of time (5 minutes) if not at all (using a configuration option) because it gets pretty annoying very fast indeed :( "Discard selected files" button would be awesome though! :)

PR #530, opened 2015-04-07
Add ahed & behind stats. fix 506

codingtwinky:
screen shot 2015-04-04 at 2 12 03 pm


components/staging/staging.js
```diff
@@ -83,6 +83,26 @@ var StagingViewModel = function(server, repoPath) {
...
+      return;
+    }
+    self.ahead(ahead ? ahead.split('\n') : []);
+  });
+  this.server.get('/behind', { path: this.repoPath }, function(err, behind) {
+    if (err) {
+      console.log("error while acquiring behind info", err);
+      return;
+    }
+    console.log(behind);
```
campersau: There isn't a `console.log` for `ahead` so why is it here?

codingtwinky: Because I'm clumsy... :P

FredrikNoren:
Sorry for the hopelessly slow response time for this one, but there should probably be some clicktests for this as well.

Generally, I think we can almost skip the other "server" tests and just write the same tests in the clicktests instead, since that's what matters anyway (that it works end-to-end).

Also I'm not super clear on the value of having this, since you can kind of see how much ahead or behind you are in the graph?

nicolaiskogheim:
I am in favour of these stats.

codingtwinky:
This is one of those icing on the top that isn't so critical.   will add a click test later

components/staging/staging.html
```diff
@@ -13,10 +13,14 @@
...
-                  <span class="commit-message-title-counter" data-bind="text: commitMessageTitleCount"/>
+            <div>
+              <span class="checkmark" data-bind="css: { checked: amend }">&#10003;</span>
+                <a href="#" data-bind="click: toggleAmend">Amend last commit</a>
+              </span>
+              <span class="commit-message-title bootstrap-tooltip" data-bind="text: '#' + commitMessageTitleCount()" data-toggle="tooltip" data-placement="bottom" data-original-title="# of characters in commit message" data-delay='{"show":"1000", "hide":"0"}'/>
+              <span class="commit-message-title bootstrap-tooltip" data-ta-container="ahead" data-bind="text: '&#9650;' + ahead().length"  data-toggle="tooltip" data-placement="bottom" data-original-title="# commits current is ahead compare to origin/master" data-delay='{"show":"1000", "hide":"0"}'/>
+              <span class="commit-message-title bootstrap-tooltip " data-ta-container="behind" data-bind="text: '&#9660;' + behind().length"  data-toggle="tooltip" data-placement="bottom" data-original-title="# commits current is behind compare to origin/master" data-delay='{"show":"1000", "hide":"0"}'/>
```
campersau: Nit: Space after `bootstrap-tooltip` in `class`-attribute. (This does not exist in the lines above.)

PR #534, opened 2015-04-12
Now backdrop disappering when user click "Don't Load" or "Load Anyway"

erayalakese:
Now backdrop disappering when user click "Don't Load" or "Load Anyway" in *TooManyFilesDialogViewModel* modal box.

It was a bug. Page was remaining 'blackdroped' when user click *Don't load* or *Load Anyway*

### DIFF
--- a/components/dialogs/dialogs.js
+++ b/components/dialogs/dialogs.js
@@ -125,8 +125,8 @@ function TooManyFilesDialogViewModel(title, details) {
   this.taDialogName('yes-no-dialog');
   this.result = ko.observable(false);
   this.alternatives([
-    { label: "Don't load", primary: true, taId: 'noLoad', click: function() { self.result(false); self.close(); } },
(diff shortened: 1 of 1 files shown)


codingtwinky:
:+1:

erayalakese:
@codingtwinky 

I've found real problem. When modal box appeared, these events triggered on every page focus.
image

So it's causing lots of `modal-backdrop fade in` div boxes.
image

And `close()` is closing just one of them. So backdrop is still appearing on the page.

codingtwinky:
Hm, I guess it was my fault to assumed that bootstrap opens one modal and subsequent modal open is ignored.  I think it is better to prevent from opening multiple modal is better then closing multiple dialogs.

erayalakese:
@codingtwinky exactly! 

I find out the real problem after my commit.  I will send another commit today.

FredrikNoren:
Fixed in #549 instead

PR #536, opened 2015-04-13
Easy Browsing

erayalakese:
Now it's really easy to browse folders, just push right arrow to browse into selected folder in autocomplete list.


FredrikNoren:
Sweet!

PR #459, opened 2014-11-13
Add branch dropdown

anbuf:
This new feature has the purpose to have a list of all the branches in a place and allow to easily checkout a desired branch. This component is based on the "remote" component. It used the same UI and a big part of the code. 
For the moment, no tests are implemented. 
In the future, it would be great to add a search box to this component with an autocompletion, so it will be easier to   search a branch if you have a lot of branch.

PR #550, opened 2015-05-08
Added branch list fix: #459

codingtwinky:
Added branch list, this is same as #459.

* ~~Display text is "@<<current branch>>".  I'm not sure if this is a good idea but simply displaying branch name wasn't indicative enough.~~ having branch icon in the box should be indicative enough
* Branch list sorted

screen shot 2015-05-05 at 7 42 28 pm
screen shot 2015-05-05 at 7 41 56 pm
screen shot 2015-05-05 at 7 42 17 pm

### DIFF
--- a/clicktests/test.noheader.js
+++ b/clicktests/test.noheader.js
@@ -47,10 +47,4 @@ suite.test('Shutdown', function(done) {
   environment.shutdown(done);
 });
 
-
-suite.test('Shutdown', function(done) {
-  environment.shutdown(done);
-});
-
-
 testsuite.runAllSuits();
--- a/clicktests/test.noheader.js
+++ b/clicktests/test.noheader.js
@@ -43,6 +43,10 @@ suite.test('Check for refresh button', function(done) {
   });
 });
 
+suite.test('Shutdown', function(done) {
+  environment.shutdown(done);
+});
+
 
 suite.test('Shutdown', function(done) {
   environment.shutdown(done);
--- a/components/branches/branches.js
+++ b/components/branches/branches.js
@@ -32,5 +32,10 @@ BranchesViewModel.prototype.checkoutBranch = function(branch) {
   // checkout branch
 }
 BranchesViewModel.prototype.updateBranches = function() {
-  // update branch list
+  var self = this;
+  this.fetchingProgressBar.start();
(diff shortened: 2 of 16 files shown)


components/branches/branches.js
```diff
@@ -0,0 +1,75 @@
...
+  return new BranchesViewModel(args.server, args.repoPath);
+});
+
+function BranchesViewModel(server, repoPath) {
+  var self = this;
+  this.repoPath = repoPath;
+  this.server = server;
+  this.branches = ko.observableArray();
+  this.fetchingProgressBar = components.create('progressBar', { predictionMemoryKey: 'fetching-' + this.repoPath, temporary: true });
+  this.fetchEnabled = true;
```
campersau: unused variable

components/branches/branches.js
```diff
@@ -0,0 +1,75 @@
...
+      var sorted = branches.sort(function(a, b) {
+        if (a.name < b.name)
+           return -1;
+        if (a.name > b.name)
+          return 1;
+        return 0;
+      });
+      self.branches(sorted);
+      self.current(undefined);
+      branches.map(function(branch) {
```
campersau: I would use `.forEach` for that. So no new array gets allocated.

FredrikNoren:
This is really cool and much needed, but the row of stuff is getting a little too crazy (small displays will have a problem here).

How about we put all of this (submodules+branches) in a dialog instead, and just hide it in a [...] button? Like this:

screen shot 2015-08-27 at 9 05 17 pm

Both of them are kind of "things you need but not all the time" anyway, so I think it would make sense to put them in a place that you can get to but that's not on the "front page".

codingtwinky:
Buttons now expand on hover...  I'm not a big fan of bundling buttons together... :P

screen shot 2015-08-29 at 5 06 22 pm

clicktests/test.noheader.js
```diff
@@ -43,6 +43,10 @@ suite.test('Check for refresh button', function(done) {
   });
 });
 
+suite.test('Shutdown', function(done) {
```
campersau: Isn't this the same as [below](https://github.com/FredrikNoren/ungit/pull/550/files#diff-adc25198ac2351b2aa9cc1351a87804bR51)?

FredrikNoren:
That's really neat, I like that!

PR #554, opened 2015-05-19
Removes notes from commit graph

### LINKED ISSUE #553: Do not show notes as orphans
When working with git-tfs, it adds a note with the workitem id to the commit that has been checkedin.

In ungit, those notes apear as orphaned commits
orphaned commits

ylecuyer:
FIxes #553

### DIFF
--- a/source/git-api.js
+++ b/source/git-api.js
@@ -270,7 +270,7 @@ exports.registerApi = function(env) {
   app.get(exports.pathPrefix + '/log', ensureAuthenticated, ensurePathExists, function(req, res){
     var limit = '';
     if (req.query.limit) limit = '--max-count=' + req.query.limit;
-    git(['log', '--decorate=full', '--date=default', '--pretty=fuller', '--all', '--parents', '--numstat', '--topo-order', limit], req.query['path'])
+    git(['log', '--decorate=full', '--date=default', '--pretty=fuller', '--branches', '--tags', '--remotes', '--parents', '--no-notes', '--numstat', '--topo-order', limit], req.query['path'])
       .parser(gitParser.parseGitLog)
       .always(function(err, log) {
         if (err) {


ylecuyer:
@FredrikNoren @codingtwinky any comment ?

codingtwinky:
I don't personally use git-tfs nor know much about it but yes this is a good way to fix this #553.

However, if I were to make a small possible suggestion I would do this somewhere in the [server side](https://github.com/FredrikNoren/ungit/blob/master/source/git-api.js#L273) despite it might screw up the "limit" variable as a person who writes software that are heavily NIC bounded and hate to see unnecessary traffic.  (Only reason why I'm bringing up this miniscule point is because I see how you are doing the [looping](https://github.com/FredrikNoren/ungit/pull/554/files#diff-4f0cddea03b554c0ea9e04d16204f852R174) and I assume that you are one of those who are trying to be very performance conscious like me...)

ylecuyer:
One way to reproduce the issue is:

```
mkdir test
cd test
git init
echo "test" > a.txt
git add -A
git commit -m "Initial commit"
git notes add -m 'Test' `git rev-parse HEAD`
ungit
```

I have already tried fixing the issue server side but it is not possible.

campersau:
You can hide the notes by setting each glob individually `--branches --tags --remotes` instead of setting `--all` and then the `--no-notes` parameter does not get ignored.
The full command would look like this:

```sh
$ git log --decorate=full --date=default --pretty=fuller --branches --tags --remotes --no-notes --parents --numstat --topo-order
```

And here is a patch for `source/git-api.js`

```diff
--- a/source/git-api.js
+++ b/source/git-api.js
@@ -270,7 +270,7 @@ exports.registerApi = function(env) {
   app.get(exports.pathPrefix + '/log', ensureAuthenticated, ensurePathExists, function(req, res){
     var limit = '';
     if (req.query.limit) limit = '--max-count=' + req.query.limit;
-    git(['log', '--decorate=full', '--date=default', '--pretty=fuller', '--all', '--parents', '--numstat', '--topo-order', limit], req.query['path'])
+    git(['log', '--decorate=full', '--date=default', '--pretty=fuller', '--branches', '--tags', '--remotes', '--parents', '--no-notes', '--numstat', '--topo-order', limit], req.query['path'])
       .parser(gitParser.parseGitLog)
       .always(function(err, log) {
         if (err) {
```

But maybe we should make this feature optional/configurable?

ylecuyer:
I modified the PR with your hint @campersau

I think it should not be configurable because it's not useful to see notes as commits.

Issue #542, opened 2015-04-21
Couldn't load ungit. v0.9.2 with node v0.12 


yanosh-igor:
On v0.8.4 everything works fine with node v0.12.
But with v0.9.2 when I type ungit it opens browser and loading very long and at the end shows "Connection lost Refresh the page to try to reconnect".

Attached console information and screen from browser:
![console2 - ungit 2015-04-21 13 04 27](https://cloud.githubusercontent.com/assets/435827/7249978/0c2a12bc-e827-11e4-8b39-b44d37db669a.png)

![dukejobs _ e_ - google chrome 2015-04-21 13 03 44](https://cloud.githubusercontent.com/assets/435827/7249991/2b2389a0-e827-11e4-8c7e-cea8333ec4c4.png)

Could you please suggest something.
.

FredrikNoren:
Try uninstalling `npm uninstall -g ungit`, restarting your machine and then installing `npm install -g ungit@latest`

codingtwinky:
Also do you see anything interesting on the client side log?

FredrikNoren:
@codingtwinky the last thing in his log is /baserepopath which previously froze the whole process so I suspect that somehow he's still on old code somehow

codingtwinky:
@yanosh-igor is it possible that having two different versions, dev version and global installed npm version, and are you confusing one from another?

yanosh-igor:
Tried reinstall ungit(remove/install).
Run ungit command in console, sometimes its loaded properly and sometimes not.

yanosh-igor:
@codingtwinky no, I have installed ungit only globally.

yanosh-igor:
Have noticed that ungit loads only from subdirectory, not from root directory where my git project initialized.

codingtwinky:
My suspicion is that [this line](https://github.com/FredrikNoren/ungit/blob/master/source/git-api.js#L501) doesn't work for you as your git repo location is at `E:/dukejobs`.

@FredrikNoren what exactly is the purpose of `isRoot` function?  Maybe I'm missing something but function suppose to determine whether a path is NOT root rather then is root.  Also for windows why is it length > 3?

brianwalleshauser:
@yanosh-igor I was having the exact same issue

@codingtwinky this is probably terrible, but I went to that line and changed the 3 to a 1.  Everything seems to be working for now

FredrikNoren:
[...]

PR #556, opened 2015-05-22
fix baserepopath for windows

### LINKED ISSUE #537: Server hangs on start
For any repo, even an empty one, the server quickly hangs and then the webapp says "Connection lost".

last console line is always " - info: GET /api/baserepopath?path=whatever"

node.exe uses a lot of processing power and ~100MB of memory while hanging

I'm on Windows 7, fresh install of node. I can provide more logs and info if instructed how to

campersau:
The `/baserepopath` api resulted in an infinite loop on windows because the length of the root path is 3 and not 1.

The fix in https://github.com/FredrikNoren/ungit/pull/540 didn't work because the `isRoot` function would always return `true` if the path length was ***greater*** than 3 (win) or 1 (other). The change was reverted in https://github.com/FredrikNoren/ungit/pull/551.
 
This fix also checks at the end if `currentPath` is a git directory which allows `subst` git directories on windows to work correctly. (On windows you can create virtual drives which are linked to directories.)

This should now really fix #537, #542

---

~~We could also use [`git rev-parse --show-toplevel`](http://stackoverflow.com/a/957978)~~

### DIFF
--- a/source/git-api.js
+++ b/source/git-api.js
@@ -498,14 +498,19 @@ exports.registerApi = function(env) {
   app.get(exports.pathPrefix + '/baserepopath', ensureAuthenticated, ensurePathExists, function(req, res){
     var currentPath = path.resolve(path.join(req.query['path'], '..'));
     function isGitDirectory(currentPath) {
-      return fs.existsSync(path.join(currentPath, '.git')) && fs.statSync(path.join(currentPath, '.git')).isDirectory();
+      var gitDirectory = path.join(currentPath, '.git');
+      return fs.existsSync(gitDirectory) && fs.statSync(gitDirectory).isDirectory();
+    }
+    function isRoot(currentPath) {
+      if (os.platform().indexOf('win') == 0) return currentPath.length <= 3;
+      else return currentPath.length <= 1;
     }
 
-    while (currentPath.length > 1 && !isGitDirectory(currentPath)) {
+    while (!isRoot(currentPath) && !isGitDirectory(currentPath)) {
       currentPath = path.resolve(path.join(currentPath, '..'));
     }
 
-    if (currentPath.length > 1) res.json({ path: currentPath });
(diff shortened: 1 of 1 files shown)


campersau:
I have added a second commit which uses the `git rev-parse --show-toplevel` command which worked for me too. So you can decide which method do you like most :smirk:

codingtwinky:
So `/baserepopath` is one of the apis that is terribly named.  It is not actually looking for "base repo" as you might think, it is actually looking for parent git repo of base repo.

For example, let's think of a case where `/var/home/repo1` and `/var/home/repo1/repo2`, where both of those two are git repos.  at the `repo2`'s location, `/baserepopath` is expecting `/var/home/repo1`.  The reason is that ungit on it's own keep track of base repo so it isn't really need for it and we need something like this for submodule related operation.

Now having said that only time it is used for is at [here](https://github.com/FredrikNoren/ungit/blob/master/components/repository/repository.js#L58) and it could use refactoring or better logic.

campersau:
Ah thank you @codingtwinky for the explanation I haven't looked where the api is actual used.
This makes the fix with `git rev-parse --show-toplevel` invalid. I will remove this commit soon.

ginus:
if the project  dir is top of a disk it still not work

campersau:
@ginus can you please describe exactly what does not work?
(There still seems to be a bug that when you are in a submodule folder you can not go to the base repo)

ginus:
@campersau my project dir is f:\apisvr, and the .git in it, without submodule. when I start ungit from the  cmd I see the server side stoped at /baserepopath api and browser side tell me connection lost.

campersau:
@ginus what version of ungit do you use? This fix is not released yet, it was just merged into master. If you want you can install the latest master manually see https://github.com/FredrikNoren/ungit/issues/569#issuecomment-109941943 .

PR #557, opened 2015-05-22
show additions and deletions in staging area

### LINKED ISSUE #132: Show number of lines added/removed in staging

campersau:
fixes #132

image

### DIFF
--- a/components/staging/staging.js
+++ b/components/staging/staging.js
@@ -290,6 +290,8 @@ var FileViewModel = function(staging, name, textDiffType) {
   this.isShowingDiffs = ko.observable(false);
   this.diffProgressBar = components.create('progressBar', { predictionMemoryKey: 'diffs-' + this.staging.repoPath, temporary: true });
   this.textDiffType = textDiffType;
+  this.additions = ko.observable('-');
+  this.deletions = ko.observable('-');
   this.diff = ko.observable(self.getSpecificDiff());
 }
 FileViewModel.prototype.getSpecificDiff = function() {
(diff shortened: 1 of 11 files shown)


components/staging/staging.js
```diff
@@ -325,6 +335,8 @@ FileViewModel.prototype.setState = function(state) {
   this.isNew(state.isNew);
   this.removed(state.removed);
   this.conflict(state.conflict);
+  this.additions(state.additions);
+  this.deletions(state.deletions);
```
codingtwinky: instead of `this.deletionsFormatted` or `this.additionsFormatted`, I think you could do something like`this.additions(state.additions ? '+' + state.additions: '')`;

campersau: @codingtwinky I removed the `*Formatted` observables.

FredrikNoren:
Very nice! Any chance we can add a clicktest to this to make sure it stays working?

campersau:
@FredrikNoren I have added a click test. But I had to change the `waitForElement` function to return the found element so I can access the `textContent` to verify that the count is correct. Is this the correct way to do it or should I create a new helper method like `waitForElementTextContent`?

FredrikNoren:
@campersau `waitForElementTextContent` makes sense to me, but it's not a huge deal either way.

FredrikNoren:
@campersau Looking good! Can you please bump it (my bad, I've merged stuff in between that got in the way) and it's good to go

campersau:
Rebased and travis is green.

Issue #541, opened 2015-04-21
Most recent commit diffs don't open


mesosteros:
New issue: the most recent commit does not show me the diff for each file when I click it. However if the commit isn't the most recent, it does:

![commit diff](https://cloud.githubusercontent.com/assets/1943233/7249325/7ac4b7ec-e811-11e4-9e52-7924345ce422.gif)

The cursor is off in the gif because of the gif capturing software. I'm clicking the file names.

codingtwinky:
I just checked out 0.9.2 and i don't see this in mac...  Do you see any errors on server or browser log?

mesosteros:
No errors. This is in Windows 8.1 x64 bits.
Turns out it isn't the most recent commit: just some commits. I have no idea why though.
Server shows no errors, and in the browser log I only get these (which are not errors:

Fire change event due to re-activity
Event: Object {event: "working-tree-changed"}
Event: Object {event: "init-tooltip"}

codingtwinky:
Could you please run this command `git diff fb049b55^^! -- getYourTarget/client/lib/router.js` at `C:\users\mesos_000\Desktop\Projects` and share the output?  also could you try with single carrot as well just like this?  `git diff fb049b55^! -- getYourTarget/client/lib/router.js`

campersau:
I have the same issue on Windows 8.1 x64 that some diffs do not open.

I have tracked it down to these lines: https://github.com/FredrikNoren/ungit/blob/master/source/git.js#L59-L81
The `stdout` variable is empty if the `gitProcess` is closed. It looks like the child process starts so quickly that the `data` event handlers are attached too late to get called.

If I run the git comand in the command prompt I get the expected output.

---

Here is a test case to demonstrate the issue.
Create a new file in your ungit repository and run it with node. Toggle the two `commands` variables. to see the different results.
```js
var child_process = require('child_process');
var path = process.cwd(); // set this to the ungit repository path if the cwd is somewhere else

// not working commands
var commands = ['diff', 'e6f115fcdedd97fb68701614bd85b368c1a1f53c^^!', '--', 'components/stash/stash.js'];
// working commands
//var commands = ['diff', '34e717f76a195582501e038567b47e083827a07f^^!', '--', 'source/git-parser.js'];

var gitProcess = child_process.spawn('git', commands, { cwd: path });

var stdout = '';
gitProcess.stdout.on('data', function(data) {
    stdout += data.toString();
});
var stderr = '';
[...]

PR #560, opened 2015-05-26
use child_process.exec for windows

### LINKED ISSUE #559: credentials helper does not work correctly
With the latest version of ungit the credentials helper in ungit does not show up in the browser. Instead I have to insert my credentials inside the command prompt.

Here is an example output:
```
2015-05-26T15:37:40.375Z - info: git executing: P:\ungit -c color.ui=false -c core.quotepath=false -c core.pager=cat -c credential.helper="P:/ungit/bin/credentials-helper 0 8448" fetch origin --prune

Password for 'https://campersau@github.com':
```

Maybe it is also related to the `child_process.spawn` change?

Windows 8.1 x64
Node v0.12.3
Git 1.8.3.msysgit.0

campersau:
I know this is sad but I could not find any other way to fix the issues with `child_process.spawn` on windows...
This fixed #559 for me.

Please test this fix on other OSes too.

### DIFF
--- a/source/git.js
+++ b/source/git.js
@@ -56,35 +56,56 @@ var gitQueue = async.queue(function (task, callback) {
   git.runningTasks.push(task);
   task.startTime = Date.now();
 
-  var gitProcess = child_process.spawn(
-    'git',
-    task.commands,
-    {
-      cwd: task.repoPath,
-      maxBuffer: 1024 * 1024 * 100,
-      encoding: task._encoding,
-      timeout: task._timeout
+  var gitProcess;
+
+  if (isWindows) {
+    gitProcess = child_process.exec(
+      'git ' + task.commands.join(' '),
+      {
+        cwd: task.repoPath,
+        maxBuffer: 1024 * 1024 * 100,
+        encoding: task._encoding,
+        timeout: task._timeout
+      },
+      function(error, stdout, stderr) {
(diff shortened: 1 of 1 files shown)


Ajedi32:
If you couldn't get `child_process.spawn` to work correctly on Windows, perhaps you might want to put together a reproducible example and submit a bug report to [joyent/node](https://github.com/joyent/node).

In the mean time this seems like an acceptable fix to, provided it doesn't introduce any regressions. For example, one of the reasons we switched to `child_process.spawn` in the first place was to fix #486. Does this patch re-introduce that problem?

campersau:
@Ajedi32 here is an example for this bug: https://github.com/FredrikNoren/ungit/issues/541#issuecomment-105544642
Sadly it does re-introduce the problem on windows...

Ajedi32:
Cool. You want to open an issue for that on [joyent/node](https://github.com/joyent/node)? I'd do it myself, but I'm not nearly as familiar with the problem as you are, and I have limited experience working with node. (You can reference the issue in this repository from the created issue on node's repo as `FredrikNoren/ungit#541`.)

campersau:
Since I found the real bug for #541 (see #563) I will close this PR and try to find another way to fix #559 .

codingtwinky:
Hmm...  This is a lovely bug...  Thanks

PR #562, opened 2015-05-27
support git protocol in urls in submodule parser

campersau:
the parser now parses the following uris to http urls:
```
https://github.com/FredrikNoren/ungit.git => https://github.com/FredrikNoren/ungit.git
git@github.com:FredrikNoren/ungit.git => http://github.com/FredrikNoren/ungit.git
git://github.com/FredrikNoren/ungit.git => http://github.com/FredrikNoren/ungit.git
```
this also fixes a bug when the uri contains `http` not at the start of the uri

### DIFF
--- a/source/git-parser.js
+++ b/source/git-parser.js
@@ -308,9 +308,16 @@ exports.parseGitSubmodule = function(text, args) {
   var getUrl = function(line) {
     var url = line.substr(line.indexOf("= ") + 1).trim();
 
-    // When a repo is checkout with ssh instead of an url
-    if (line.indexOf('http') < 0) {
-      url = 'http://' + url.substr(url.indexOf('@') + 1).replace(':', '/');
+    // keep a reference to the raw url
+    submodule.rawUrl = url;
+
+    // When a repo is checkout with ssh or git instead of an url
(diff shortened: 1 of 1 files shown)


Ajedi32:
Out of curiosity, why do the repository references need to be rewritten like that? git should be able to handle those formats as-is, right?

campersau:
I think @codingtwinky wanted to rewrite the references so they can be opened in the browser.
Which of course only makes sense if a webserver exists behind the url. (Github repositories support that.)

source/git-parser.js
```diff
@@ -308,9 +308,16 @@ exports.parseGitSubmodule = function(text, args) {
   var getUrl = function(line) {
     var url = line.substr(line.indexOf("= ") + 1).trim();
 
-    // When a repo is checkout with ssh instead of an url
-    if (line.indexOf('http') < 0) {
-      url = 'http://' + url.substr(url.indexOf('@') + 1).replace(':', '/');
+    // keep a reference to the raw url
+    submodule.rawUrl = url;
```
codingtwinky: I don't think we need this

codingtwinky:
:+1:

PR #563, opened 2015-05-27
fix windows diffs

### LINKED ISSUE #541: Most recent commit diffs don't open

campersau:
the double carrot (^) which was introduced in #349 is not required if we use `child_process.spawn`

fixes #541

### DIFF
--- a/source/git.js
+++ b/source/git.js
@@ -10,7 +10,6 @@ var addressParser = require('./address-parser');
 var GitTask = require('./git-task');
 
 var gitConfigArguments = ['-c', 'color.ui=false', '-c', 'core.quotepath=false', '-c', 'core.pager=cat'];
-var isWindows = /^win/.test(process.platform);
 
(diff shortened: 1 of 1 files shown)


Ajedi32:
Ah! That makes sense. Good bug hunting! :+1:

codingtwinky:
:+1: 

So I hate to be that guy who holds on the pull requests forever but as you may know #514 changes how diff work in a very big way.  Could I get you to potentially help me with it?  Please feel free add to the PR.

campersau:
@codingtwinky the diffs in #514 are working on windows

FredrikNoren:
Sweet @campersau! I'll merge this one, hope it doesn't cause too much trouble for #514 @codingtwinky

PR #564, opened 2015-05-29
fix credential helper

### LINKED ISSUE #559: credentials helper does not work correctly

campersau:
quotes are not needed when using child_process.spawn

fixes #559

### DIFF
--- a/source/git-api.js
+++ b/source/git-api.js
@@ -111,7 +111,7 @@ exports.registerApi = function(env) {
 
   function credentialsOption(socketId) {
     var credentialsHelperPath = path.resolve(__dirname, '..', 'bin', 'credentials-helper').replace(/\\/g, '/');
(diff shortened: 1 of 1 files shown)


codingtwinky:
:+1:

PR #573, opened 2015-06-03
fix diff linenumbers for new files

campersau:
linenumbers should start with 1 (not with 0)

image


campersau:
The second commit adds the range info to new file diffs.

image

campersau:
Note that this is also fixed by https://github.com/FredrikNoren/ungit/pull/514

FredrikNoren:
@campersau I merged #514 instead, closing this.

PR #574, opened 2015-06-03
cleanup tests after they finished

campersau:
Cleanup the `test-temp-dir*` after the tests are done.

Without this change my temp folder was full of `test-temp-dir*` folders and `test-temp-dir*.gitignore` files.


FredrikNoren:
@campersau nice!

PR #576, opened 2015-06-06
fix too many files dialog

campersau:
before the dialog was shown but if the refresh event for the staging area was fired while the dialog was open the files would be rendered before the dialog was even closed.

this also removes duplicated code for setting `inRebase`, `inMerge` and `commitMessage*` and also makes sure that the `callback` is only called once it is done.

bug introduced with #549


FredrikNoren:
Nice!

PR #586, opened 2015-06-28
Prevent animation on commit diff view expand

codingtwinky:
On git diff, one of the problem is that nodes are pushed down as git diff is expanded and trigger animation.  Which is fine in most cases but in cases when there are a lot of changes it is rather annoying to see animation happening and inefficient.

This fix is to have git diff view to be overlay over nodes rather then pushing them down to trigger animation.

### DIFF
--- a/components/commit/commit.less
+++ b/components/commit/commit.less
@@ -32,6 +32,10 @@
           box-shadow: 5px 5px 0px rgba(0, 0, 0, 0.2);
           padding: 10px;
           padding-top: 0px;
+          position: absolute;
+          background-color: #4A5665;
+          width: 100%;
+          border-radius: 0px 0px 3px 3px;
         }
         .btn-group {
           padding: 2px;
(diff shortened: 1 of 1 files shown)


FredrikNoren:
@codingtwinky agree that the animation is frustrating, but I'm not sure if overlay is the right solution. There's code that makes sure that if you are at the bottom of a diff and close it, it doesn't jump around. Also if the diff is long I don't want it you to end up way past the commit you were looking at when you close it (which would happen if it's overlay).

Maybe see if there's a way to solve it in the current system without making it overlay?

codingtwinky:
Hmm, I see your point.  We could have give set time to animation.  Instead of having linear animation that take longer as longer distance it travels, we could animate within 2 seconds or so regardless of distance it has to travel.

Would that be better?

FredrikNoren:
It might work. Probably worth trying at least

codingtwinky:
All this time I thought animation was done via css...  But it isn't...  I may convert animations to css first...

codingtwinky:
Closing as updated graphic layer for easier animation is coming.

PR #587, opened 2015-07-01
Prompt a message in browserify-component grunt task

linquize:
For browserify-components grunt task, prompt a message if the js is not found in a particular component directory.

This helps people who compile from source to identify a compilation problem, after an old component *.js is deleted, but the directory is left.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -218,7 +218,13 @@ module.exports = function(grunt) {
         bundleExternal: false,
         debug: true
       });
-      b.add('./components/' + component + '/' + component + '.js');
+      var src = './components/' + component + '/' + component + '.js';
+      if (!fs.existsSync(src)) {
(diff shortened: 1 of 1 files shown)


Gruntfile.js
```diff
@@ -218,7 +218,12 @@ module.exports = function(grunt) {
         bundleExternal: false,
         debug: true
       });
-      b.add('./components/' + component + '/' + component + '.js');
+      var src = './components/' + component + '/' + component + '.js';
+      if (!fs.existsSync(src)) {
+        grunt.log.error(src + ' does not exist. If this component is obsolete, ' +
+          'please remove that directory or perform a clean build.');
+      }
+      b.add(src);
```
FredrikNoren: How about doing else { b.add(src) } (i.e. just log an error but don't crash when there's no component.js)

linquize: Updated at @ fb30351

PR #591, opened 2015-07-16
hiding y overflow bar for diff2html

codingtwinky:
~~Making submodule test more resilient via using ungit as submodule instead of temp git directory.~~

It also has a diff2html css fix for hiding y overflow bar as well.


codingtwinky:
It was my intention to increase stability of submodule test, which intermittently failed in past, via doing submodule add on local ungit diff.  This works on local however it seems that phantom js is picking up different location then expected for travis and fails.

it now only has diff2html css fix which is same as https://github.com/rtfpessoa/diff2html/pull/13

PR #597, opened 2015-07-22
Catch git spawn exception and better array type test

OreOreDa:
This modification is to catch the git spawn exception, and fail gracefully in case of errors (like network problems). Moreover, the test to check for type "Array" was not completely accurate


Ajedi32:
Did you mistakenly change the line endings? The diff seems a bit large. (See [Dealing with line endings](https://help.github.com/articles/dealing-with-line-endings/#platform-all))

OreOreDa:
My mistake, sorry

PR #598, opened 2015-07-22
Catch git spawn exception / better test for Array type

OreOreDa:
Catch git spawn exception / better test for Array type

### DIFF
--- a/source/git.js
+++ b/source/git.js
@@ -79,6 +79,9 @@ var gitQueue = async.queue(function (task, callback) {
   gitProcess.stderr.on('data', function(data) {
     stderr += data.toString();
   });
+  gitProcess.on('error', function (error) {
+      callback(error);
+  });
 
   gitProcess.on('close', function (code) {
     if (config.logGitCommands) winston.info('git result (first 400 bytes): ' + task.commands.join(' ') + '\n' + stderr.slice(0, 400) + '\n' + stdout.slice(0, 400));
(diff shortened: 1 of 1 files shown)


OreOreDa:
Indeed, I am developping on Windows, my mistake!
Thank you for the tip, here is the pull request with a correct diff!

source/git.js
```diff
@@ -353,7 +356,7 @@ git.commit = function(repoPath, amend, message, files) {
   if (message === undefined)
     return task.setResult({ error: 'Must specify commit message' });
 
-  if ((!(files instanceof Array) || files.length == 0) && !amend)
+  if ((!(Object.prototype.toString.call(files) === '[object Array]') || files.length == 0) && !amend)
```
Ajedi32: Just curious: how is this better? Isn't `instanceof` the correct way of checking types in JavaScript?

OreOreDa: Hi,
I used the commit API, and it rejected my commit because of the following test in "git.js" :  if ((!(files instanceof Array) || files.length == 0) && !amend)    return task.setResult({ error: 'Must specify files or amend to commit' });It turns out that the test "instanceof Array" fails.As per some blogs (e.g. http://perfectionkills.com/instanceof-considered-harmful-or-how-to-write-a-robust-isarray/),http://perfectionkills.com/instanceof-considered-harmful-or-how-to-write-a-robust-isarray/), the test should be written :I made the suggested modification, and everything worked like a charm :)
Kind regards,Olivier LE QUELLEC
Date: Thu, 23 Jul 2015 06:16:48 -0700

Ajedi32: Wow. That's pretty unintuitive. Still, I think we could do better here. My suggestion would be to use [`Array.isArray()`][isArray]. (Ungit doesn't support IE <9, does it?)

[isArray]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

OreOreDa:
I was not sure about that, and did not want to break anything.
I'll make the correction.

Ajedi32:
No problem.

FYI, you don't necessarily need to close the pull request to make changes; you can just update the pull request by pushing to the appropriate branch on your fork.

PR #601, opened 2015-07-23
Catch git spawn exception and better Array type test

OreOreDa:
The git process exception are currently never caught, making ungit crash in case of network or GIT server failure. The proposed correction is simply a call to the callback with error, when this happens.

As for the array, it appeared the test can fail, under some odd conditions.


FredrikNoren:
@OreOreDa Nice!

Issue #589, opened 2015-07-06
HiDPI support


wormeyman:
Ungit being mostly text already has great HiDPI support I have however noticed a few things that are blurry at HiDPI:
- ungit logo
- remote icon
- branch icon
- tag icon

wormeyman:
The Ungit logo could be swapped with the large version, and github does provide some alternative free icons i'm not sure where you got the icons from, if i knew where they came from we could just swap with svg.

https://octicons.github.com/
https://octicons.github.com/icon/tag/
https://octicons.github.com/icon/git-branch/
https://github.com/FredrikNoren/ungit/blob/master/assets/client/images/logoLarge.png

PR #605, opened 2015-08-01
Using Octicon

codingtwinky:
This will be replacing some of the images with octicon as suggested in #589.  However ungit logo hasn't been addressed as suggested in #589.

* imageEmbeded grunt job seems to having some trouble and didn't work for me.  In fact imageEmbed job halted grunt job without any error and prevented subsequent grunt jobs to run.  May need to investigate this.
* I may further differentiate "remote" indicator with origin vs actual remote.


FredrikNoren:
The remote icon seems to be missing from several nodes for me:

screen shot 2015-08-27 at 8 27 49 pm

codingtwinky:
Should be fixed now.

Issue #593, opened 2015-07-17
Package Ungit as an Electron App


Ajedi32:
Seems to me it's about time we simplify the installation process for ungit. [Electron](http://electron.atom.io/) has been out for a while now and seems to be pretty mature at this point ([even Microsoft is using it](http://arstechnica.com/information-technology/2015/04/microsofts-new-code-editor-is-built-on-googles-chromium/) now). 

This would make ungit much more user friendly. No need to download NPM or Node separately, and no need to run ungit in a separate console window; just launch it like a regular desktop application. And as an extra benefit, you get stuff like automatic updates and access to native OS features for free. Seems like a no-brainer to me.

PhnxRbrn:
I second this!

FredrikNoren:
:+1: definitely something I'm interested in, would be awesome to get rid of the awkward npm and node setup we have now

kwonoj:
I'm quite interested about this topic also and gave a quick try, was able to bring ungit as standalone (not packaged yet, dev build only). But it has one hack while bootstrapping application instance. While current ungit initialization rely on console output from forever instance ('## ungit started ##'), this mechanism does not works on electron since all of console output is redirected into chrome logging and does not goes to stdout so forever-monitor does not able to raise event for stdout properly. 

Still looking into way to utilize current codebase without heavily modifying it but also able to hook it inside of electron app bootstrapping. Meanwhile, if there's any suggestion I'd gladly give it a try by myself.

Ajedi32:
I'm certainly no expert here, but would the functionality of `forever-monitor` even be necessary with Electron? It seems Electron has its own architecture for managing the "server" process. (I can definitely understand trying to keep as much of the current architecture as possible with the initial "just get it working" bit though.)

kwonoj:
I'm also exploring of possibilities to reach stage of make it work, not concluded exact way to approach to either to use current bootstrapping using forever-monitor or other way. While initial exploring I don't recall specific mechanism provided by electron, may need read docs once again.

Ajedi32:
I didn't read a whole lot, but my impression was that the server in ungit would correspond to the "main process" in Electron. (<http://electron.atom.io/docs/v0.29.0/tutorial/quick-start/>) Maybe not though?

kwonoj:
Since main process is default node process for electron, my original thought was it's still required to spawn server instance as same as current ungit does. Maybe you're correct and worth to give it a try.

kwonoj:
[...]

PR #607, opened 2015-08-18
Added windows shortcut

internetimagery:
To further abstract from the command line. This shortcut uses windows built in environment variables to open ungit in whatever directory the shortcut is clicked. Nifty to copy into a project for people who don't want to touch cmd. :)


campersau:
I think the file can be named`ungit.lnk`. What do you think?

Ajedi32:
Have you tested this to make sure it works for people other than you? I'm not on Windows right now, but inspecting the link file in a text editor I did see `\\HP-ZBOOK\Users\00maczone\AppData\Roaming\npm\ungit.cmd`, which seems likely to be specific to your setup.

internetimagery:
Yeah it can be named anything. Also it could have an ungit icon added I assume (I believe icons are baked into the file?)

Yeah I've tested it on a couple computers. Been installing ungit on some animators computers I'm collaborating with, who are less technical.  :)
Windows is weird... I see what you mean. But after opening it in an editor as well and scrolling forever I can see that path you mentioned, and also the unresolved path is there:

C:\Users\%USERNAME%\AppData\Roaming\npm\ungit.cmd

Is it there for you? Had me worried. :palm_tree:

Ajedi32:
Yeah, the unresolved path is also present. Like I said, I'm not on Windows, so I can't really test the shortcut myself, but as long as you've tested it on other computers I think it should be fine.

Ajedi32:
Also related: #593 (Both these issues make using Ungit easier for non technical users.)

codingtwinky:
I don't know how I feel about this.  isn't it arguable that end goal of #593 and this pr same?  I would vote for #593 as it will be OS independent, especially if #593 prevent us from having binary file being checked into git repo.

Hopefully pr for #593 come soon.

codingtwinky:
Closing in favor of #610

Issue #313, opened 2014-02-21
Allow hosting under a path


l8nite:
I'd like to run ungit under a path (like http://myserver/ungit) -- 

I tried setting urlBase in the configuration, but then I got some weird stuff like http://myserver:8080/ungit:8080 in the browser.

PR #609, opened 2015-08-25
Support the use of a URL root path

geeklain:
It adds support to run ungit under a specific URL root path like http://host:port/root/path .
This fixes https://github.com/FredrikNoren/ungit/issues/313

On the server side, the basic idea is to add a top level middleware in express that rewrites URLs to remove the root path and then moves forward to other middlewares (that do not need any change).
The socket io version had to be bumped up to better support the root path starting with a slash (http://socket.io/docs/migrating-from-0-9/#configuration-differences)

On the client side, there was quite a few changes to support the concept of a root path.

All the tests are succeeding.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "semver": "~5.0.1",
     "serve-static": "~1.10.0",
     "signals": "~1.0.0",
-    "socket.io": "~1.3.6",
+    "socket.io": "~1.3.7",
     "superagent": "^0.21.0",
     "temp": "~0.8.3",
     "uuid": "~2.0.1",
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "semver": "~5.0.1",
     "serve-static": "~1.10.0",
     "signals": "~1.0.0",
-    "socket.io": "~0.9.16",
+    "socket.io": "~1.3.6",
     "superagent": "^0.21.0",
     "temp": "~0.8.3",
     "uuid": "~2.0.1",
--- a/source/config.js
+++ b/source/config.js
@@ -10,7 +10,10 @@ var defaultConfig = {
   port: 8448,
 
   // The base URL ungit will be accessible from.
-  urlBase: "http://localhost",
+  urlBase: 'http://localhost',
+  
+  // The URL root path under which ungit will be accesible.
+  rootPath: '',
 
   // Directory to output log files.
   logDirectory: null,
@@ -126,6 +129,7 @@ var argv = yargs
 .describe('cliconfigonly', 'Ignore the default configuration points and only use parameters sent on the command line')
 .describe('port', 'The port ungit is exposed on')
 .describe('urlBase', 'The base URL ungit will be accessible from')
+.describe('rootPath', 'The root path ungit will be accessible from')
 .describe('logDirectory', 'Directory to output log files')
 .describe('logRESTRequests', 'Write REST requests to the log')
 .describe('logGitCommands', 'Write git commands issued to the log')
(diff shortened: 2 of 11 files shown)


source/config.js
```diff
@@ -11,6 +11,10 @@ var defaultConfig = {
 
   // The base URL ungit will be accessible from.
   urlBase: "http://localhost",
+  
+  // The URL root path under which ungit will be accesible.
+  // Either blank or starting with a slash and no trailing slash
```
Ajedi32: Is this constraint enforced anywhere in the code? Seems like a user could easily violate this constraint without realizing it.

geeklain: I looked for other config params and there does not seem to be any validation for those.
Is there a good place to validate the config?

codingtwinky: Perhaps we can use [url.resolve()](https://nodejs.org/api/url.html#url_url_resolve_from_to)?  Not sure if we need additional validation though...

package.json
```diff
@@ -46,7 +46,7 @@
     "semver": "~5.0.1",
     "serve-static": "~1.10.0",
     "signals": "~1.0.0",
-    "socket.io": "~0.9.16",
+    "socket.io": "~1.3.6",
```
codingtwinky: I believe the reason why we didn't uptick socket.io was because 0.9.16 and 1.3.6 of socket.io didn't work well with phantom.js and screw up the click tests.  It is good to know that clicktests passes with new version.

@FredrikNoren There was no other reason to prevent us from upgrading socket.io right?

source/config.js
Ajedi32: I'm just saying that, from a UX perspective, it would be nice if ungit automatically appended a slash to the start of the config variable if needed, and removed any trailing slash. Just one less thing for the user to worry about.

Alternately, if taking care of this constraint automatically would take too much development effort, that requirement for the formatting of that setting should definitely be documented somewhere other than just a comment in the source code.

geeklain: I'll look into adding a validation for the complete url: urlBase+port+rootPath

codingtwinky: I'm not sure if I have explained correctly, if we do `url.resolve(urlBase + ':' + port, rootPath)` at the later part of this file, we don't need validation for checking prefix, postfix slashes within rootPath or port as it is handled via url package.  

What I mean by additional validation is the character encoding and non alphabet characters and such....

FredrikNoren:
@codingtwinky yeah exactly, and that's awesome that if it works now! We should remove the holdback code in the gruntfile

@geeklain this is looking awesome. only thing I'm missing is a clicktest for this as well (see how for instance the authentication test works), so that this doesn't break in the future

geeklain:
@FredrikNoren OK I'll look closer at the clicktest

geeklain:
@FredrikNoren I modified the generic clicktest to test different rootPaths.

I also added a rootPath clean up routine in config.js to cleanup leading and trainling slashes. I think that would help the users.

FredrikNoren:
@geeklain Ok awesome, looks ready to merge. Can you update the PR to solve merge conflicts and I'll merge it?

geeklain:
@FredrikNoren All clear - rebased and tests passed - ready to merge

package.json
linquize: as at today, the latest version is 1.3.7

geeklain:
oh man! Tests keep failing because at least one of them encounters a weird error... Transient errors. It's driving me nuts

codingtwinky:
Right, I have been having some issues with it as well.  I have fixed some of them in #626 but as with any intermittent issues, it is hard to verify.

geeklain:
Hurray! Tests passed without a hiccup after re-rebasing.
@codingtwinky @FredrikNoren ready to merge.

FredrikNoren:
@geeklain Nice!

sebastianmay:
We're going to use ungit on a shared environment, so we set up ungit for a few users, each with an own instance, through a reverse proxy. Everythings seems to work fine, except the favorite repositories on the home screen. I'm not quite sure, if we have set it up wrong or made some bad configurations, but we've a small fix in the home.js, which helps with this problem.
The list of favorite repositories, doesn't care about the rootPath configuration.
So we edited line 15 in components/home/home.js from
this.link = '/#/repository?path=' + encodeURIComponent(path);
to this.link = ungit.config.rootPath + '/#/repository?path=' + encodeURIComponent(path);
So far it is working nice and we can use the favorites over the reverse proxy.
Can anyone confirm this behaviour or give us a hint, what we set up wrong?
Or is this the right solution?
Thanks a lot for your help.

Ajedi32:
@sebastianmay You might want to file a new issue or submit a pull request for that. It'd be easy for a comment like that in an already-merged PR to simply go unnoticed.

codingtwinky:
@sebastianmay Yes, I do believe this is a valid issue and your fix is needed.  If you would like, please do send us a pull request if you don't mind.  Thanks!

sebastianmay:
Somehow, I'm not able create a pull request from my fork.
The only change necessary is in the file components/home/home.js line 15:
add "ungit.config.rootPath + " befor the '/#/repository'
This worked for us.

Ajedi32:
Huh. Yeah, that's weird. I was going to just send the PR for you since you've already got the commit in your fork, but GitHub's throwing a 500 error right now: https://github.com/FredrikNoren/ungit/compare/FredrikNoren:master...sebastianmay:patch-1

The relevant commit is here: https://github.com/sebastianmay/ungit/commit/2962d72e7838a68609c798f017d0efa1ed269bf8

PR #610, opened 2015-08-29
Initial native container packaging support based on electron

kwonoj:
Initial proposal of issue https://github.com/FredrikNoren/ungit/issues/593 .

I have aimed for 

* Provide capabilities to create native application container for ungit
* Completely non-invasive to current ungit codes to avoid production release's regression, until actual transition from npm based happens

bootstrapper of electron (main.js) is nearly transparently identical to npm based bootstrapper yet due to above reason. In further it's possible to consider refactoring current ungit code to more electron-friendly, when actual transition from npm module happens. (i.e, server module / page loading handling, etcs).

grunt default && grunt package will create packages under ./build, yet doesn't have way to create installer.


kwonoj:
Not sure why latest build fails only for job 919.1 with timeout..

Gruntfile.js
```diff
@@ -174,7 +175,36 @@ module.exports = function(grunt) {
           'test/**/*.js',
         ]
       }
-    }
+    },
+    copy:{
```
campersau: Minor: missing space after `copy:`

Gruntfile.js
```diff
@@ -174,7 +175,36 @@ module.exports = function(grunt) {
...
+        { expand: true, src: ['node_modules/**'], dest: 'build/resource/' },
+        { expand: true, src: ['package.json'], dest: 'build/resource/'} 
+        ]
+      }
+    },
+    clean: {
+      electron: ['./build']
+    },
+    electron: {
+	  package: {
```
campersau: Minor: use 2 spaces instead of tabs here and the following lines

FredrikNoren:
Ok this is cool as hell. Tried it on my computer and seems to work great.

Some things we'll have to sort out to make this the official "main" way ungit is shipped:
1. When starting it, at least on osx, the application icon is the electron icon [1]. Can probably use this one: https://github.com/FredrikNoren/ungit/blob/master/icon.png
2. We need a way to handle upgrade messages. Right now it will probably say "do npm install.." in the app. Might be as easy as checking if it's running in electron and just change that message to "Go to github.com/FredrikNoren/ungit to download the latest version" though. It's using npm to check what the latest version is and I think that's fine as ungit probably will need to ship through npm after we've implemented this too (to support running it on a server usecase).
3. What happens when there's no git installed on the system? There should probably at least be a warning in the UI and a way forward for users to install git. (Right now all of that checking is in the cli, before the UI).
4. Testing. I'm on OSX and it seems to work here but would love to hear someone test this out on windows and linux as well.

Hm there's probably more but that's what I could think of now. Would love @codingtwinky @Ajedi32 @campersau and the rest to take a look at this too, but great start!


[1]
screen shot 2015-08-29 at 12 48 27 pm

kwonoj:
1. Yes, I'm aware also. electron-packger (https://github.com/maxogden/electron-packager) does support icon embedding, but seems only .ico format is honored. Since first PR was aimed to 'make it work' first, I've skipped this.

2. As mentioned installer-executable package is yet created while packaging, plan is pack it via Squrriel (https://github.com/Squirrel/Squirrel.Windows / https://github.com/Squirrel/Squirrel.Mac) which original atom.io does use, which have built in auto upgrade mechanism that user even doesn't need to see those messages, would like move forward to next PR includes this.

3. For now, UI will shows 'cannot run git' but does not have redirection to install git. As starting point, I think it's fine but maybe can consider to include git installation also while creating installer script.

image

4. My testing was done against Windows environment, so someone can test on linux would be great.

Once this PR's merged, I'll start following changes for 1,2,3 for immediate tasks.

campersau:
This also works great on windows 8.1 x64 version:
image

In addition to @FredrikNoren points:
 1. I would love to run multiple instances of the ungit app.
 2. The name of the app in the taskmanager is currently `Electron`. (You can also see it in the window title during startup.)
image

kwonoj:
@campersau 
1. I do too! currently to avoid possible edge cases, I've blocked multiple instance executions. Primary focus for now is creating end-to-end package including installers to easier tryout but obviously can work on that too.

2. Yes, aware of it. currently there's two way of changing it.
: manually edit resource of prebuilt binary (http://electron.atom.io/docs/v0.31.0/tutorial/application-distribution/)
: rebuild electron binary from scratch to create custom name

both of them requires some effort and especially it's tied to windows-based build environment (first requires window to exeucte resedit, second requires VS2013 to build) I haven't mind concrete decision which way to go. Need some evaluation to make final proposal on this.

public/main.js
```diff
@@ -0,0 +1,95 @@
...
+
+app.on('window-all-closed', function() {
+  if (process.platform != 'darwin') {
+    app.quit();
+  }
+});
+
+app.on('ready', function() {
+  checkIfUngitIsRunning(function(err1, ungitRunning) {
+    if (ungitRunning) {
```
codingtwinky: If ungit is already running why not navigate to that url instead of quitting?

public/main.js
```diff
@@ -0,0 +1,95 @@
...
+  var server = net.createServer(function(c) { });
+  server.listen(config.port, function(err) {
+    server.close(function() {
+      callback(null, false);
+    });
+  });
+  server.on('error', function (e) {
+    if (e.code == 'EADDRINUSE') {
+      callback(null, true);
+    }
```
codingtwinky: What about other errors?  can we at least log it?

kwonoj:
1. Assume instance A started and following Instance B navigated to its url and then A is exited, then B become obsolete by server instance A is now terminated. For those reason current PR explicitly allows single instance before making per instance independent behavior.
2. Need to be in further. This handling comes as-is from npm bootstrapper of ungot (./bin/ungit) and I haven't modified for electron friendly. First thing I should make change is replacing logs from console output before adding additoonal statements.

codingtwinky:
So I get that subsequent processes depend on first process but am I correct in saying that user cannot open multiple tabs of ungit solely via electron?  If this is true then It's pick your poison kind of situation and I'd personally rather have multiple tabs over worrying about having first process being alive.

kwonoj:
So I get that subsequent processes depend on first process but am I correct in saying that user cannot open multiple tabs of ungit solely via electron? 
: Not sure if I understood correctly, in current PR answer to user cannot open multiple instance of electron is true. (there isn't any concept of tab in each electron instance, so I interpret it as instance). Cases I'm saying is not about first process instance, if simply redirects rest of instance to first instance's server as soon as user exist first one, all of rest becomes not working cause server relying on is no more active.

As mentioned in above, please consider this as starting point. Obviously allowing multiple instance should be way to go that I totally agree and should evaluate possible way to do it in following changes.

codingtwinky:
I understand.  Forgive me but one of the biggest reason what got me hooked on to ungit was the I can have multiple views per repo without having multiple server instances.  As a person who always runs with 4+ windows it be hard for me to see that missing...  

We might have to spawn a detached server process that kills itself upon connection count goes to 0 or something.  That way multiple frontend, browser, electron instance or whatever, instance can be served by a single server instance.

kwonoj:
I can have multiple views per repo without having multiple server instances. As a person who always runs with 4+ windows it be hard for me to see that missing
:+1: I'm totally with you.

We might have to spawn a detached server process that kills itself upon connection count goes to 0 or something. That way multiple frontend, browser, electron instance or whatever, instance can be served by a single server instance.
: That's one possible way to go. current high level option in my mind is one dedicated server (as you suggested) vs. per-instance but hasn't concluded exact way yet, reason I mentioned will evaluate further. each of them has  pros and cons.

FredrikNoren:
Hm looks like the program keeps running after I've closed it on mac, but I can't open it again. (It's just the icon in the program bar)

kwonoj:
Since not a Mac user can't say for sure, but maybe because of current bootstrapper to handle application quit as below that does not close instance? It is recommended handling, but also problematic since current PR doesn't provide way to revive window once closed by not allowing multiple instance of process. I'd like to suggest let application completely exit regardless of platform for now until changes for handling multiple instance comes in, what do you think?

// Quit when all windows are closed.
app.on('window-all-closed', function() {
  // On OS X it is common for applications and their menu bar
  // to stay active until the user quits explicitly with Cmd + Q
  if (process.platform != 'darwin') {
    app.quit();
  }
});

kwonoj:
Made changes for proposal I wrote above to let application exits completely on Mac as same as Windows does while multiple instance / window support is not in place.

FredrikNoren:
Re. one vs. several servers: I can't think of any reason to not just run one server per instance?

Multiple windows does open another small problem though: the repositories list on the home screen is saved in localStorage right now, so if you have multiple windows that would probably not sync between them (unless electron shares localStorage between windows?).

kwonoj:
Re. one vs. several servers:

In my opinion, there are lot of unknowns for those topics starting from very basic wiring to make it work to problem you have mentioned. I consider those topic requires sufficient evaluations to have separate issue / PR to make it actually happens. I had somewhat similar experience about instance strategy before (though at the moment backend was native component instead of node.js based one), it wasn't trivial. I'd like to suggest to deal those further goal to not to include in current PR and let current PR narrows down to focus initial bootstrapping, if you're ok with it.

FredrikNoren:
@kwonoj Yeah that sounds good to me. Only thing I'd want to add for this PR then is a few lines somewhere that explains how to build it as an electron app (just `grunt package` right?) and what the known limitations are right now (even better if those are linked to tickets for them).

kwonoj:
Agreed. I'll add topics for build in readme.md and update PR, then file issues separately to track further,

kwonoj:
Updated documentation, in contributing.md instead of readme.md since this PR is yet reached public release stages with lot of known limitations, mostly for developers for now. Once changes are settled down enough and having package for easier deployment (i.e, installer, get rid of known issues, etcs) then document can be updated and readme.md can have topics to instruct it.

CONTRIBUTING.md
```diff
@@ -30,6 +30,21 @@ To get started developing on Ungit:
  4. Run `grunt` to build (compile templates, css and js).
  5. Type `npm start` to start ungit, or `npm test` to run tests.
  6. (Optional). Run `grunt watch` to automatically rebuild stuff when you change files.
+ 
+Run ungit as standalone application
+-----------------------------------
+
+To provide easier access to launch ungit, very early stage of standalone application container using [election](http://electron.atom.io/) is available.
```
FredrikNoren: election -> electron

CONTRIBUTING.md
```diff
@@ -30,6 +30,21 @@ To get started developing on Ungit:
...
+Please note this is not yet ready for public release and being developed having several known & unknown limitations.
+
+To get started:
+ 1. Follow steps in 'Getting started' to get development environment is ready.
```
FredrikNoren: "get a development environment ready"

CONTRIBUTING.md
```diff
@@ -30,6 +30,21 @@ To get started developing on Ungit:
...
+ 2. Run `grunt default && grunt package`. This'll compile latest ungit and will create standalone application package under `build/`
```
FredrikNoren: This'll -> This will
"will create _a_"

kwonoj:
Thanks, updated.

FredrikNoren:
Ok cool, I'll merge this now. Just a heads up that this will probably only live as an experiment until the issues have been addressed and we can actually start shipping Ungit as a desktop app instead. But good progress so far!

kwonoj:
Cool, thanks! Moving forward to update each known issues :)

PR #612, opened 2015-09-02
Add more tooltips

linquize:
Add more tooltips


nicolaiskogheim:
This is nice. Can we have ha tooltip for stash entries saying "Apply this stash"? I have since long lost the count of how many times I have clicked a stash entry, expecting to see the patch, just to see it get pop'ed off. If you don't know what was in the stash, this can be a real pain. And probably if you do know as well.

FredrikNoren:
Nice! @nicolaiskogheim feel free to add a PR

PR #613, opened 2015-09-03
Make the color of version number string easier to read

linquize:
Make the color of version number string easier to read


linquize:
ping

PR #616, opened 2015-09-09
update package configuration for electron package description

kwonoj:
This PR updates package to have description other than default (electron).

Basically there are 2 ways to update package description
* Modify prebuilt binaries resource (http://electron.atom.io/docs/v0.31.0/tutorial/application-distribution/#rebranding-with-downloaded-binaries) 
: way this PR follows. In case of windows binary, it uses rcedit (https://github.com/atom/rcedit) to compiled binary.

* Rebuild binary from source with description (http://electron.atom.io/docs/v0.31.0/tutorial/application-distribution/#rebranding-by-rebuilding-electron-from-source)
: As briefly mentioned in initial PR, rebuilding binary was avoided due to requires windows platform (with visual studio installed) also increases build time a lot to create package. 

Since rcedit to edit windows binary is native win32 application, if you're packaging on non-windows platform and want to create windows binary with updated description, wine(https://www.winehq.org/) need to be installed. If wine is not installed windows package will have default description instead.


Gruntfile.js
```diff
@@ -204,11 +204,20 @@ module.exports = function(grunt) {
           name: 'ungit',
           dir: './build/resource',
           out: './build',
+          icon: './icon.ico',
```
kwonoj: electron-packager does not accept .png format as application icon unfortunately.

FredrikNoren:
Nice, merging!

kwonoj:
Thanks, moving forward to next goals!

PR #617, opened 2015-09-09
Remove css font IE fix to fix grunt warning

linquize:
Remove css font IE fix to fix grunt warning, as ungit does not support IE below 9


linquize:
ping

PR #623, opened 2015-10-02
git 2.6.0 does not allow hunk with no change

linquize:
git 2.6.0 does not allow hunk with no change

Fix test failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,15 @@
-sudo: false
+sudo: required
 language: node_js
 node_js:
   - "0.10"
   - "0.12"
   - "iojs"
+env:
+  - GIT_VERSION=default
+  - GIT_VERSION=edge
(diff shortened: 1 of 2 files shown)


codingtwinky:
According to [migration guide for travis container](http://docs.travis-ci.com/user/migrating-from-legacy/), it seems that there is no way to allow `sudo` in travis container.  I guess it's something we would have to keep our eyes out for.

linquize:
Since a build matrix is added (that `apt-get install git`), I have to use traditional VM, not container-based.

codingtwinky:
you are absolutely right, I'm hoping travis would change to allow sudo in containers in near future.

linquize:
Travis `sudo` attribute name is strange. If you need to use sudo, it will give you a VM. If not needed, it will give you a container.

PR #625, opened 2015-10-06
Fix intermittent test failures

codingtwinky:
At least for me, intermittent test failures follows after shutdown operation.  And it is possible that ungit's git fetch operation is lagging behind and result in 'git-error' and result in false negative test run.

This was super annoying to track down and confirm but it seems this fix does it.


FredrikNoren:
Sweet, yeah was bothering me too. Nice to get it fixed

Issue #606, opened 2015-08-06
Using css animations


codingtwinky:
Currently, ungit does manual animation via js, [edges](https://github.com/FredrikNoren/ungit/blob/master/components/graph/graph-graphics/edge.js#L27) and [nodes](https://github.com/FredrikNoren/ungit/blob/master/components/graph/graph-graphics/node.js).

It is my understanding that it would be much more efficient to do animations via css.  As a none front end dv, I want to make sure if this is a good idea and there is no objections against it.

cc @FredrikNoren

FredrikNoren:
Honestly I don't remember all of the specifics of how the animations are done. It might be worth a try, if it's possible to do I definitely agree it's preferable to js animations

nicolaiskogheim:
I don't know much about this tool, but I know it's fast. It is as fast as CSS but should be much easier to work with.

Check it out if you've got the time: [http://greensock.com/gsap](http://greensock.com/gsap)

"Cage match" between GSAP and CSS: [https://greensock.com/transitions/](https://greensock.com/transitions/)

codingtwinky:
So I haven't had good experiences with any softwares that offers both free version and licensed version at the same time, possible exception of intelliJ.  Not only that, gsap or famo.us isn't too convincing unless one were to do elaborate animations to me.

Ajedi32:
I've heard that for some more complex animations, certain JS libraries can actually perform _better_ than CSS by doing optimizations (like recycling existing objects) that native CSS animations can't do.

For simple transitions though CSS is really, really easy to use (in fact, you can pretty much just ignore it in your JS code most of the time) and [performs quite well](http://www.html5rocks.com/en/tutorials/speed/high-performance-animations/).

codingtwinky:
My understanding is same as @Ajedi32 as well.  I personally don't see ungit using advanced animations that would be excelled by these js libraries, I think it is a best to stick with css.

PR #630, opened 2015-10-15
SVG Animation rewrite

codingtwinky:
related: #606 #586

Yet another massive changes by me.  Honestly, I'm not sure how I feel about this changes as it is not critical and it is a "nice to have" 

## Goals:
1. **Better animation method**: Currently, ungit handles animation in an [javascript event loop](https://github.com/FredrikNoren/ungit/blob/c36b9e7d85ab86097bae11af63c077b96d0fe0f0/public/source/main.js#L162-L169) that calls [updateAnimationFrame](https://github.com/FredrikNoren/ungit/blob/c36b9e7d85ab86097bae11af63c077b96d0fe0f0/components/graph/graph-graphics/edge.js#L27-L45) in a perpetual loop.  Javascript handles is surprisingly well but it bring couple of issues.  Doing too many unnecessary calculations and inflexibility of animation.
2. **Remove unnecessary div layer**: This was little bit confusing, but each git node click actions are not binded to SVG circle node.  It is actually binded to a div that is overlaid over SVG circle.  So animation logic would animate both overlaid "click" div and underlying SVG.  For every git nodes we load, we essentially create 2n doms.  
3. **Refactor**: One can always use refactoring.  i.e. [actions.js](https://github.com/FredrikNoren/ungit/blob/c36b9e7d85ab86097bae11af63c077b96d0fe0f0/components/graph/graph-graphics/actions.js) is actually animation logic on hover and more naming and logic simplifications.

## Complication~~s~~:
First approach to goal 1 was via CSS.  After all, it is the goto method for animation in web.  And one would think that CSS is fully ready to animate SVG, considering that there are even examples of [CSS lava lamp](http://codepen.io/chrisgannon/pen/QjKXXe) using animated gradients and all this [cool CSS SVG animation examples](https://jakearchibald.com/2013/animated-line-drawing-svg/) right?  Wrong, **[you cannot animate SVG path or line coordinates using CSS](https://stackoverflow.com/questions/32729752/css-transition-animation-doesnt-work-on-svg-paths-d-attribute-change/32730094?noredirect=1#comment53302659_32730094)**.  Which was a stunner for me.  How do you have a lines wiggle around dynamic path but don't have a simple basic x y coordinate transition?  

There are two other ways to animate SVG: SMIL or Javascript since CSS is inadequate.  SMIL didn't work too nicely with live bindings of KO in some browsers during my quick experiment.  Not to mention doing additional dom and being ugly.

Which means javascript is the only option to animate SVG lines.  Which also means I'm stuck with very thing that I hoped to replace in the first place.  :rage:

## Result:
Ended up using http://snapsvg.io/ and use it as SVG animator.  Could have lighter, simplified javascript animation logic but Snap seems to be light enough and it was rather simple to use so pros and cons seems to be at it's favor.  

It is interesting to note that SVG element creation is [done via KO](https://github.com/codingtwinky/ungit/blob/d21c8647ac2e33eb62f90250b3c7763f8f0b9247/components/graph/graph-graphics.html#L38) but only the [animations are handled by snap](https://github.com/codingtwinky/ungit/blob/d21c8647ac2e33eb62f90250b3c7763f8f0b9247/components/graph/animateable.js).  So it has very minimal dependency to snap and maintains previous logics to parse git node elements.

So basically this pull request adds no obvious advantages, other then reduced line count and usage of SVG animation framework with fun little animation, yet terrifying rewrite of entire presentation layer.

I would love to hear whether this is worth it or not: @FredrikNoren @campersau @Ajedi32.  
Because honestly, I'm not sure and I wrote the darn thing.


Ajedi32:
> 698 additions and 1,199 deletions

Nice. :+1:

Honestly I'm no expert on the internals of ungit, but if you think it increases performance and simplifies the underlying code then it definitely seems like an overall improvement to me.

codingtwinky:
So during very informal performance testing, it seems that performance degradation between this pr and prior one is about the same.  I can't claim performance improvement.

public/source/server.js
```diff
@@ -96,7 +97,18 @@ Server.prototype._getCredentials = function(callback) {
 Server.prototype.watchRepository = function(repositoryPath, callback) {
   this.socket.emit('watch', { path: repositoryPath }, callback);
 };
-
+Server.prototype.queryPromise = function(type, url, arg) {
```
campersau: Maybe implement
`getPromise`
`postPromise`
`delPromise`
or make `query` to return a promise with backward compatibility for `callback`?

components/graph/git-graph-actions.js
```diff
@@ -243,22 +233,27 @@ GraphActions.Checkout.prototype.icon = 'glyphicon-folder-open';
 GraphActions.Checkout.prototype.perform = function(callback) {
   var self = this;
   var context = this.graph.currentActionContext();
-  var refName;
-  if (context instanceof RefViewModel) refName = context.refName;
-  else refName = context.sha1;
+  var refName = context instanceof RefViewModel ? context.refName : context.sha1 ;
```
campersau: Minor: space between `context.sha1 ;`

components/graph/git-node.js
```diff
@@ -177,52 +154,45 @@ GitNodeViewModel.prototype.showBranchingForm = function() {
...
+      var newRef = self.graph.getRef('tag: refs/tags/' + self.newBranchName());
+      newRef.node(self);
+      self.branchesAndLocalTags.push(newRef);
+    }).finally(function() {
+      self.branchingFormVisible(false);
+      self.newBranchName('');
+    });
 }
 GitNodeViewModel.prototype.toggleSelected = function() {
+  console.log(this);
```
campersau: Remove this.

Ajedi32:
Ah, well if it's mostly just a refactoring then I don't really have any opinion on this. "698 additions and 1,199 deletions" still does sound pretty attractive to me though. :smile: (Though obviously I realize that only gives a rough idea of the content of this PR.)

I'll leave it up to those of you more familiar with ungit's codebase to decide whether you think this refactoring is worthwhile.

codingtwinky:
Long story short, did a lot of performance and memory optimizations...  :P

test steps, 
1. load a repo and take a memory snap shot
2. 375 nodes and take a memory snap shot
3. navigate to default page, (should have lost all 375 node doms)
4. load a repo again and take a memory snap shot

#### before this PR
screen shot 2015-10-21 at 7 48 24 pm

#### PR without performance optimizations
screen shot 2015-10-21 at 7 48 12 pm

#### PR with performance optimizations
screen shot 2015-10-21 at 7 48 30 pm

codingtwinky:
After days of usage and it seems to be holding quite well.  Along with additional memory and speed enhancement it is quite nice to see little fun animations.

FredrikNoren:
@codingtwinky Very cool! And well needed, the animation system grew quite complicated and hard to work with. And 509 less lines of code to maintain is always a win :D

PR #637, opened 2015-10-29
Authentication Dialog: Display username if embedded in URL

linquize:
Authentication Dialog: Display username if embedded in remote URL

For example: `git remote add origin you@example.com`

Then it should fill username `you` in the Dialog, you only need to type the password.

### DIFF
--- a/test/spec.credentials-helper.js
+++ b/test/spec.credentials-helper.js
@@ -11,7 +11,7 @@ describe('credentials-helper', function () {
 		var socketId = Math.floor(Math.random() * 1000);
 		var payload = { username: 'testuser', password: 'testpassword' };
 		var server = http.createServer(function (req, res) {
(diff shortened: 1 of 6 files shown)


test/spec.credentials-helper.js
```diff
@@ -27,6 +27,8 @@ describe('credentials-helper', function () {
 			server.close();
 			done();
 		});
```
codingtwinky: Shouldn't there be a test with and without the username embedded in url?

linquize:
Is current `test/spec.credentials-helper.js` enough?

codingtwinky:
But using it without user name embedded in url is still an option right?  Wouldn't that give us different code path and better coverage?

FredrikNoren:
Closing due to inactivity

Issue #635, opened 2015-10-26
GUI - its really dark


gedw99:
i don't mean to complain, but it would be good if the colours were a little brighter in the web GUI.
For example reading the upstream branch names is difficult.

codingtwinky:
Perhaps, not everybody is working in a dungeon chained to a desk like I do so low contrast UI of ungit is understandably difficult.  Below is what I did with #640 and if you have a more specific request please do let us know.

<img width="656" alt="screen shot 2015-11-21 at 3 03 30 pm 1" src="https://cloud.githubusercontent.com/assets/5281068/11320952/1c3cedfe-9061-11e5-9028-7b661cf6a170.png">

nicolaiskogheim:
I wonder if the branch names deservers buttons/labels too, or something like that. One could argue that the branch names is much more important to be able to read quickly than the actions (push, checkout, delete), but now it's the text that never changes that is easiest to read. After having seen those action-buttons once, you probably don't even read what they say.

I have a habit of coming of as grumpy, while I'm not, when I write english, so I'll just tack on a smilyface at the end here: :smiley: 

And while I'm at the keyboard: I love your work @codingtwinky!

codingtwinky:
Thanks,

But I'm not sure if I agree with branches being a button.  Although it is important, button to me implies a git action will happen in ungit's case, which isn't true for branches.  

At very list, I do believe representations of branches needs to be different from other git actions as it implies different user experiences, and since git actions are buttons, branches shouldn't be a button.

nicolaiskogheim:
I'm sorry, I don't think I managed to get my point through. I agree with you that it should not look like clicking a branch name has an action.

What I meant is making the branch name the important part. When I scan over the nodes I'm usually looking for a branch. When I know where the right branch is, I always know where the actions are, I never have to look for them, and I wouldn't have to even if the buttons were very small (not suggesting you make the buttons small, just making a point). The UI is good, I just don't agree with the current visual hierarchy regarding branch names.

gedw99:
Heyjust catching up on this.
The image with the more district colours is great I think.
It was more that the contrast was too low, and the colours certainly remedies that.

Thanks again the all the ard work

PR #640, opened 2015-11-05
Add more distinct accent to current branch and remote

codingtwinky:
fix: #635 

@gedw99 How is this one?  

I have added more font size change to currently checked out branch so it be more distinct.  

screen shot 2015-11-04 at 6 16 01 pm


gedw99:
I would prefer its more readable. Maybe I am just getting old though. :)

FredrikNoren:
@codingtwinky hm maybe try vertically centering the texts? That looks a little off to me

codingtwinky:
will be fixing v-align once I get to final aesthetic...  Believe it or not correcingt these v-align css to line up correctly were one of the hardesting I had to do in last massive UI update...  (I was so close to using "!important" on some of the css fields...  :P)

@gedw99 do you have any suggestion on how can it be more distinct?

Issue #629, opened 2015-10-13
Future testing improvements


codingtwinky:
- [ ] [root url path test](https://github.com/geeklain/ungit/blob/b1b766b3c08170b783fe66d05e64d8f2f3c1a2dc/clicktests/test.generic.js#L14) was added recently and travis run [986](https://travis-ci.org/FredrikNoren/ungit/builds/84687260) is before merge and [989](https://travis-ci.org/FredrikNoren/ungit/builds/84724572) is after merge.  It added +3 minutes, which isn't bad.  But if we can indirectly test root url path via applying root url path in other tests we can shave 3 minutes
- [ ] [submodule test is flacky](https://travis-ci.org/FredrikNoren/ungit/jobs/84687261).  It is rare but seen it 3rd time now.
- [ ] generate test coverage report
- [ ] [generate js browser compatability report](https://github.com/caniuse-js/caniuse-js)

PR #642, opened 2015-11-10
Coverage

codingtwinky:
adding unit test coverage.  coverage for click test needs better instrumentation and will submit as a separate pr.

It only checks for `./source` as unit test are focused in there but out of this we can see that `source/utils/blockable.js` is not being used and be removed. 

relate #629


campersau:
Very nice! Here are my current coverage stats:

```
=============================== Coverage summary ===============================
Statements   : 76.96% ( 862/1120 )
Branches     : 63.94% ( 289/452 )
Functions    : 72.73% ( 176/242 )
Lines        : 78.48% ( 824/1050 )
================================================================================
```
image

FredrikNoren:
@codingtwinky very cool!

PR #646, opened 2015-11-15
Change commit ordering to date-order.

### LINKED ISSUE #641: Branch several commits behind master is filtered out
If you have a feature branch that's a bit behind master, the commits will be filtered out even if you currently have the branch checked out. Screenshot to demonstrate:

bug

keiter:
Fixes #641

### DIFF
--- a/source/git-api.js
+++ b/source/git-api.js
@@ -273,7 +273,7 @@ exports.registerApi = function(env) {
   app.get(exports.pathPrefix + '/log', ensureAuthenticated, ensurePathExists, function(req, res){
     var limit = '';
     if (req.query.limit) limit = '--max-count=' + req.query.limit;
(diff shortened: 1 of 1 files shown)


keiter:
Do you guys have a strong opinion on `topo-order` vs. `date-order` as a nonoptional default? I'd be fine with a configurable option as well, but `topo-order` really only barely works in our workflow as described in the issue.

codingtwinky:
I personally don't see huge issues with this.  However, order of nodes are important on how we color and displayed.  So may need more burn time.  

have you noticed any other side effects?  I'd imagine it be fine and will test on my own later.

keiter:
I'm using the patched version at work, and it appears to work well thus far. Coloring is working as it did previously AFAIK. The difference is mostly in the layout of the graph, as the graph nodes are laid out in order of commit date you can get them "interleaved" when there is paralell work going on. Screenshot:

date-order

codingtwinky:
It looks good to me and I think it makes more sense.

nicolaiskogheim:
This looks really neat. :+1:

PR #654, opened 2015-12-01
Fix stash bracket escape problem for windows

codingtwinky:
fixes: https://github.com/codingtwinky/atom-ungit/issues/43

Now I'm wondering if this works in other versions of windows because I feel like other windows users should have seen this issue.  

it would be great if someone can confirm this is an issue or this fix will create an issue for them.

### DIFF
--- a/source/config.js
+++ b/source/config.js
@@ -192,5 +192,6 @@ if (argv.$0 === 'ungit' && argv._ && argv._.length > 0) {
   module.exports = rc('ungit', argv.default(defaultConfig).argv);
 }
 module.exports.homedir = homedir;
+module.exports.isWindows = /^win/.test(process.platform);
 
 cleanUpRootPath();
--- a/source/git-api.js
+++ b/source/git-api.js
@@ -618,9 +618,9 @@ exports.registerApi = function(env) {
   });
 
   app.delete(exports.pathPrefix + '/stashes/:id', ensureAuthenticated, ensurePathExists, function(req, res){
-    var type = 'drop';
-    if (req.query.pop === 'true') type = 'pop';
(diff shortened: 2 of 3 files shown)


campersau:
Hmm.. for me this fix does not work.

Command:
```
-c color.ui=false -c core.quotepath=false -c core.pager=cat stash pop stash@\{0\}
```

Error:
```
stash@\{0\} is not a valid reference
```

Changed code:
```js
git(['stash', type, 'stash@\\{' + req.params['id'] + '\\}'], req.query['path'])
```

The code before this change worked just fine.

```
$ ver
Microsoft Windows [Version 10.0.10586]

$ git --version
git version 2.6.3.windows.1

ungit@0.9.3
```

Ajedi32:
That's weird. Doesn't ungit now pass these arguments directly to git without running them through the shell first? So you shouldn't need to escape anything for Windows. Or did that get changed back? I don't remember.

codingtwinky:
How we execute the command [hasn't changed](https://github.com/FredrikNoren/ungit/blame/master/source/git.js#L62).  There were some changes around how we are using promises but not directly on how we are executing the commands.

I would have to read up more about the `child_process.spawn()` but for some reason spawned shell is parsing passed in argument.

Ajedi32:
Yeah, that's weird. `child_process.spawn` shouldn't be spawning a shell at all (unless of course, you're explicitly telling it to execute bash or whatever, which you're not).

codingtwinky:
I have been rulling that out because it's all nodes after all and it should be same.  but it is possible that env variable is different and does something different or some specific rules set for windows or something.

PR #658, opened 2015-12-12
Using css triangles instead of images

codingtwinky:
I thought I did this having css replace PNG files


FredrikNoren:
That's neat, didn't know that was possible!

PR #659, opened 2015-12-15
Fix image diff issue #657

### LINKED ISSUE #657: Image diff doesn't work
Right now, image changed diff view will not work as [`.spawn()`](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options) ignores encoding option and file is not correctly delivered.  It reads image file as a text file and try to deliver.

screen shot 2015-12-09 at 5 21 34 pm

codingtwinky:
Thought sent this in but I guess I haven't.

screen shot 2015-12-13 at 7 57 12 pm

fixes #657

Didn't know how easy it was to stream data using pipe but I will be replacing back end to use streams where it can when I do conversions for promises.

### DIFF
--- a/source/git-api.js
+++ b/source/git-api.js
@@ -199,14 +199,9 @@ exports.registerApi = function(env) {
   });
 
   app.get(exports.pathPrefix + '/diff/image', ensureAuthenticated, ensurePathExists, function(req, res) {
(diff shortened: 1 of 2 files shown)


FredrikNoren:
Lgtm

Issue #663, opened 2015-12-28
Stable build on Windows


FredrikNoren:
Currently incredibly unstable on windows, let's use this as a tracking issue to improve things.

GammaNu:
1.1.29 still very instable on windows (at least with uBlock on)

codingtwinky:
Ad blocker is a known issue and I'm sure you know this but you can find more about it in [readme](https://github.com/FredrikNoren/ungit#known-issues).  

For windows, please file individual tickets for each issues and will try to take a look at it.  But considering I don't have windows machine, it can be difficult.  I would welcome your PR though, should you want to fix on your own

campersau:
Builds on windows are finally stable https://github.com/FredrikNoren/ungit/pull/1263

PR #664, opened 2015-12-28
Improve clicktests #663

FredrikNoren:
Various fixes to get them more stable

### DIFF
--- a/clicktests/helpers.js
+++ b/clicktests/helpers.js
@@ -4,34 +4,66 @@ helpers.log = function(text) {
   console.log((new Date()).toISOString(), text);
 }
 
-helpers.waitForElement = function(page, selector, callback) {
+helpers.elementExists = function(page, selector) {
+  helpers.log('Querying element exists: ' + selector);
+  var element = page.evaluate(function(selector) {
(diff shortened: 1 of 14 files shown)


clicktests/helpers.js
```diff
@@ -4,44 +4,66 @@ helpers.log = function(text) {
...
+  return element;
+}
+
+helpers.elementVisible = function(page, selector) {
+  helpers.log('Querying element visible: ' + selector);
+  var element = page.evaluate(function(selector) {
+    var element = document.querySelector(selector);
+    if (!element) return null;
+    var rect = element.getBoundingClientRect();
+    if (rect.width == 0 || rect.height == 0) return null;
```
codingtwinky: Should we add [`:visible`](https://stackoverflow.com/questions/8774089/detect-if-an-element-is-visible) to the selector?  I think that will capture when `display: none` case.

FredrikNoren:
Still flaky but at least it _can_ pass on windows now :)

@codingtwinky Just FYI I had to disable some tests to get it to work at all on windows. On my todo to get them back working at some point.

PR #667, opened 2016-01-03
Add a way to hide commit message area in staging

### LINKED ISSUE #643: too much screen space lost because of commit message inputs
I quite like ungit's UI, but there are several weak points. One of them being staging component.
I useally work on one commit few hours and all that time, 1/3 of my screen has to be filled with en empty space like here:
screenshot from 2015-11-11 18-54-03

If I could propose an improvement, I think commit message should open in a popup or a modal. So we would only have a commit button in the staging area. That button could easily be in the corner overlaying the staging window content-that way the content could take up the whole screen.

What do you think, should I try making a PR?

codingtwinky:
Fix #643

screen shot 2016-01-03 at 2 43 30 pm

screen shot 2016-01-03 at 2 45 58 pm

(Not sure if I'm a fan of how the toggle button is displayed...)


FredrikNoren:
Hm idk about this, it feels like we're getting a little too much buttons and options. If anything I could imagine changing the max width of it to be smaller (the original issue seems to be around it using too much space on a big screen).

codingtwinky:
Well, I do see that argument that it is getting crowded.  But I do see the point of this feature though.  Screen space lost due to commit message is some what annoying when there are merge conflicts to deal with.  

Maybe I should do UI differently?

codingtwinky:
@FredrikNoren Does this work?  We can bring down font size or try different icons...

screen shot 2016-01-08 at 6 37 54 pm

Issue #621, opened 2015-09-30
1.0.0!


dbkaplun:

FredrikNoren:
Yes! I want that to be #593 though, that @kwonoj is working on. Basically packaging ungit as an electron app.

codingtwinky:
I think this would be very nice.  One quick thought though, can we do one or couple more releases prior so we can have more battle test #593 and #602?

These are couple of big features that may need more burn time in the user's hands.

FredrikNoren:
@codingtwinky Yup I think that makes sense, we should probably feature freeze once that is out and focus only on the electron release and have that really polished.

linquize:
1.0 means plugin API stable?

Ajedi32:
@linquize Good point. If we're following SemVer, that would be what a 1.0 release would mean. I think most people here though seem to be thinking of the version number from a much more squishy "user's perspective", rather than from the perspective of the Ungit plugin API.

FredrikNoren:
Yeah I feel like the plugin system never really caught on so almost
thinking we should remove it for 1.0. Complicated to support properly and
adds unknowns with electron. Can keep the code but no "official" support
(remove from docs etc.). (Unless it turns out people are using it, but
haven't seen any evidence of that so far)

Ajedi32:
Aren't some of the components in the ungit core implemented as plugins though?

FredrikNoren:
Yeah, I think the code should remain, just changing the documentation to not "encourage" plugins

Issue #669, opened 2016-01-08
es6?


codingtwinky:
Thoughts on [es6](https://nodejs.org/en/docs/es6/)?

I don't think moving client side to es6 is a good idea without babel or something of sort as support on it is very spotty, but I think it make sense for serverside.  I'm a particularly a big fan of arrow functions, map, weakmap, and string templates personally...

(I figure I will continue my strict of continuous refactoring without adding any values...  :P)

Ajedi32:
Yep, ES6 is awesome. In addition to the features you mentioned, I'm also a fan of object destructuring, the class sugar syntax, and async/await (though that's actually part of ES7 I believe).

Babel is also really nice since it frees you from having to constantly worry about backwards compatibility. Even Node doesn't yet support all ES6 features.

FredrikNoren:
I'm all for it, don't mind dropping es5 support on the server side. And
client side with Babel sounds good.

codingtwinky:
Hooray, I think I will dedicate this issue as doing es6 for serverside.  V8 shouldn't need any Babel.  May create separate tickets for doing es6 for client side later.  

Maybe wrap clientside es6 effort and promisify together...

codingtwinky:
Closing as serverside es6 is implemented

PR #672, opened 2016-01-09
Enable es6

codingtwinky:
Simply enables es6 compilation for selective serverside codes.

updated: 16/1/21

* server side codes in `.../source` get compiled into `.../src` via babel dynamically based on the node env's es6 compatibility.
* coverage job uses compiled `.../src`
* For electron, `.../source` will be compiled with static preset with [es2015](https://babeljs.io/docs/plugins/preset-es2015/) and [stage-0](https://babeljs.io/docs/plugins/preset-stage-0/).  (which I think is conservative enough...)
* `.../source/utils/*`, `.../source/bugtracker.js` and `.../source/config.js` are refactored with es6 as an experiment.

It maybe a good idea to merge this after [ungit@1.0.0](https://github.com/FredrikNoren/ungit/issues/621) release.


codingtwinky:
So it seems that this will kill node 0.10 and 0.12 is not liking es6 syntax as expected.

Are we okay with dropping the support for this?  I know node is dropping support for 0.10 soon but I'm not sure on the time line of 0.12.

Ajedi32:
@codingtwinky Here's the timeline: https://raw.githubusercontent.com/nodejs/LTS/master/schedule.png 0.12.x support will be dropped in December.

And yeah, like I said in #669, not all ES6 features are supported by Node yet. Even Node 5 is still missing support for some features: https://kangax.github.io/compat-table/es6/#node5

If you want full ES6 compatibility, definitely go with Babel.

codingtwinky:
It was my intention to use raw es6 on the server side without babel as it does have some nice performance efficiency and drop support for some of the older ones. 

Is that too aggressive?

Ajedi32:
I personally have no problem with dropping Node 0.10 and 0.12 support. Though I imagine that might cause minor headaches for some users who will need to upgrade to a newer node in order to keep using ungit.

If performance is your concern, one possibility is to just use the new native ES6 features in your code, then use Babel to degrade gracefully on older node versions. You could use [`babel-features`](https://github.com/hax/babel-features) to detect what features you're using aren't available on the current Node version, and automatically compile those features at runtime with [`babel-register`](https://babeljs.io/docs/usage/require/).

codingtwinky:
That is so cool!  I didn't know about that babel could do that!  I will be experimenting with those features.

codingtwinky:
it is ready for review.  As this changes A LOT, I would love to get your thoughts and blessings...  

cc: @FredrikNoren

FredrikNoren:
@codingtwinky Looks reasonable to me, guess this just enables us to write es6 classes (in the backend only?) but only changes one class to use es6? The one thing: Do we really need to babel stuff? Or could we just drop support for old nodejs versions instead?

codingtwinky:
@FredrikNoren You are correct, but converted classes are more than just one: `.../source/utils/*`, `.../source/bugtracker.js` and `.../source/config.js` as a trial.  Once this get merged in, we need to slowly convert serverside codes to use es6.

Yes, it maybe adding too much dependencies of Babel plugins and etc now that I'm looking back, which is [terrifying](https://github.com/azer/left-pad/issues/4).  I'm all for dropping support for old node and not doing this dynamic es6 compatibility checked compilation.  I can change it so that it is compiled using Babel with most conservative setting that will work in even the environments without any ES6 support (which I'd imagine we would have to do for client side).  That will allow us to drop most of the babel plugins dependencies.

FYI, here is the terrifying list of `npm list | grep babel` as of now.
```
├─┬ babel-preset-es2015@6.3.13
│ ├─┬ babel-plugin-check-es2015-constants@6.5.0
│ │ └─┬ babel-runtime@5.8.35
│ ├── babel-plugin-transform-es2015-arrow-functions@6.5.2
│ ├── babel-plugin-transform-es2015-block-scoped-functions@6.5.0
│ ├─┬ babel-plugin-transform-es2015-block-scoping@6.5.0
│ │ ├─┬ babel-template@6.5.0
│ │ ├─┬ babel-traverse@6.5.0
│ │ ├─┬ babel-types@6.5.2
│ ├─┬ babel-plugin-transform-es2015-classes@6.5.2
│ │ ├─┬ babel-helper-define-map@6.5.0
│ │ ├── babel-helper-function-name@6.5.0
│ │ ├── babel-helper-optimise-call-expression@6.5.0
│ │ ├── babel-helper-replace-supers@6.5.0
│ │ └── babel-messages@6.5.0
│ ├── babel-plugin-transform-es2015-computed-properties@6.5.2
│ ├── babel-plugin-transform-es2015-destructuring@6.5.0
│ ├── babel-plugin-transform-es2015-for-of@6.5.2
│ ├── babel-plugin-transform-es2015-function-name@6.5.0
│ ├── babel-plugin-transform-es2015-literals@6.5.0
│ ├─┬ babel-plugin-transform-es2015-modules-commonjs@6.5.2
│ │ └── babel-plugin-transform-strict-mode@6.5.2
│ ├── babel-plugin-transform-es2015-object-super@6.5.0
│ ├─┬ babel-plugin-transform-es2015-parameters@6.5.0
│ │ ├── babel-helper-call-delegate@6.5.0
│ │ └── babel-helper-get-function-arity@6.5.0
│ ├── babel-plugin-transform-es2015-shorthand-properties@6.5.0
│ ├── babel-plugin-transform-es2015-spread@6.5.2
│ ├─┬ babel-plugin-transform-es2015-sticky-regex@6.5.0
│ │ └─┬ babel-helper-regex@6.5.0
│ ├── babel-plugin-transform-es2015-template-literals@6.5.2
│ ├── babel-plugin-transform-es2015-typeof-symbol@6.5.0
│ ├─┬ babel-plugin-transform-es2015-unicode-regex@6.5.0
│ └─┬ babel-plugin-transform-regenerator@6.5.2
│   ├── babel-plugin-syntax-async-functions@6.5.0
├─┬ babel-preset-features@2.0.5
│ ├── babel-features@2.0.0-beta
│ ├── babel-plugin-jscript@1.0.4
│ ├─┬ babel-plugin-transform-async-to-generator@6.5.0
│ │ └── babel-helper-remap-async-to-generator@6.5.0
│ ├── babel-plugin-transform-es2015-constants@6.1.4
│ ├── babel-plugin-transform-es2015-modules-amd@6.5.0
│ ├─┬ babel-plugin-transform-es2015-modules-systemjs@6.5.0
│ │ └── babel-helper-hoist-variables@6.5.0
│ ├── babel-plugin-transform-es2015-modules-umd@6.5.0
│ ├── babel-plugin-transform-es3-member-expression-literals@6.5.0
│ ├── babel-plugin-transform-es3-property-literals@6.5.0
│ ├── babel-plugin-transform-es5-property-mutators@6.5.0
│ └── babel-plugin-transform-jscript@6.5.0
├─┬ babel-preset-stage-0@6.3.13
│ ├─┬ babel-plugin-transform-do-expressions@6.5.0
│ │ └── babel-plugin-syntax-do-expressions@6.5.0
│ ├─┬ babel-plugin-transform-function-bind@6.5.2
│ │ └── babel-plugin-syntax-function-bind@6.5.0
│ └─┬ babel-preset-stage-1@6.5.0
│   ├─┬ babel-plugin-transform-class-constructor-call@6.5.0
│   │ └── babel-plugin-syntax-class-constructor-call@6.5.0
│   ├─┬ babel-plugin-transform-class-properties@6.5.2
│   │ └── babel-plugin-syntax-class-properties@6.5.0
│   ├─┬ babel-plugin-transform-decorators@6.5.0
│   │ ├─┬ babel-helper-explode-class@6.5.0
│   │ │ └── babel-helper-bindify-decorators@6.5.0
│   │ └── babel-plugin-syntax-decorators@6.5.0
│   ├─┬ babel-plugin-transform-export-extensions@6.5.0
│   │ └── babel-plugin-syntax-export-extensions@6.5.0
│   └─┬ babel-preset-stage-2@6.5.0
│     ├── babel-plugin-syntax-trailing-function-commas@6.5.0
│     ├─┬ babel-plugin-transform-object-rest-spread@6.5.0
│     │ └── babel-plugin-syntax-object-rest-spread@6.5.0
│     └─┬ babel-preset-stage-3@6.5.0
│       └─┬ babel-plugin-transform-exponentiation-operator@6.5.0
│         ├─┬ babel-helper-builder-binary-assignment-operator-visitor@6.5.0
│         │ └── babel-helper-explode-assignable-expression@6.5.0
│         └── babel-plugin-syntax-exponentiation-operator@6.5.0
├─┬ grunt-babel@6.0.0
│ └─┬ babel-core@6.5.2
│   ├─┬ babel-code-frame@6.5.0
│   ├─┬ babel-generator@6.5.0
│   ├── babel-helpers@6.5.0
│   ├─┬ babel-register@6.5.2

```

Ajedi32:
Yeah, usually Babel is only used in development, and the compiled files are used in production.

I know you were concerned about the performance costs of transpiled Babel code, but in all honesty I kinda suspect that doing the transpilation on the fly would actually be slower than just using transpiled code (even for features natively supported in newer node versions).

I still think Babel is still probably a good idea, because for some of the newer ES6 features even the latest versions of Node don't support them natively, but it's probably best to just do the transpilation in development.

FredrikNoren:
@codingtwinky Ah I see, yeah if we're transpiling we should probably just do it before uploading to npm (i.e. what @Ajedi32 said), I guess there's not much downside to running the transpiled code everywhere (except the maybe performance but I don't think we need to worry)?

Generally anything that makes the installation process flakier I would advice against, since I see so much people having trouble with the installation. Anything that can fail we should remove.

codingtwinky:
I agree with those wise words.

I will remove this dynamic environment dependent babel transpiling, although I think it's supa cewl I agree that it is bit risky and not worth it.

codingtwinky:
@FredrikNoren @Ajedi32 @campersau 

Now grunt job statically uses `['es2015', 'stage-0']` presets which is conservative to for all node versions and it should work for client side as well when we begin using babel for clientside.

campersau:
:+1:

codingtwinky:
Merging it for now but please do post any and all comment so we can fix in other PRs.

PR #680, opened 2016-01-22
dependency bump

codingtwinky:
* :up: 
* fixes issues when diff line gets pushed down like below.

screen shot 2016-01-21 at 8 58 22 pm


FredrikNoren:
:+1: 

stupid windows build that keeps failing..

Issue #686, opened 2016-02-01
[v0.10.0]scroll down on safari till the start node.


kaizvn:
I browse `ungit` newest version 0.10.0 on my Safari (v.9.0.3, OSX 10.11.3) and it keep scrolling down till the start node. My project which have been developing in 6 months, it took a very long time to finish fetching ( I interrupted it instead).
Luckily this behavior doesn't happen on Chrome, please check.

FredrikNoren:
I'm seeing that too. It's automatically scrolling down for some reason, weird.

FredrikNoren:
I just tried 0.9.3 and that seems broken too in Safari (also jumping down, though it stops but that seems like it's because it fails to load more nodes for some reason).

Also tried overriding the scroll methods on window to try to find where in the code it happens, but they never seemed to be called. Also tried commenting out scrolls in the code and they weren't called either, so wondering if this is some weird css issue

nicolaiskogheim:
Tried bisecting?

FredrikNoren:
@nicolaiskogheim Nope only went back to 0.9.3, but yeah that would probably be the easiest way to find what's wrong

codingtwinky:
Reopening as wrong tag was referenced

PR #692, opened 2016-02-02
fix remote tag duplicate problem

codingtwinky:
fix: #686 

git ref bug fixes


FredrikNoren:
:+1:

FredrikNoren:
Same comment here now that there's a change log: we should add it there as part of the pr!

Issue #688, opened 2016-02-01
New UI hover effects


stu177:
The new hover effects in 0.10.0 to me seem to hinder usability. I found the buttons before much easier to use and navigate. Maybe it's just me but wonder if other people feel the same.

barbarah:
I'm also not a great fan of the new UI hover effects. For me the biggest problem is that I cannot see to  which fork I'm pushing to. The fork you push to is the fork selected in the top right, when I hover over the push button the fork name in the top right is hidden. Furthermore the new effects feel a bit messy/chaotic. I don't think it's an improvement.

campersau:
For reference: This behaviour got introduced with https://github.com/FredrikNoren/ungit/pull/550

codingtwinky:
Hmm... I do apologize, and intention was to reduce the overall width of the repository-buttons group's width.    I do agree that having target repo displayed all the time is helpful as push op target is depends on that.  

But do you feel that it is messy/chaotic due to timing of the expansion and reduction or do you not like the this iconized effort at all?

chrisuehlinger:
I'm a big fan of this tool and recommend it to everyone at work (including designers and other people who are not hardcore developers). I recommend it based on its user-friendliness, but also because it takes the words people have heard ("I'll **rebase** that tomorrow", "We need more review before we **merge**", "I **push**ed that on Tuesday") and puts them in a graphical context they can understand. With the words hidden behind hover effects, the value of ungit as a teaching tool is noticeably hindered, and there isn't another tool I can recommend to people instead.

I also love ungit because of its attention to the mobile experience. Hover effects won't work on my iPad, and so if I pull up a repo on my iPad to show someone, they need to already know what the icons mean in order to understand what I'm showing them.

If you guys really feel like there's a need for the hover effects, can you at least add a config option to turn them off (or even make them an opt-in for power users)?

daedalus28:
I am also a huge fan of the tool, and with this update for `patch` support I think it's literally the best git client, period. There's nothing left that I miss from SourceTree or any other client I've used before (except maybe first class support for `git-flow`). 

With that said, I'm afraid I need to second the sentiment regarding the hover effects. In general, "mystery meat" icon navigation only works when the icons are both immediately obvious as to what they represent and also ideally have some precedent in the user's mind for what they ought to do. In the case of ungit, much of the value lies in its accessibility when users may not fully understand what's going on (for example, I've never fully understood rebase until I saw it so clearly laid out in ungit).

[...]

PR #693, opened 2016-02-03
Removing hover effect

codingtwinky:
Removing hover effect

fix: #688

screen shot 2016-02-02 at 5 59 44 pm


FredrikNoren:
@codingtwinky let's add this to the change log too :) other than that :+1:

PR #694, opened 2016-02-03
Fix git hooks failing on non-ascii files

### LINKED ISSUE #676: LANG and LC_ALL environment variables cause issues with git hooks
There are a bunch of git hooks that read the locale from the environment variables.  Ungit has the environment set to `env: { LANG: 'en_US', LC_ALL: 'C' }`.  The `LC_ALL` is ascii here which breaks hooks running on unicode files.

bakkerthehacker:
Fixes #676


FredrikNoren:
lgtm. @codingtwinky you good with this?

codingtwinky:
:+1: it works on my end and I don't see the problem.

PR #696, opened 2016-02-07
bump

codingtwinky:
:up:

* includes security [fix](https://nodesecurity.io/advisories/moment_regular-expression-denial-of-service) in moment.js
* skipping phantomjs upgrade to 2.x.x from 1.9.x as it causes some failures...  not to mention it being very slow


FredrikNoren:
:+1:

PR #699, opened 2016-02-07
removing autofocus tag for safari

codingtwinky:
fix: #686 

in fact, `autofocus = "false"` still causes scroll effect on safari.  

(reason number 528412 for why front end dev is not for me...)


FredrikNoren:
Wow, can't believe you found this.

Only thing I'm wondering is: should we do the auto-focusing in js to keep the behaviour?

codingtwinky:
I was really surprised myself, I thought that part of dom is not loaded until `branchingFormVisible := true` so it shouldn't have affected it.  I guess way vastly ko binding works is different for safari?

Will do a hack to work around this later after...

codingtwinky:
What was the purpose of autofocus?  I see that text box is highlighted on expand in chrome, firefox and safari automatically.  If this is the case we wouldn't need it right?  

I really don't want to downgrade browsers to find out where it breaks and implements javascript equivalent of this...

FredrikNoren:
Oh ok yeah if that's the case then we don't need to do anything, good to merge

Issue #685, opened 2016-02-01
Double tag issue


FredrikNoren:
Getting two tags from time to time:

![screen shot 2016-01-31 at 8 04 55 pm](https://cloud.githubusercontent.com/assets/1288009/12706603/c320ae8a-c856-11e5-86a0-f60bc6c5ef7e.png)

PR #700, opened 2016-02-07
Branch logic fix/refactoring

codingtwinky:
One of the cause of #685 is due to manually registering created git-ref to git-node and it was bit confusing.

We are now only using the logic that automatically associates git-ref to git-node bi-directionally when git-node is assigned to git-ref and not manually configuring association.

(same as #698 but during testing I have delete the branch and it got messy...  (╯°□°）╯︵ ┻━┻  )

### DIFF
--- a/components/graph/git-node.js
+++ b/components/graph/git-node.js
@@ -140,14 +140,10 @@ GitNodeViewModel.prototype.setData = function(logEntry) {
   this.commitComponent.setData(logEntry);
 
   if (logEntry.refs) {
-    var refVMs = logEntry.refs.map(function(ref) {
-      var refViewModel = self.graph.getRef(ref);
-      refViewModel.node(self);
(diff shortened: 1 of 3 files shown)


components/graph/git-node.js
```diff
@@ -140,14 +140,10 @@ GitNodeViewModel.prototype.setData = function(logEntry) {
   this.commitComponent.setData(logEntry);
 
   if (logEntry.refs) {
-    var refVMs = logEntry.refs.map(function(ref) {
-      var refViewModel = self.graph.getRef(ref);
-      refViewModel.node(self);
-      return refViewModel;
+    logEntry.refs.map(function(ref) {
```
campersau: If you do not return anything inside the `map` callback you can use `forEach` which does not create an array.

PR #703, opened 2016-02-12
Serverside error handling improvement

codingtwinky:
* Properly handling few errors that are not really errors
* More subdued error logging in serverside
  * Not using "Unhandled" as often client side is handling it
  * Not showing stack trace to reduce humane alertness.

cc: @campersau 

```javascript
> Object.defineProperty(Error.prototype, 'toJSON', {
...   value: function() {
.....     var alt = {};
.....     Object.getOwnPropertyNames(this).forEach(function(key) {
.......       alt[key] = this[key];
.......     }, this);
.....     return alt;
.....   },
...   configurable: true
... });
[Error]
> var m = {a: 1, b: 2}
undefined
> var err = new Error('abc');
undefined
> JSON.stringify(m)
'{"a":1,"b":2}'
> JSON.stringify(err)
'{"stack":"Error: abc\\n    at repl:1:11\\n    at REPLServer.defaultEval (repl.js:252:27)\\n    at bound (domain.js:287:14)\\n    at REPLServer.runBound [as eval] (domain.js:300:12)\\n    at REPLServer.<anonymous> (repl.js:417:12)\\n    at emitOne (events.js:95:20)\\n    at REPLServer.emit (events.js:182:7)\\n    at REPLServer.Interface._onLine (readline.js:211:10)\\n    at REPLServer.Interface._line (readline.js:550:8)\\n    at REPLServer.Interface._ttyWrite (readline.js:827:14)","message":"abc"}'
> 

```

### DIFF
--- a/components/repository/repository.js
+++ b/components/repository/repository.js
@@ -60,7 +60,7 @@ RepositoryViewModel.prototype.updateAnimationFrame = function(deltaT) {
 RepositoryViewModel.prototype.refreshSubmoduleStatus = function() {
   var self = this;
   this.server.get('/baserepopath', { path: this.repoPath }, function(err, baseRepoPath) {
-    if (err) {
+    if (err || !baseRepoPath.path) {
       self.isSubmodule(false);
       return true;
     }
--- a/source/git-api.js
+++ b/source/git-api.js
@@ -414,9 +414,15 @@ exports.registerApi = function(env) {
 
   app.get(exports.pathPrefix + '/baserepopath', ensureAuthenticated, ensurePathExists, function(req, res){
     var currentPath = path.resolve(path.join(req.query.path, '..'));
-    jsonResultOrFailProm(res, gitPromise(['rev-parse', '--show-toplevel'], currentPath).then(function(baseRepoPath) {
-      return { path: path.resolve(baseRepoPath.trim()) };
-    }));
+    jsonResultOrFailProm(res, gitPromise(['rev-parse', '--show-toplevel'], currentPath)
+      .then(function(baseRepoPath) {
+        return { path: path.resolve(baseRepoPath.trim()) };
+      }).catch(function(e) {
+        if (e.errorCode === 'not-a-repository') {
+          return {};
+        }
+        throw e;
+      }));
   });
 
   app.get(exports.pathPrefix + '/submodules', ensureAuthenticated, ensurePathExists, function(req, res){
--- a/source/git-api.js
+++ b/source/git-api.js
@@ -109,15 +109,8 @@ exports.registerApi = function(env) {
     return promise.then(function(result) {
         res.json(result || {});
       }).catch(function(err) {
+        winston.warn('Responding with ERROR: ', JSON.stringify(err));
         res.status(400).json(err);
-
-        if (err instanceof Error) {
-           console.error('Unhandled error: ' + err.message);
-           console.error(err.stack);
-         } else {
-           console.error('Unhandled error');
-           console.error(err);
-         }
       });
   }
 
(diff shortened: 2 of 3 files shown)


campersau:
> Not showing stack trace to reduce humane alertness

I think this is not true. You are logging the stack trace but since you are using `JSON.stringify` the stack takes up only one line.

But I don't really get why you are doing `JSON.stringify` in the first place. Because you can just pass the `err` to `winston`:
```js
winston.warn('Responding with ERROR: ', err);
```
And if you want to suppress logging the stack trace I would make it configurable just like [these](https://github.com/FredrikNoren/ungit/blob/master/source/config.js#L133-L135).
So if a user encounters an error he can set the config option and replay what he just did and send us the log. What do you think?

---

The other changes are great :+1:

codingtwinky:
I should explained my thought process more clearly.  

For me, seeing stack traces in log like below is a visual key for something **terrible** happened and requires further investigation.

```javascript
> winston.warn(err);
warn:  Error: abc
    at repl:1:10
    at REPLServer.defaultEval (repl.js:252:27)
    at bound (domain.js:287:14)
    at REPLServer.runBound [as eval] (domain.js:300:12)
    at REPLServer.<anonymous> (repl.js:417:12)
    at emitOne (events.js:95:20)
    at REPLServer.emit (events.js:182:7)
    at REPLServer.Interface._onLine (readline.js:211:10)
    at REPLServer.Interface._line (readline.js:550:8)
```

For this case, [rest api error line](https://github.com/fredriknoren/ungit/blob/master/source/git-api.js#L114-L119), doesn't mean something unexpected or terrible happened.  It is often the case that errors are minor or expected and we simply return the error back to client side and let them handle it.  

Now, that doesn't negate the necessity of stack traces for debugging purpose in some cases and we shouldn't ignore the stack trace.  When I would like to preserve non critical stack traces in logs, I would print them in one line as such and thanks to added `Error#toJSON()` it is much easier now.

```javascript
> var err = new Error('abc')
undefined
> JSON.stringify(err)
'{}'
> Object.defineProperty(Error.prototype, 'toJSON', {
...   value: function() {
.....     var alt = {};
.....     Object.getOwnPropertyNames(this).forEach(function(key) {
.......       alt[key] = this[key];
.......     }, this);
.....     return alt;
.....   },
...   configurable: true
... });
[Error]
> JSON.stringify(err)
'{"stack":"Error: abc\\n    at repl:1:11\\n    at REPLServer.defaultEval (repl.js:252:27)\\n    at bound (domain.js:287:14)\\n    at REPLServer.runBound [as eval] (domain.js:300:12)\\n    at REPLServer.<anonymous> (repl.js:417:12)\\n    at emitOne (events.js:95:20)\\n    at REPLServer.emit (events.js:182:7)\\n    at REPLServer.Interface._onLine (readline.js:211:10)\\n    at REPLServer.Interface._line (readline.js:550:8)\\n    at REPLServer.Interface._ttyWrite (readline.js:827:14)","message":"abc"}'
```

Basically idea is this: Logs are for humans to read, and seeing a formatted stack trace in logs should be a visual cue for us to start investigate.  But if one would like to preserve none critical stack trace in log, do it in subtle style such as writing them in one line.

This kind of practice maybe over doing things in simpler application as ungit but I thought it may apply here.

campersau:
I thought that at this point always something **terrible** has happend :smirk: 

I know there are [some errors](https://github.com/FredrikNoren/ungit/blob/master/public/source/server.js#L173-L182) which are ignored at the client side but other than that?

Also I think that people will only really look at the console / log if something bad has happend on the client side and then they will immediately see what the source of the problem was.
So that is my point of view :smile:

codingtwinky:
For example, when merge or stash op with conflicts will be at that line.  Which is not an error that is handled by displaying resolve conflict screen.  Which is not a terrible error.  

And my thinking is that when people do look at the log, it should obvious between minor error vs. major error so it would be easier for people to figure out what went wrong fast.

codingtwinky:
Merging, pleas open new issues or pr if there are any questions.

PR #707, opened 2016-02-27
Fix maxConcurrentGitOperations

bakkerthehacker:
Somewhere in the move to a promise based git system, ungit stopped using this config option.  

This changes adds the `maxConcurrentGitOperations` option back.  This change is quite important to me as several repositories only work well with `maxConcurrentGitOperations = 1`.  Without it, ungit complains about `index.lock` files


codingtwinky:
I like the thought but why are we bringing in async lib?  There is no need for this.

codingtwinky:
Merging as there is no clear and easy alternate way to do this in bluebird.  Although there are some other bluebird plugins we could use it hardly seems worth it.

Issue #177, opened 2013-08-31
Support bare repositories


FredrikNoren:
Right now ungit crashes (#168) if you try to run it in a bare repository.

mesosteros:
+1 for this. This feature would allow us to effectively have a sort of personal github/gitlab. It would be really useful for small teams. We'll be using normal repos for now but would appreciate this.

michaelachrisco:
I am having the same issue with remote repos that may or may not be set up. Some of the errors are resolved if you run `ungit --no-autoFetch`

codingtwinky:
Quite frankly, i thought this was fixed.  When i do `mkdir test; cd test; git init` and than going into test director didn't cause any errors for me.  What error are you seeing specifically?

campersau:
@codingtwinky try `git init --bare`. For  local bare repositories ungits `quickstatus` just returns `"uninited"` (no errors).

Ajedi32:
Yeah, "bare" is not the same thing as "empty". A bare repository is basically just the `.git` folder, with no working directory. It's often used on servers where a working directory isn't needed. (So you can push and pull from the repository, but the server doesn't need to a copy of the source at any given revision.)

codingtwinky:
Sorry for my own incompetence that took forever to fix such easy fix.

It is should now support bare repo without any error messages but with disabled staging and stash as it is not applicable to bare repo.

michaelachrisco:
Thanks @codingtwinky!

PR #720, opened 2016-03-19
Fix barerepo

codingtwinky:
fix: #177 

screen shot 2016-03-18 at 6 08 14 pm

### DIFF
--- a/clicktests/test.all.js
+++ b/clicktests/test.all.js
@@ -8,3 +8,4 @@ require('./test.generic');
 require('./test.noheader');
 require('./test.discard');
 require('./test.branches');
+require('./test.bare')
--- a/clicktests/test.all.js
+++ b/clicktests/test.all.js
@@ -8,4 +8,4 @@ require('./test.generic');
 require('./test.noheader');
 require('./test.discard');
 require('./test.branches');
-require('./test.bare')
+require('./test.bare');
--- a/clicktests/test.bare.js
+++ b/clicktests/test.bare.js
@@ -1,0 +1,46 @@
+
+var helpers = require('./helpers');
+var testsuite = require('./testsuite');
+var Environment = require('./environment');
+var webpage = require('webpage');
+
+var page = webpage.create();
+var suite = testsuite.newSuite('authentication', page);
+
+var environment;
+var testRepoPath;
+
+suite.test('Init', function(done) {
+  environment = new Environment(page);
+  environment.init(function(err) {
+    if (err) return done(err);
+    // testRepoPath = environment.path + '/testrepo';
+    testRepoPath = '/tmp/testrepo'
+    environment.createRepos([
+      { bare: true, path: testRepoPath }
+      ], done);
+  });
+});
+
+suite.test('Open path screen', function(done) {
+  page.open(environment.url + '/#/repository?path=' + encodeURIComponent(testRepoPath), function () {
+    helpers.waitForElementVisible(page, '.graph', function() {
+      done();
+    });
+  });
+});
+
+suite.test('updateBranches button without branches', function(done) {
+  helpers.waitForElementVisible(page, '[data-ta-clickable="branch"]', function(err) {
+    helpers.click(page, '[data-ta-clickable="branch"]');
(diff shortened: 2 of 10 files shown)


clicktests/test.all.js
```diff
@@ -8,3 +8,4 @@ require('./test.generic');
 require('./test.noheader');
 require('./test.discard');
 require('./test.branches');
+require('./test.bare')
```
campersau: Nit: missing semicolon at the end.

clicktests/test.bare.js
```diff
@@ -0,0 +1,46 @@
...
+
+var environment;
+var testRepoPath;
+
+suite.test('Init', function(done) {
+  environment = new Environment(page);
+  environment.init(function(err) {
+    if (err) return done(err);
+    // testRepoPath = environment.path + '/testrepo';
+    testRepoPath = '/tmp/testrepo'
```
campersau: Nit: missing semicolon at the end.

source/git-api.js
```diff
@@ -459,12 +459,14 @@ exports.registerApi = function(env) {
...
-        return gitPromise(['rev-parse', '--is-inside-work-tree'], req.query.path)
-          .catch(function(err) {
-            return 'uninited';
-          }).then(function(result) {
-            if (result.toString().indexOf('true') == -1) return 'uninited';
-            else return 'inited';
+        return gitPromise.revParse(req.query.path, '--is-inside-work-tree')
+          .then(function(isWorkingDir) {
+            if (isWorkingDir) {
+              return 'inited'
```
campersau: Nit: missing semicolon at the end.

source/git-api.js
```diff
@@ -459,12 +459,14 @@ exports.registerApi = function(env) {
...
+            } else {
+              return gitPromise.revParse(req.query.path, '--is-bare-repository')
+                .then(function(isBareDir) { return isBareDir ? 'bare' : 'uninited'; })
```
campersau: Nit: missing semicolon at the end.

source/git-promise.js
```diff
@@ -378,4 +378,13 @@ git.commit = function(repoPath, amend, message, files) {
   });
 }
 
+git.revParse = function(repoPath, type) {
+  return git(['rev-parse', type], repoPath)
+    .catch(function(err) {
+      return false;
+    }).then(function(result) {
+      return result.toString().indexOf('true') > -1
```
campersau: Nit: missing semicolon at the end.

campersau:
If the stash is not expanded and you change to a bare repository then the stash button is visible:
image

If you then click on the stash button to expand it, it disappears.

codingtwinky:
I wonder if you can tell that I'm work on semicolon less languages...  

Thanks @campersau and will be fixing that soon.

codingtwinky:
FYI, git diff in bare repo is not working.  does anybody know a way around this?  maybe @Ajedi32?

simple `git diff` doesn't work at all.

screen shot 2016-03-22 at 5 45 41 pm

Ajedi32:
When you run `git diff` with no arguments it's comparing against the working directory. (It's checking for unstaged, uncommitted changes in the working directory.) So it doesn't really make sense to use that command in a repo that doesn't _have_ a working directory.

You'll have to be more explicit about what revisions you want it to compare.

campersau:
@codingtwinky the problem isn't the actual `git diff` command created by `diffFile` but the `git.status` call in [`diffFile`](https://github.com/FredrikNoren/ungit/blob/master/source/git-promise.js#L252). Which is only used to determine if the file is new.
`git.status` does a lot of things. E.g. also calling `git diff --numstat` for receiving additions / deletions which will fail in bare repositories for the reasons described by @Ajedi32 

The actual diff command used by `diffFile` is already correct and will look like this:
`git diff f4f439e2^ f4f439e2 -- "components/repository/repository.js"`

So there are two solutions which come to my mind:
1. Handle failures of the `git.status` call and proceed with the actual diff logic.
2. Add a parameter to `git.diffFile` and do not call `git.status` at all if the parameter is set.

codingtwinky:
Merging this for now.  created #728 and will be fixing it later

PR #724, opened 2016-03-29
Dependency bumps

codingtwinky:
* yarg@4.3.2 handles "version()" slightly different
* diff2html@2.0.0-beta9 is breaking: https://github.com/rtfpessoa/diff2html/issues/65
* phantomjs@2.1.3 is still super slow and result in timing out


FredrikNoren:
Wow the builds are just horrible. We really need to fix them (I wish I didn't have a million other things to do now :/ )

codingtwinky:
I think it might have something to do grunt plugins is yet to be able to handle grunt@1.0.0 for older nodes or npm.  What my latest change works in node 4.1 so hopefully this will fixes other versions as well.

package.json
```diff
@@ -17,10 +17,10 @@
     "0ungit-credentials-helper": "./bin/credentials-helper"
   },
   "dependencies": {
-    "async": "~1.5.2",
-    "bluebird": "~3.2.2",
-    "blueimp-md5": "~2.1.0",
-    "body-parser": "~1.14.2",
+    "async": "~2.0.0-rc.2",
```
campersau: This bump breaks the directory autocomplete since [`async.filter` now wants an error object as first argument](https://github.com/caolan/async/blob/master/CHANGELOG.md#breaking-changes).

Will submit a PR shortly.

Issue #444, opened 2014-09-12
UI Improvements


CezaryDanielNowak:
Hi,
- main page: There should be a list of recent git repos (localStorage?)
- modified files view/commited files view: File paths should be selectable (`a` instead of `button`s)
- "Amend last commit" still have old "✓" sign.
- there should be "branch" dropdown for easy branch change
- there should be "undo" button for lots of actions (remove file)
- there should be option "discard change in this line" (eg. if you want to remove console.log)
- stash: There should be a option for preview stash content.
- file preview in commited files log is not scrolling horizontally
- ungit version in top area should be linked to `https://github.com/FredrikNoren/ungit`

PR #725, opened 2016-04-01
Wordwrap fix: #721

codingtwinky:
screen shot 2016-03-31 at 7 19 03 pm

PR #726, opened 2016-04-02
show diffs for stash items

campersau:
reference #444

Clicking on the stash title will show the changed files.

image

image

### DIFF
--- a/components/staging/staging.html
+++ b/components/staging/staging.html
@@ -49,19 +49,19 @@
             <!-- ko component: stashProgressBar --><!-- /ko -->
           </button>
           <div class="btn-group pull-right btn-group-sm">
-            <button class="btn btn-default bootstrap-tooltip" data-ta-clickable="stage-sideBySideDiff" data-bind="click: textDiffTypeChange.bind($data, 'sidebysidediff'), css: {active: textDiffType() === 'sidebysidediff'}" data-toggle="tooltip" data-placement="bottom" data-original-title="Show side by side diff view" data-delay='{"show":"2000", "hide":"0"}'>
-              <span data-bind="text: 'Side by Side'"></span>
-            </button>
             <button class="btn btn-default bootstrap-tooltip" data-ta-clickable="stage-defaultDiff" data-bind="click: textDiffTypeChange.bind($data, 'textdiff'), css: {active: textDiffType() === 'textdiff'}" data-toggle="tooltip" data-placement="bottom" data-original-title="Show inline diff view" data-delay='{"show":"2000", "hide":"0"}'>
               <span data-bind="text: 'Default'"></span>
             </button>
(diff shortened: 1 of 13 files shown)


campersau:
@codingtwinky wouldn't it be better to move the `Default` and `Side by Side` buttons and the `textDiffType` logic to the `commitdiff` component?

Because these things are now duplicated in the `stash` and `commit` component.

FredrikNoren:
Really nice! Only two comments from me are the usual suspects: add to CHANGELOG and, if we can, a basic click-test so it doesn't break in the future

codingtwinky:
@campersau I do like the thought and if we do that we can reduce duplicate logic.  we can also further remove same duplicate logic via recycling comittdiff component and use it in staging, right now there are many duplicate logics between diff for staging and diff for commits as code base is different. 

If you would like to do that with in this PR please let me know so I can pause #725 as i'd imagine there will be many conflicts.

campersau:
@codingtwinky okay lets do this after this PR and #725 are merged.

codingtwinky:
@campersau #725 is merged, have you had the chance to work on this yet?  Is there anything I can do to help?  I think this would be nice...

campersau:
@codingtwinky I will try to work on this this week.

components/staging/staging.html
```diff
@@ -49,20 +49,20 @@
...
-            <button class="btn btn-default bootstrap-tooltip" data-ta-clickable="stage-wordwrap" data-bind="click: toggleWordWrap.bind($data, true), css: {active: wordWrap}" data-toggle="tooltip" data-placement="bottom" data-original-title="Wrap words per line" data-delay='{"show":"2000", "hide":"0"}'>
-              <span data-bind="text: 'Word Wrap'"></span>
-            </button>
             <button class="btn btn-default bootstrap-tooltip" data-ta-clickable="stage-nowrap" data-bind="click: toggleWordWrap.bind($data, false), css: {active: !wordWrap()}" data-toggle="tooltip" data-placement="bottom" data-original-title="Not wrapping words per line" data-delay='{"show":"2000", "hide":"0"}'>
               <span data-bind="text: 'Default'"></span>
             </button>
+            <button class="btn btn-default bootstrap-tooltip" data-ta-clickable="stage-wordwrap" data-bind="click: toggleWordWrap.bind($data, true), css: {active: wordWrap}" data-toggle="tooltip" data-placement="bottom" data-original-title="Wrap words per line" data-delay='{"show":"2000", "hide":"0"}'>
+              <span data-bind="text: 'Word Wrap'"></span>
+            </button>
           </div>
```
campersau: I just moved these things around to be consistent with the other buttons. (Show `Default` as first option.)

campersau:
I think this is good to go now:
- moved the `Default` and `Side by Side` buttons and the `textDiffType` logic to the `commitdiff` component
- added clicktests for the `stash` component. stash -> open diff -> pop stash

codingtwinky:
:+1:

Thanks for moving the toggle button logic to commit diff!

PR #727, opened 2016-04-02
fix callback arguments for async.filter so directory autocomplete wor…

campersau:
…ks again

this was a breaking change in the async library 2.0
https://github.com/caolan/async/blob/master/CHANGELOG.md#breaking-changes

see comment: https://github.com/FredrikNoren/ungit/pull/724#discussion_r58298088


codingtwinky:
Nice catch!

PR #729, opened 2016-04-04
fix diffs for bare repositories

### LINKED ISSUE #728: bare repo diff not working
image

See: #720 fore more details

campersau:
by checking if the repository is bare and if so not call `git.status`

closes #728


codingtwinky:
to paraphrase @FredrikNoren: can we please update the change log and add more tests?

:D

campersau:
Added test and update changelog.

codingtwinky:
Thanks!

Issue #515, opened 2015-03-10
error when I checkout a remote branch


hsoc:
In one particular repo I get the following error. I happens all the time. The lockfile (or any other git process) doesn't exist. (Also, it happens even after re-cloning)

    Unhandled git error!
    Ungit tried to run a git command that resulted in an unhandled error. An automatic bug report was sent.
    Command

    git -c color.ui=false -c core.quotepath=false -c core.pager=cat reset --hard "refs/remotes/origin/master"

    Error: Command failed: fatal: Unable to create 'HIDDEN/.git/index.lock': File exists.

    If no other git process is currently running, this probably means a
    git process crashed in this repository earlier. Make sure no other git
    process is running and remove the file manually to continue.

codingtwinky:
Hi, sorry for taking so long.

[git-api.js#L182](https://github.com/FredrikNoren/ungit/blob/master/source/git-api.js#L182) is where doing this operation and as you can see this is a simple git operation.  I don't believe there are any other git operations that would create `index.lock` by ungit while this operation is running.  So I don't think it's ungit who is creating `index.lock`.  

Are you sure you don't have any other user or any other git operation was running out side of ungit at this time?

hsoc:
Hi, as far as I see there is no actual lock file in the filesystem so I can't remove it. This happens when I am checking out a remote repo. I'll gladly provide you with more feedback.

codingtwinky:
![screen shot 2016-04-04 at 5 16 43 pm](https://cloud.githubusercontent.com/assets/5281068/14267849/05e80a4c-fa89-11e5-8f27-52d41e6b223d.png)


### Cause: 

Hmm this issue is more in depth than I thought.  As you may know ungit parallelize many git tasks and for example, "commit" operation is a sequence of many different git commands.  when commit is big or repo is complicated this obviously increases the latency on each git commands.  And some git commands cannot be ran parallel and causes locking issue as this.  And I bet similar thing is happening in this issue where remote branch's git operations are stepping on each other.  But by the time one goes to the file dir to check for `.lock` file git command is done and disappears.  therefor, one would never see the lock file.  I really should have paid more attention to @bakkerthehacker 's #707 as he clearly illustrates this but one reason or another I have over looked it at the time.

### workaround: 

As @bakkerthehacker pointed out, this can be fixed via doing `maxConcurrentGitOperations = 1`.  If many multiple users are sharing same ungit instances, this can potentially slow down the server severely.  

### proposed solution: 

[...]

PR #731, opened 2016-04-13
Fix ".lock" file conflicts in parallelized git operations

codingtwinky:
fix: #515 
related: #707 

Currently there are race condition in several git operations we do.  I fixed one obvious bug in /commit but for others it was more complicated due to work space change triggers "stash and pop" operations and etc.  So I just ended up added retry with random backoff time like any other parallelized systems does... 

Although this bug will be very unlikely to be appear again I have kept the fix from #707 as it is good to limit number of git operations.

### DIFF
--- a/source/git-promise.js
+++ b/source/git-promise.js
@@ -352,8 +352,6 @@ git.commit = function(repoPath, amend, message, files) {
   }).then(function(status) {
     var toAdd = [];
     var toRemove = [];
-    var addPromise;     // a promise that add all files in toAdd
-    var removePromise;  // a proimse that removes all files in toRemove
     var diffPatchPromises = []; // promiese that patches each files individually
 
     for (var v in files) {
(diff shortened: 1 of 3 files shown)


source/git-promise.js
```diff
@@ -374,14 +385,14 @@ git.commit = function(repoPath, amend, message, files) {
       }
     }
 
-    if (toAdd.length > 0) {
-      addPromise = git(['update-index', '--add', '--stdin'], repoPath, null, null, toAdd.join('\n'));
-    }
-    if (toRemove.length > 0) {
-      removePromise = git(['update-index', '--remove', '--stdin'], repoPath, null, null, toRemove.join('\n'));
-    }
+    var commitPromiseChain = new Promise(function(resolve) { resolve() }) // start with dummy
```
campersau: You can use `Promise.resolve()` instead of `new Promise(function(resolve) { resolve() })`.

FredrikNoren:
I've noticed that the lock file problem has gotten worse the past few months too. Really great seeing this solved. It's a little unfortunate we have to rely on retries, but at the same time I don't really have a better suggestion, so 👍  from me.

Looks like github for windows also ran into this problem: http://githubengineering.com/git-concurrency-in-github-desktop/

source/git-promise.js
```diff
@@ -374,14 +385,14 @@ git.commit = function(repoPath, amend, message, files) {
...
 
-    if (toAdd.length > 0) {
-      addPromise = git(['update-index', '--add', '--stdin'], repoPath, null, null, toAdd.join('\n'));
-    }
-    if (toRemove.length > 0) {
-      removePromise = git(['update-index', '--remove', '--stdin'], repoPath, null, null, toRemove.join('\n'));
-    }
+    var commitPromiseChain = Promise.resolve()
+      .then(function() {
+        if (toRemove.length > 0) return git(['update-index', '--remove', '--stdin'], repoPath, null, null, toRemove.join('\n'))
```
campersau: Missing semicolon at the end.

PR #739, opened 2016-04-28
Allow Ungit to functional under sub dir of a git dir

### LINKED ISSUE #734: viewing a commit's file changes is broken
When I click on a previous commit bubble, I am able to see a list of files that were changed as well as the number of lines changed in each file. However, when I click on a file, I am unable to see the actual changes made to that file, although I can see the UI twitch on click and `Event: Object {event: "graph-render"}` appears in the JS console.

ungit@0.10.1

codingtwinky:
Allows operation at a sub dir of git dir.

i.e. if there is a git directory `/tmp/test`, navigating to any sub dir under `/tmp/test` in ungit should be functional as if one is at `/tmp/test`.

I think this will fix #734 but I'm too lazy to test it... :P maybe I will do it over the weekends... :D

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](http://semver.org/) and
 ### Fixed
 - Fix maxConcurrentGitOperations not limiting git processes [#707](https://github.com/FredrikNoren/ungit/issues/707)
 - Fix ".lock" file conflicts in parallelized git operations [#515](https://github.com/FredrikNoren/ungit/issues/515)
+- Allow Ungit to function under sub dir of a git dir [#734](https://github.com/FredrikNoren/ungit/issues/734)
 
 ### Changed
 - Cleaner rebase conflict message display [#708](https://github.com/FredrikNoren/ungit/pull/708)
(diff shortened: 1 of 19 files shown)


codingtwinky:
In my testing this fixes #734.

Merging.

campersau:
This breaks a test in windows. You can see it here: https://ci.appveyor.com/project/FredrikNoren/ungit/build/1.0.224#L1821 (line 1821)

It also automatically changes `\` to `/` in the path input. Which is bad on windows because we don't get autocompletion. Also I am now unable to navigate to a git subdirectory. For example it automatically redirects from `P:\ungit\components` to `P:/ungit` or was this intended?
image

codingtwinky:
is navigation to sub directory of a git directory a valid use case? I'm not sure if it is.    

Right I see that windows test are unhapy and it will need more thought.

campersau:
I am also not sure. Navigating to submodules works. Just wanted to know if this was intended :smiley:

codingtwinky:
Hooray! unless objections arises, I don't think navigation to sub dir of a git dir is a valid use case thus such action will result in redirection to git root dir.

If I can't come up with a valid fix for windows I will back out this merge before tonight

PR #744, opened 2016-05-01
return normalized path for windows

codingtwinky:
To fix broken windows path problem created by #739


campersau:
Yes, this fixes it. Clicktests passed on my machine.

PR #745, opened 2016-05-06
Node6 test fix

codingtwinky:
Ungit is functional with node6 but test... is a different story.

For whatever the reason, pahntom's [onLoadFinished](http://phantomjs.org/api/webpage/handler/on-load-finished.html) didn't work properly for us, it never fires for some of the click tests and thus ungit exiting due to inactivity timeout. I have looked into it but I don't have any solid clues as all api calls returned, except for gravatar, which we ignore, and socketIO, which is a persistent connection.  And test still failed when I took out socketIO so I'm very confused to why this is happening.  

For now, I have removed dependency on onLoadFinished.


source/git-promise.js
```diff
@@ -179,6 +179,10 @@ git.status = function(repoPath, file) {
               .then(function(commitMessage) {
                 status.commitMessage = commitMessage;
                 return status;
+              }).catch(err => {
+                // 'MERGE_MSG' file is gone away, which means we are no longer in merge state
+                // and state changed while this call is being made.
+                status.inMerge = status.inMerge = false;
```
campersau: Why is `status.inMerge` duplicated?

codingtwinky: I was just checking to see if anybody was paying a attention....

codingtwinky:
TLDR:  Because we were actually returning html dom rather than simple object, our page.evaluate was very expansive and causing timeouts for our click tests.  It is fixed and tests should be more stable and fast.

-----


```
[ui] : Event: [object Object]
2016-05-08T23:45:44.745Z Trying to click [data-ta-clickable="submodules-menu"]
2016-05-08T23:45:44.751Z Waiting for element visible: [data-ta-container="submodules"] [data-ta-clickable="subrepo"]
2016-05-08T23:45:44.751Z Querying element visible: [data-ta-container="submodules"] [data-ta-clickable="subrepo"]
2016-05-08T23:45:49.911Z Element visible: [data-ta-container="submodules"] [data-ta-clickable="subrepo"]
2016-05-08T23:45:50.293Z ## Test ok: submodules - 02 Submodule add
2016-05-08T23:45:50.293Z ## Running test : submodules - 03 Submodule update
2016-05-08T23:45:50.636Z Trying to click [data-ta-clickable="update-submodule"]
```

One of the main reason why our click tests were unstable is due to the fact that Ungit test server is under utilized used during test run and script automatically terminates thinking that it has failed or stuck.  As you can see above, it took ~5 seconds to execute "Querying element visible" to "Element visible" and sometimes this takes even longer and causes such instability in click tests.

Reason why [`helpers.elementVisible()`](https://github.com/FredrikNoren/ungit/pull/745/commits/ecc2dc3de544e59d45e1db1a365a6e2f0c09d82e#diff-39385e855fcad5c7d314c1088ad22c01L27) was so slow was because phantom query was returning actual dom object rather than a simple object.  Later versions of phantomjs returns an complex html dom element that is tied to the browser dom object when evalute function returns html dom and it took forever for phantomjs to build such complex object.

FYI, this was behavior was able to be exaggerated via parallelized click test, `grunt clickParallel`, and I think this will help us with faster churn of click test and increasing general stability of it.

codingtwinky:
Seems like we have first legitimate full build pass including windows build.

However, I still see instabilities in [cherry pick test](https://ci.appveyor.com/project/FredrikNoren/ungit/build/1.0.245) and [rebase test](https://ci.appveyor.com/project/FredrikNoren/ungit/build/1.0.244) as I don't see clear fixes yet and it seems like intermittent failures.

Issue #753, opened 2016-06-01
Revert a merge commit


sedge:
This operation fails because `ungit` doesn't detect that the commit selected is a merge commit, which requires an extra option:
![screen shot 2016-06-01 at 2 15 31 pm](https://cloud.githubusercontent.com/assets/1616860/15720689/6dbd3f46-2803-11e6-8f4d-50b239f842eb.png)

sedge:
🎉

codingtwinky:
Just give a heads up.  Ungit will pick 1st parent blindly on revert merge.

PR #757, opened 2016-06-14
Handle revert merge

codingtwinky:
fixes: #753 

To be honest, I really don't like this one.  one should be able to choose either 1st parent or 2nd parent but this one is hardcoded to parent1.  which is not ideal.  

I almost want to recommend users to moving the ref point instead of reverting.  How do you guys feel?  

(Upon confirmation I will add tests, and change log edit)

cc: @FredrikNoren @Ajedi32 @campersau

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](http://semver.org/) and
 - Removed deprecated npmconf package [#746](https://github.com/FredrikNoren/ungit/issues/746)
 - More helpful warning messages [#749](https://github.com/FredrikNoren/ungit/pull/749/files)
 - Deleting already deleted remote tag [#748](https://github.com/FredrikNoren/ungit/pull/748)
+- Fix to handle revert merge commit [#757](https://github.com/FredrikNoren/ungit/pull/757)
 
 ### Changed
 - Cleaner rebase conflict message display [#708](https://github.com/FredrikNoren/ungit/pull/708)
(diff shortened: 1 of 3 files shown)


Ajedi32:
`git revert` isn't the same thing as moving the ref point. In particular, there's a whole bunch of situations where the later isn't a good idea, like when others have work based off of the branch you're reverting. I think it makes sense to support both workflows.

Yeah, ideally there should be some way to select the mainline parent for the revert. The first parent _is_ probably going to be the correct choice ~75% of the time though (my totally unsubstantiated guess).

You could maybe improve that to like ~90% by guessing the parent based on the commit message used for the merge commit. E.g. Use the second parent instead of the first if the reverted commit's message says `Merge branch 'master' into feature` instead of the other way around and you're currently on the master branch.

codingtwinky:
Right, I understand the difference between git revert and the moving the ref point.  But what I'm trying to say is that end result will be same and it's operation is much easier in ungit UI.  (Moving ref backward is a destructive op so perhaps it isn't "nice" either)

I may think on this more but I probably implement it as it is and maybe improve it later...

Ajedi32:
The end result isn't the same. Especially if you have extra commits after the one you're trying to revert.

If you have:

```
a - b - c - d
```

And run `git revert c`, the end result will look like this:

```
a - b - c - d - e
```

And the working directory will still have the changes from d.

Whereas if you run `git reset --hard b`, the end result will look like this:

```
a - b
```

And the working directory will (obviously) not contain the changes from `c` or `d`.

> I may think on this more but I probably implement it as it is and maybe improve it later...

Yeah, being able to revert merge commits, even if only in a rather limited way, is a definite improvement over not being able to revert them at all.

codingtwinky:
Git [2.9.0](https://github.com/blog/2188-git-2-9-has-been-released) is released and it seems that it is breaking the "git edge" build.  

This is supper annoying...

PR #764, opened 2016-06-26
sysinfo.js refactor

codingtwinky:
fully promisifying sysinfo.js 

~~[work still under progress...]~~

ready for review

### DIFF
--- a/source/config.js
+++ b/source/config.js
@@ -5,6 +5,8 @@ const path = require('path');
 const fs = require('fs');
 const yargs = require('yargs');
 const homedir = require('os-homedir')();
+const winston = require('winston');
+const child_process = require('child_process');
 
 const defaultConfig = {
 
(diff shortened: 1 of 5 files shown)


source/config.js
```diff
@@ -212,3 +214,22 @@ Object.defineProperty(Error.prototype, 'toJSON', {
   },
   configurable: true
 });
+
+try {
+  module.exports.gitVersion = /.*?(\d+[.]\d+[.]\d+).*/.exec(child_process.execSync('git --version').toString())[1];
```
FredrikNoren: I think the reason these were async before was to shorten the startup time. Not sure if it actually is a big difference though, but might be good to look out for (i.e. if it's significantly slower maybe we should think about keeping the async version)

FredrikNoren:
other than that one comment it looks good to me

FredrikNoren:
Oh lol, I'm tired. Already merged :)

source/config.js
codingtwinky: I do favor async calls above sync calls generally but for this case it was much easier to do it this way and seems like unnecessary complication for little benefits.  I don't think start up time was hindered very much.

codingtwinky:
No please do code review even when it is merged for any and all corrections necessary.  My codes can be very sloppy sometimes... 😢

PR #765, opened 2016-06-28
lockPath feature

exsilium:
Firstly, ungit is awesome! Thank you very much for this and I hope you keep developing it further! :sunglasses: 

Based on the contribution doc, this is an early pull request for a concept work to start a discussion around a feature request with the upstream, and if seen feasible, also to focus more on the implementation side.

As of now, ungit already has an option for `forcedLaunchPath` which makes it possible to bind to a specific repository on launch. It was also a delight to find out that there's a (hidden? :flashlight: ) feature which allows to disable the header repository browser using `noheader=true` request.

I'm proposing to take this one step further with also allowing ungit to be restricted to a single repository during startup. As such, a concept of `lockPath: true/false` is proposed which would bind ungit to the repository specified in the config or startup parameters making it impossible (well, at least not easy :wink: ) for users to browse to another repo.

The current concept is just to show what I had in mind and it certainly has a number of deficiencies.. some that immediately comes to mind:
- does not currently respect `process.cwd()` but relies on `forcedLaunchPath` to specify the repo;
- header can be made visible again and autocomplete of directories remains functional, altough switching to another space is restricted. The `lockPath` should at least disable the autocomplete function in a repo browser;

I guess it makes sense also to touch on the topic, why such a feature is requested? I would like to make it possible to embed ungit to a web IDE and restrict it to the same workspace that the IDE is bound to. A screenshot for viewing pleasure:

cloud9-ungit

If this kind of use (or abuse?) of ungit is discouraged, please feel free to close the PR.

Cheers! :beers:

### DIFF
--- a/source/config.js
+++ b/source/config.js
@@ -55,6 +55,9 @@ const defaultConfig = {
   // Instead of launching ungit with the current folder force a different path to be used. Can be set to null to force the home screen.
   forcedLaunchPath: undefined,
 
+  // Lock the path and consequently turn off the repository browser.
+  lockPath: false,
+
   // Closes the server after x ms of inactivity. Mainly used by the clicktesting.
   autoShutdownTimeout: undefined,
 
@@ -156,6 +159,7 @@ let argv = yargs
 .describe('showRebaseAndMergeOnlyOnRefs', 'Set to false to show rebase and merge on drag and drop on all nodes')
 .describe('maxConcurrentGitOperations', 'Maximum number of concurrent git operations')
 .describe('forcedLaunchPath', 'Define path to be used on open. Can be set to null to force the home screen')
+.describe('lockPath', 'Lock the path and consequently turn off the repository browser')
 .describe('autoShutdownTimeout', 'Closes the server after x ms of inactivity. Mainly used by the clicktesting')
 .describe('maxNAutoRestartOnCrash', 'Maximum number of automatic restarts after a crash. Undefined == unlimited')
 .describe('noFFMerge', 'Don\'t fast forward git mergers. See git merge --no-ff documentation')
--- a/source/config.js
+++ b/source/config.js
@@ -55,7 +55,7 @@ const defaultConfig = {
   // Instead of launching ungit with the current folder force a different path to be used. Can be set to null to force the home screen.
   forcedLaunchPath: undefined,
 
-  // Lock the path and consequently turn off the repository browser.
+  // Lock the path to forcedLaunchPath or to the directory from where ungit is launched. Consequently opens the new browser session without header.
   lockPath: false,
 
   // Closes the server after x ms of inactivity. Mainly used by the clicktesting.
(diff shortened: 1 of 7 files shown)


FredrikNoren:
Looks good to me, two minor notes:

1) Can you make sure `forcedLaunchPath` is in the config.js file too so people can find out about it?
2) Please add a description to CHANGELOG.md of this change too, in the Unreleased section

codingtwinky:
@exsilium Have you had the chance to take a look at @FredrikNoren s comments?  I do think this would be a good addition and thanks for helping!

exsilium:
@codingtwinky , @FredrikNoren - Hi! I did and thanks for the positive feedback! I did plan to work a bit more on the feature but then summer happened 🌞 - I apologise for the slowness on my behalf. I still have the intent of submitting update to this PR :)

exsilium:
Hi! Amendments were made to the lockPath feature. Again, apologies for the lag.. Feel free to try it out and let me know what you think. Namely:
- `process.cwd()` is now respected when the `forcedLaunchPath` setting is unset
- Even if header is enabled during lockPath use, an alternative version is shown that does not reveal the directory structure via autocomplete. Additionally, some of the other elements have been disabled in that view (home, back, addToRepoList).

@FredrikNoren, to address your original comments:

1. The `forcedLaunchPath` is already in there, it's an existing feature ;) Or was your comment hinting that the config entry about it should be enhanced somehow?
2. Clear, added an entry.

Cheers! 🍺 ☀️

codingtwinky:
Ah thanks! I think we would have to resolve merge conflicts, can you possibly rebase or merge with the latest code?

exsilium:
Np, rebased version committed. Cheers! 🍻

codingtwinky:
Shoot, I should have merged this before but now it is conflicted again...  @exsilium can you please forgive me and rebase or merge one more time?  Or do you mind me doing it for you?

exsilium:
@codingtwinky - Np! Rebased version is at your disposal. 👍 🍻

codingtwinky:
Hmm this is not working for me.  I have set `lockPath: true, forcedLaunchPath: "/tmp/test` and I get below picture and it seems that I can navigate around beyond the provided path.

Am I missing something?

screen shot 2016-09-06 at 16 50 50

source/config.js
```diff
@@ -55,6 +55,9 @@ const defaultConfig = {
   // Instead of launching ungit with the current folder force a different path to be used. Can be set to null to force the home screen.
   forcedLaunchPath: undefined,
 
+  // Lock the path and consequently turn off the repository browser.
```
codingtwinky: Can we add a description about how this field is related to the `forcedLaunchPath`?

exsilium:
@codingtwinky - Interesting, did you try to actually move to another path? for example `/tmp`? It should return you back to `/tmp/test` - additionally the autocomplete for paths has been disabled.

But in case the lockPath is set the expected behavior here is that the browser should be launched with the `noheader=true` set. Can you share what's the opening URL you get when launching ungit when `lockPath` has been set?

I can amend the description that the lockPath fixes the path to forcedLaunchPath or if unset to the path that the ungit was launched from.

exsilium:
Additionally, if the shown header is creating confusion in case the `noheader=true` flag is missing from the requests, I could potentially remove it altogether...

jung-kim:
close for inactivity

PR #769, opened 2016-07-11
Loosen npm requirement

avindra:
If you want to depend on npm, at least let us use the one that ships with node if we can.

Currently, this installs a separate copy of npm just for this package using node 6.3.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "lodash": "~4.12.0",
     "mkdirp": "~0.5.1",
     "moment": "~2.13.0",
-    "npm": "~3.9.1",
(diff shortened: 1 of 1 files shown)


lfx:
What is your OS (please be specific)? For example on ubuntu npm do not come with node.

avindra:
SuSE Linux Enterprise Server 11 SP3

Using the statically compiled Linux x64 binaries for 6.3.0:

https://nodejs.org/dist/latest/node-v6.3.0-linux-x64.tar.xz

avindra:
Thanks @FredrikNoren  :confetti_ball:

Issue #420, opened 2014-08-05
When other branches get far ahead of your current one, your current branch is not displayed


hirden:
When you are on a branch which then get passed in commits by another remote, your current "branch track" is not displayed.
![ungit issue](https://cloud.githubusercontent.com/assets/82352/3810770/00bb233e-1c95-11e4-9e22-03e02284e85f.PNG)

daedalus28:
I know this is an old ticket, but I wanted to bring this up as I have a use case where scrolling isn't really viable. In my project I have some branches that are very far behind and when switching to them (using the fancy new branch picker dropdown :smile:), I can't see the current commit to merge in the latest version of the main branch. These branches are few months behind and there are so many commits to scroll though that I really can't find it and have to jump to the command line to make it work.

The commit collapsing idea is cool, but even a "jump to" or just "load n more months of history" command would do it.

PR #776, opened 2016-08-08
Active branch focused git.log() result.

codingtwinky:
fix: #420

~~Currently, what this fix is doing is that if "head" ref is missing from git.log, it will continuously increase git.log result list until it does.  Effectively always loading active git branch on load.
Now, this is not optimal and not my favorite.  I'm planning on writing more optimized solution as it can be bit slow and little barbaric solution as it is...~~

Added two config flags

* numberOfNodesPerLoad: number of nodes to load for each git.log call
* alwaysLoadActiveBranch: Always load with active checkout branch. ()

When `alwaysLoadActiveBranch=true`, terrible name but not sure what I can call it,  ungit will load checked out node on load by skipping previous nodes untill active checked out node is found for `git.log` call.  

I have also added a cheeky little button where you can load skipped nodes.

After more testing we might want to change the default value of `alwaysLoadActiveBranch` to true.

jrbixycto9

### DIFF
--- a/source/git-parser.js
+++ b/source/git-parser.js
@@ -97,7 +97,7 @@ exports.parseGitLog = (data) => {
       const refs = row.substring(refStartIndex + 1, row.length - 1);
       currentCommmit.refs = refs.split(/ -> |, /g);
     }
-    currentCommmit.isHead = !!_.find(currentCommmit.refs, (item) => { return item === 'HEAD' });
+    currentCommmit.isHead = !!_.find(currentCommmit.refs, (item) => { return item.trim() === 'HEAD' });
     commits.isHeadExist = commits.isHeadExist || currentCommmit.isHead;
     commits.push(currentCommmit);
     parser = parseHeaderLine;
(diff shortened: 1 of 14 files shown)


source/git-parser.js
```diff
@@ -96,6 +96,8 @@ exports.parseGitLog = (data) => {
       const refs = row.substring(refStartIndex + 1, row.length - 1);
       currentCommmit.refs = refs.split(/ -> |, /g);
     }
+    currentCommmit.isHead = currentCommmit.refs.includes('HEAD');
+    commits.isHeadExist = commits.isHeadExist || currentCommmit.isHead;
```
codingtwinky: Abusing javascripts aray flexibility!! yay!!!

FredrikNoren:
Hey this is really cool! Really like it.

Two points:
1) The idea with the dotted circle was that it's the representation of the "next commit". It's kind of where the stuff in the staging area goes. How about something like this instead?:
collapse
2) Bug: Seems like creating a new branch isn't working while in the "collapsed" mode

codingtwinky:
1. I may do that as a future improvement later when I stop drinking like a russian pirate 
2. Quick UI test by me had branch operations working even in collapsed mode.  Will add tests later

codingtwinky:
merging for now.  

I really hope people who needed #420 experiment with this new feature and discuss how we can further improve this new feature.

PR #781, opened 2016-08-24
Fix for favorites linking in case of rootPath option is used

exsilium:
As reported on May 11th by @sebastianmay for rootPath feature issue #609 I just recently hit that one as well and I hope this will get pulled to the next version. Thanks and Cheers!

:beers:


codingtwinky:
:+1:

But if you could write a click test for it I will give you a virtual highfive...

exsilium:
No problem, I'll take a look :)

codingtwinky:
Feel free to add tests later!

exsilium:
Thanks, still in progress... I was strayed from the path for a while without realizing that the `~/.ungitrc` interferes with the test suite :feelsgood:

viharm:
This issue also occurs on links to parent repository when viewing submodules.

Issue #783, opened 2016-09-05
External conflict merge tool support.


Lochemage:
It seems the only way to resolve any conflicts you may have is to manually launch your merge tool or use your code editor to manually edit the files.

It would be handy to include a `"Launch Merge Tool"` button that will automatically launch your merge tool of choice (assuming that you have properly configured it with your git).

Lochemage:
A fix for this has been merged in.

PR #784, opened 2016-09-05
External merge tool support

Lochemage:
This is to address issue #783. 

I have added a new configuration option `"mergeTool"`. Its default value `false` means not to use this feature. Setting this to `true` will launch your default merge tool (as configured in your git config). Setting this to a `string value` will launch the merge tool configured under the given name.

Basically under the hood, it runs the [git mergetool command](https://git-scm.com/docs/git-mergetool) and will only work if you have actually configured it properly. I've found this [stackoverflow](http://stackoverflow.com/questions/426026/git-on-windows-how-do-you-set-up-a-mergetool) discussion that explains how to set up your mergetool. Basically you need to edit your `~/.gitconfig` file or use the CLI.

Here are some more details on what I have added/changed:
* Configuration option `"mergeTool"` as I have detailed above.
* New URL POST route `/launchmergetool` added to the API.
* Modified existing `Staging` component:
  * Renamed the `"Resolve Conflicts"` hover text to be `"Mark as Resolved"`. This is to make things more clear when you see it aside the new button I have made (below).
  * Created a second button named `"Launch Merge Tool"` that will appear when hovering over the `"Conflicts"` label and only if you have the `"mergeTool"` config option enabled.

Known Issues:
* Sometimes an external tool may take several seconds before it actually appears in view, during this time there is no 'loading' feedback presented to you on the front end. This may be fixable, but will require some experimentation.
* Since the `ungit` interface can technically be accessible from a remote PC via the web address, if you launch the merge tool, it only launches on the machine that is actually running the server. This seems like an unlikely case, however.
* I did not know how to write unit tests for this feature, especially since the merge tool has to be configured outside of `ungit`.

Please let me know what you think.

-- Jeff

P.S. I found it amusing that I checked in my `ungit` code changes using the modified version of `ungit` that I was working on :+1:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ This project adheres to [Semantic Versioning](http://semver.org/) and
 [Keep a changlelog's changelog standard](http://keepachangelog.com/)
 
 ### Added
-- New configuration option `mergeTool` allows you to assign a custom external merge tool for conflict resolution.
+- New configuration option `mergeTool` allows you to assign a custom external merge tool for conflict resolution [783](https://github.com/FredrikNoren/ungit/issues/783)
 
 ## [Unreleased](https://github.com/FredrikNoren/ungit/compare/v0.10.3...master)
 - Fix for favorites linking in case rootPath is used @sebastianmay [#609](https://github.com/FredrikNoren/ungit/issues/609) and image diffing
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/) and
 [Keep a changlelog's changelog standard](http://keepachangelog.com/)
 
+### Added
+- New configuration option `mergeTool` allows you to assign a custom external merge tool for conflict resolution.
+
 ## [Unreleased](https://github.com/FredrikNoren/ungit/compare/v0.10.3...master)
 - Fix for favorites linking in case rootPath is used @sebastianmay [#609](https://github.com/FredrikNoren/ungit/issues/609) and image diffing
 
--- a/MERGETOOL.md
+++ b/MERGETOOL.md
@@ -1,0 +1,69 @@
+If you have your own merge tool that you would like to use, such as Kaleidoscope or p4merge, you can configure ungit to use it by following these steps:  
+
+
+1. Configuring git
+------------------
+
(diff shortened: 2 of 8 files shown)


Lochemage:
It seems my amended commit is failing the appveyor test. As far as I can tell, it is not related to my change:
```
  1 failing
  1) git-api status should list the renamed file:
     Error: expected 200 "OK", got 400 "Bad Request"
      at Test.assert (C:\projects\ungit\node_modules\supertest\lib\test.js:205:15)
      at Server.assert (C:\projects\ungit\node_modules\supertest\lib\test.js:132:12)
      at emitCloseNT (net.js:1542:8)
      at _combinedTickCallback (internal/process/next_tick.js:71:11)
      at process._tickDomainCallback (internal/process/next_tick.js:122:9)
Warning: Task "mochaTest:src" failed. Use --force to continue.
```

codingtwinky:
Yeah window build is bit dodge and we need to fix it.

Although I would like to have tests for this, it be tricky for this one...  Will try it out later.

Thanks!

Lochemage:
Ah, I've only been using ungit for about a week now, and only on Windows. Are you saying that merge conflict resolution is handled differently on Linux?

codingtwinky:
Oh no, I meant the automated test suite in Windows.  Click test in windows is unstable right now for Ungit. 

And this is mostly due to my inability to get a hands on windows machines to replicate the bug and test... :(  If some body can fix them it would be super nice...  Just saying...  👀

codingtwinky:
FYI, just tried this and this is pretty cool!

Can you please update the doc that explain this feature along with a quick example? it would be super nice.

Lochemage:
Ah yes, when I run the unit tests locally on my machine, I get a lot of errors and they don't seem to be the same ones every time. I can't promise anything but I'll try and take a look.

As far as documentation, I don't mind adding more. What did you have in mind? I've put a small description in the [source/config.js](https://github.com/WebCabin/ungit/blob/external-merge-tool/source/config.js) file explaining the new property. Is there some other documentation where you would want this or possibly more detailed info?

codingtwinky:
If you add little description in README.md with an example with a helpful guide I be happy!

Lochemage:
It turned out to be a little more 'wordy' that I expected, so I have broken it into a `MERGETOOL.md` file and linked it from the README.

Please feel free to reformat it however you want.

codingtwinky:
Lovely, thanks!

PR #785, opened 2016-09-05
Commit message title limiting to 72

exsilium:
A small improvement concerning commit subjects.

The overwhelming consensus of recommended git commit message formatting is the 50/72 rule ([Exhibit A](http://chris.beams.io/posts/git-commit/), [Exhibit B](https://github.com/erlang/otp/wiki/Writing-good-commit-messages)). Git GUI clients are suggesting the same rules but are not enforcing them.

When browsing certain repos I found that ungit UI breaks down after certain length of the subject line (150+ characters). The idea of this commit is to guide the user to use shorter subject messages by cutting the visible commit subject to 72 characters (being less obtrusive than the suggested 50 limit). The rest of the commit subject is shown as a separate body div when inspecting the commit.

Thoughts? 🙏

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/) and
 
 ## [Unreleased](https://github.com/FredrikNoren/ungit/compare/v0.10.3...master)
 - Fix for favorites linking in case rootPath is used @sebastianmay [#609](https://github.com/FredrikNoren/ungit/issues/609) and image diffing
+- Limit commit title to 72 characters, the rest is truncated and shown when inspecting the commit
 
 ## [0.10.3](https://github.com/FredrikNoren/ungit/compare/v0.10.2...v0.10.3)
 
--- a/components/commit/commit.html
+++ b/components/commit/commit.html
@@ -11,7 +11,7 @@
           onerror="this.style.display='none';">
         <div>
           <div>
-            <span class="title" data-bind="text: title"></span>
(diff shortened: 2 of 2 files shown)


codingtwinky:
I like the ideas but I would like to have users be notified if the character limit is reached.

exsilium:
Cool, this commit only addresses the viewer part. Thinking this further, when writing a commit Git Tower utilizes the character counter starting from 50. Ungit currently counts from 0 and up. In my view, the same method could be used here.

1. Have the character count of Title start from 50 and decrease as title length is added.
2. When that counter goes below 0 (-1..-22) change the title input background to yellow.
3. When that counter goes below -22 change the title input background to red.

In my mind the tool does not have to police and set hard limits but rather guide the user to submit a proper commit message.

Any further thoughts on this? :beers:

codingtwinky:
I think colorization is very helpful indication.  We should be careful about silently ignoring user's inputs, which I believe this code is doing but please do correct me if I'm wrong.  

Once proper warning message is implemented I think this would be a good way to help users keep their commit message concise.

exsilium:
@codingtwinky - It doesn't ignore user's input. If people want to submit superlong commit titles, they can do it just fine. This patch modifies just the way the long title gets displayed. If the title is longer than 72 characters long, the rest of the title is shown the same way as the description is shown when clicking on the actual commit. This is done because the way things currently are, ungit UI becomes broken when the commit title is longer than 150 characters. I'll prepare some screenshots for comparison.

codingtwinky:
Ah... I see, I thought `substring` is applied for display only.  Cheers,

PR #788, opened 2016-09-08
Log improvements

Lochemage:
This one is kind of random, but I noticed you were using `winston` and it comes with a few features that I thought we would benefit from:

I've added a configuration option `"logLevel"` that allows you to set the level of logging you want to see in the server output console. It is set to `warn` by default, which is actually less logging that it currently does; I figure the majority of users are not developing for ungit and would only care about warning and error messages. You can always set this to `silly` in your local configuration file to get all the logs back.

I've also set `winston` to output log level text using colors.


codingtwinky:
Admittedly we have been very sporadic with logging standard and hopefully this will be the beginning of doing helpful logging for all of us.

Cheers!

Issue #796, opened 2016-09-28
Appveyor windows test fails with permission error within .git folder


codingtwinky:
this is being tracked at http://help.appveyor.com/discussions/problems/5332-gitindex-permission-issue-on-a-git-repo-that-is-created-within-test

```
    2) status should list the renamed file
GET /status?path=C%3A%5CUsers%5Cappveyor%5CAppData%5CLocal%5CTemp%5C1%5Ctest-temp-dir116825-1644-tq1286.wxjc4sra4i&socketId=ignore
Error: expected 200 "OK", got 400 "Bad Request"
    at Test.assert (C:\projects\ungit\node_modules\supertest\lib\test.js:205:15)
    at Server.assert (C:\projects\ungit\node_modules\supertest\lib\test.js:132:12)
    at emitCloseNT (net.js:1542:8)
    at _combinedTickCallback (internal/process/next_tick.js:71:11)
    at process._tickDomainCallback (internal/process/next_tick.js:122:9)
{ isGitError: true,
  errorCode: 'unknown',
  command: '-c color.ui=false -c core.quotepath=false -c core.pager=cat diff --no-renames --numstat --',
  workingDirectory: 'C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\test-temp-dir116825-1644-tq1286.wxjc4sra4i',
  error: 'fatal: .git/index: index file open failed: Permission denied\n',
  message: 'fatal: .git/index: index file open failed: Permission denied',
  stderr: 'fatal: .git/index: index file open failed: Permission denied\n',
  stdout: '' }
```

PR #797, opened 2016-10-02
Add retry for windows file permission issue exception

codingtwinky:
fix: #796

I guess windows throws "permission denied" when unable attain file lock.  Which is possible in ungit as high degree of parallelization of git operations we do in the background.

Hopefully this will fix windows test failures.


codingtwinky:
merging as windows test passed and hopefully we don't see any more failures out of windows system.

Issue #283, opened 2014-01-09
Update Ungit UI when the repository is modified by command line git or other tools


Ajedi32:
I have many tools I use besides ungit that integrate with git. Currently, if I make changes to the repo with any of these tools, I have to refresh Ungit to see the changes there.

That's not too big of a deal, but it would be nice if Ungit handled this in the background somehow so I don't have to worry about whether what I'm looking at is really the current state of the repository.

brettswift:
This is a couple of years old..but I'd love this as well. 

I teach a .git course at work every once in a while and like http://learngitbranching.js.org/ this would be a great tool but to visualize the repo changing at the front of the room on a big display :)

codingtwinky:
@brettswift #809 is merged I think we should be good as we are entire folder recursively and watching `.git` and `.git/refs` in the cases where recursive watches doesn't work. 

Will be closing it for now but please let me know if there any other operations that may not work.

shoutout to @Ajedi32 for guiding me through this.

PR #809, opened 2016-11-08
Git command line exec to be applied to ungit

codingtwinky:
cc: @Ajedi32 @brettswift

This will fix the git checkout and I think that was the main concern of the issue #283.  Is there any other operations I should consider?

0dzz2stlbo

### DIFF
--- a/components/graph/graph.js
+++ b/components/graph/graph.js
@@ -267,6 +267,7 @@ GraphViewModel.prototype.handleBubbledClick = function(elem, event) {
 
 GraphViewModel.prototype.onProgramEvent = function(event) {
   if (event.event == 'git-directory-changed') {
+    this._clearRefs();
     this.loadNodesFromApiThrottled();
     this.updateBranchesThrottled();
   } else if (event.event == 'request-app-content-refresh') {
@@ -304,4 +305,12 @@ GraphViewModel.prototype.checkHeadMove = function(toNode) {
   if (this.HEAD() === toNode) {
     this.HEADref.node(toNode);
   }
+}
+GraphViewModel.prototype._clearRefs = function(toNode) {
+  this.refs().forEach(function(ref) {
+    ref.node(null);
+  });
+
+  this.refs.removeAll();
+  this.refsByRefName = {};
 }
--- a/components/graph/graph.js
+++ b/components/graph/graph.js
@@ -113,9 +113,18 @@ GraphViewModel.prototype.loadNodesFromApi = function(callback) {
       var nodes = log.nodes ? log.nodes : [];
       self.limit(parseInt(log.limit));
       self.skip(parseInt(log.skip));
+      var updatedRefs = [];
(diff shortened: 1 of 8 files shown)


Ajedi32:
A good test case would be `git branch -f some-existing-branch some_commit` since that only affects a branch pointer and doesn't touch the working directory at all.

brettswift:
will this work on the `git branch` command?  I find one thing that's hard for people to learn about git is that git branch doesn't really create a new  visual branch, just a new pointer..  visually this would be nice to show :)  I'd expect two pointers to the same commit..     

I've used ungit a little bit but not a ton so not sure if this works with a manual refresh either

codingtwinky:
Sorry it took so long... there were some "gotchas" that I had to fix for this one...  Is there more I should test?  

fximzctpdg

Ajedi32:
Hmm. How about `git reset --soft <commit>`? That command changes HEAD without touching the working directory or index, so that might also be a good test case. Maybe also make sure `git tag`, `git tag -d`, `git push`, etc work as expected...

components/graph/graph.js
```diff
@@ -127,10 +146,7 @@ GraphViewModel.prototype.loadNodesFromApi = function(callback) {
 
       self.edges(edges);
       self.nodes(nodes);
-
-      if (nodes.length > 0) {
```
campersau: Why did you remove the `if (nodes.length > 0) {` check? Calling `cy()` on `undefined` will throw.

components/graph/graph.js
```diff
@@ -110,13 +110,32 @@ GraphViewModel.prototype.loadNodesFromApi = function(callback) {
...
     .then(function(log) {
-      var nodes = log.nodes ? log.nodes : [];
+      // set new limit and skip
       self.limit(parseInt(log.limit));
       self.skip(parseInt(log.skip));
-      nodes = self.computeNode(nodes.map(function(logEntry) {
-          return self.getNode(logEntry.sha1, logEntry);
-        }));
-
+      return log.nodes || []
```
campersau: Semicolon.

components/graph/graph.js
codingtwinky: Nodes will always have least length of 0 and it now is wrapped around max

components/graph/graph.js
codingtwinky: done

components/graph/graph.js
campersau: Wrapping with `Math.max` does not prevent it from beeing called.

Try the following in a console:
```js
var nodes = [];
Math.max(nodes[nodes.length - 1].cy(), 0);
```

codingtwinky: Ah, you are right... thanks

codingtwinky:
@Ajedi32 This fix definitely works for the `tag` `reset` and `push` for sure and will be adding that tests later.  

I'm trying to see if I can fix this cherry pick test fail for windows but will see...

PR #821, opened 2016-11-27
fix to many file change events from file system watchers

### LINKED ISSUE #820: Endless status requests since #809 on windows
With the new feature from #809 ungit makes endless status requests on windows even though nothing has changed.
Will investigate further over the weekend.

image

image

/cc @codingtwinky

campersau:
fixes #820 

There might be an other bug:
What if `.gitignore` changes? Should we also update the state on the server?
<sub><strike>(currently changes to to `.gitignore` file are ignored because of the `startsWith` check)</strike> also fixed with this PR</sub>

### DIFF
--- a/source/git-api.js
+++ b/source/git-api.js
@@ -38,9 +38,10 @@ exports.registerApi = (env) => {
           const watcherPath = path.join(data.path, ...subFolders);
           const relativPath = subFolders.length > 0 ? path.join(...subFolders) : '';
 
-          const runOnFileWatchEvent = function(event, filename) {
+          const runOnFileWatchEvent = (event, filename) => {
             const filePath = path.join(relativPath, filename);
             if (isFileWatched(filePath, socket.ignore)) {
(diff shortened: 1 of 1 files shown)


codingtwinky:
@campersau is there anything I can do to help you to get the tests passed in travis?  I think file watcher may not be working as intended in linux env.

campersau:
@codingtwinky Yes, I have currently no idea why it does not work on linux (always getting timeouts). I pushed some tests but linux never worked. Windows works from time to time.

codingtwinky:
It seems that problem was mis configured file watch path. I guess over active file watch trigger existed in unix OSs as well and it helped pass during the unix test.  Once you fixed it, tests start failing as it was previously relied on incorrectly triggered events.  

Now the path is corrected we should be good I think.  Let me know how you think @campersau

campersau:
@codingtwinky wow thank you! Yes that seems to be the solution 👍

PR #828, opened 2016-12-11
Fix some weak-points in the clicktests

FredrikNoren:


codingtwinky:
I have tried similar things too, but didn't help....  ㅠ.ㅜ

FredrikNoren:
I really wish we could get to a state where we can 100% trust the clicktests (or any tests that would serve the same purpose), but I don't have any good ideas how to do it :/ I'm also not sure _why_ they are failing so often

codingtwinky:
These kind of race condition exist in [unix test](https://travis-ci.org/FredrikNoren/ungit/jobs/183134182) as well.  For this specific test failure, I think two possibilities are possible: 1. file watch didn't trigger 2. race condition where front end triggered refresh operation didn't properly pickup applied change.

Considering there are other failures, I think 2 is the culprit but this is very hard to prove and test considering it's rarity, especially in unix environment. Assuming this hypothesis is correct I think here are the possible choke points

1. I do lack in google-fu but is it possible that subsequent git command doesn't pick up the changes from previous git commands?  I doubt this is the case but if there is what is the guaranteed await we would have to do?
2. Socket.io failure?  I don't think this is the case any more but it didn't play nice with phantom.js before if I remember correctly
3. Promise operation order misconfiguration? I'm sure there is a place where I have screwed up and promises adoption rate is not at 100%

I have dived into all of above 3 points but I couldn't find anything solid... If anybody have anymore to add please let me know.

PR #830, opened 2016-12-11
Fix left calculation for node diff box

codingtwinky:
screen shot 2016-12-11 at 14 20 39

### DIFF
--- a/components/commit/commit.js
+++ b/components/commit/commit.js
@@ -38,7 +38,7 @@ function CommitViewModel(gitNode) {
   });
 
   this.diffStyle = ko.computed(function() {
-    var marginLeft = Math.min((gitNode.branchOrder() * 70), 450) * -1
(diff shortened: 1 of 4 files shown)


campersau:
I think this was inteded so we don't have to scroll to the right. But yes for wide repos this looked strange...

image

(Also the scrollbar on the bottom is very big.)

components/commit/commit.js
```diff
@@ -37,9 +37,9 @@ function CommitViewModel(args) {
     return self.fileLineDiffs() && self.fileLineDiffs().length > 0;
   });
 
-  this.selectedDiffLeftPosition = ko.observable();
   this.diffStyle = ko.computed(function() {
-    if (self.selected()) return { left: self.selectedDiffLeftPosition() + 'px', width: (window.innerWidth - 220) + 'px' };
+    var marginLeft = Math.min((gitNode.branchOrder() * 70), 450) * -1
```
campersau: Nit: semicolon

PR #833, opened 2016-12-16
Fix initial page load when loaded node does not fits in screen.

### LINKED ISSUE #832: impossible to scroll down on 4k screen
this is how ungit opens up on a right half of a 4k screen:

screenshot-localhost 8448-2016-12-14-11-02-28

no scrollbar visible and even scrolling with a mouse doesn't work.
When I make the window smaller I can start scrolling.

codingtwinky:
fix #832


campersau:
I think you mean #832 (github will close the wrong issue automatically).

PR #848, opened 2017-01-08
[#847] Fix case sensitivity of email address in gravatar hash

### LINKED ISSUE #847: Avatar is not always shown even though Gravatar avatar is present

Thatkookooguy:
- Solve vase sensitivity for gravatar avatars [resolves #847]
> based on https://en.gravatar.com/site/implement/hash/
> Currently, if I write `Neilkalman@gmail.com` or `neilkalman@gmail.com`, I don't get the same result. I made it so that every email is transformed to lowercase so that md5 will work

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ Use the following format for additions: ` - VERSION: [feature/patch (if applicab
 
 - 1.0.0: Introduced Continuous delivery. [#823](https://github.com/FredrikNoren/ungit/issues/823)
 
+## [1.0.1](https://github.com/FredrikNoren/ungit/compare/v1.0.0...v1.0.1)
+
+### Fixed
+- Fixed gravatar avatar fetch if email have different cases applied. [#847](https://github.com/FredrikNoren/ungit/issues/847)
+
 ## [1.0.0](https://github.com/FredrikNoren/ungit/compare/v0.10.3...v1.0.0)
 
 ### Added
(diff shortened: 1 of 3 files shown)


Thatkookooguy:
I'm not sure if I wrote the `CHANGELOG.MD` part in the right format. If not, tell me and I'll fix it :-)

thanks!

FredrikNoren:
Nice! Almost right with the changelog, this is how it should look:

```md
- 1.0.1: [patch] Fixed gravatar avatar fetch if email have different cases applied. [#847](https://github.com/FredrikNoren/ungit/issues/847)
- 1.0.0: Introduced Continuous delivery. [#823](https://github.com/FredrikNoren/ungit/issues/823)
```

(I.e. just adding a row above the 1.0.0 row)

Thatkookooguy:
@FredrikNoren changed the `CHANGELOG.MD` file as you requested 😄 

thanks!

Thatkookooguy:
@FredrikNoren all the tests passed on my local machine. Can you please re-run the tests on `appveyor` and `travis`?

codingtwinky:
I've been noticing several phantomjs failed errors with latest phantomjs, which is unfortunate and hopefully they will fix it soon.

Not sure how to retrigger windows test but I have more fixes to stabilize tests coming and hopefully that fixes.  But rest looks good though!

Thanks!!

Thatkookooguy:
@codingtwinky are you talking about my addition? because I thought about doing another PR to add the default hash gravatar they generate for users without an avatar.

should I put it somewhere else in order to not `cause some over head when many nodes are loaded`?

codingtwinky:
Ah, let me clarify.  what I meant is that `this.authorGravatar = ko.computed(function() {...` doesn't have to be `ko.computed` and many other variables often don't need to be `ko.computed` or `ko.observable`.  For this case, `authorGravatar` will not change once it is calculated at initialization.

I was just mentioning a larger code clean up effort that I would like to do.

Thatkookooguy:
@codingtwinky oh cool :-)

PR #841, opened 2016-12-24
Finalize promisify

codingtwinky:
**WIP do not merge yet**

* Finalize using promises in the client side
* Fix how error objects are handled in promise chain
* Removal of unused legacy callback api calls helpers

(this is actually getting more complicated than I anticipated...)

PR #852, opened 2017-01-21
Finalize promisify

codingtwinky:
Not sure how #841 was closed.

But here is continued effort of finalizing promises.  I may do more code cleanup later.


components/app/app.js
```diff
@@ -103,50 +103,40 @@ AppViewModel.prototype._handleRequestRememberRepo = function(event) {
...
-  this._isShowingCredentialsDialog = true;
-  diag.closed.add(function() {
-    self._isShowingCredentialsDialog = false;
-    programEvents.dispatch({ event: 'request-credentials-response', username: diag.username(), password: diag.password() });
-  });
 }
 AppViewModel.prototype.showDialog = function(dialog) {
   var self = this;
-  dialog.closed.add(function() {
+  // console.log("!!!", dialog.closeThen)
```
campersau: Remove.

components/dialogs/dialogs.js
```diff
@@ -37,15 +38,24 @@ components.register('TooManyFilesDialogViewModel', function(args) {
...
+DialogViewModel.prototype.closeThen = function(thenFunc) {
+  this.closePromise = this.closePromise.then(thenFunc);
+  return this;
 }
 DialogViewModel.prototype.setCloser = function(closer) {
   this.close = closer;
 }
-DialogViewModel.prototype.onclose = function() {
-  this.closed.dispatch();
+DialogViewModel.prototype.publish = function() {
```
campersau: What about `show` instead of `publish`?

You could also return the `closePromise` instead of `this`. So it would be used like this:

```js
diag.show().then(function() { /* closed */ });
```

And `closeThen` can be removed.

components/graph/git-ref.js
```diff
@@ -76,58 +76,54 @@ RefViewModel.prototype.dragEnd = function() {
...
   var self = this;
   var url = this.isTag ? '/tags' : '/branches';
   if (this.isRemote) url = '/remote' + url;
-  this.server.del(url, { path: this.graph.repoPath(), remote: this.isRemote ? this.remote : null, name: this.refName }, function(err) {
-    if (!err) {
-      self.node().removeRef(self);
-      self.graph.refsByRefName[self.name] = undefined;
-    }
 
+  return this.server.delPromise(url, { path: this.graph.repoPath(), remote: this.isRemote ? this.remote : null, name: this.refName }).then(function(err) {
```
campersau: The `err` parameter is not used (at the end of the line). Maybe you want an empty `catch`?

clicktests/test.discard.js
```diff
@@ -64,7 +64,10 @@ suite.test('Should be possible to discard a created file without warning message
 });
 
 suite.test('Shutdown', function(done) {
-  environment.shutdown(done, true);
+  environment.shutdown(function() {
+    page = webpage.create();
```
codingtwinky: It's a problem when API call is made and phantomjs is terminated promise gets resolved with an empty object.  It's a very odd behavior I don't see in actual chrome.  But because empty object doesn't have many of the expected properties it break test.  Maybe there is a proper way to refresh the page rather than shuttingdown and reinit we do.

components/dialogs/dialogs.js
codingtwinky: s/publish/show/

So I'm not a big fan of object context changing in middle of function chain, that sounds little confusing.  With this suggestion, function chains switch it's context to promise from dialog object after publish is called.  If we were to chain functions it's context should be consistent through out entire chain.

CHANGELOG.md
```diff
@@ -19,6 +20,9 @@ Use the following format for additions: ` - VERSION: [feature/patch (if applicab
 - Limit commit title to 72 characters, the rest is truncated and shown when inspecting the commit
 - Updated file watch logic to closely follow git commands in another process [#283](https://github.com/FredrikNoren/ungit/issues/283)
 
+### Changed
+- Refactored to use promises on client side [#852](https://github.com/FredrikNoren/ungit/pull/852)
```
FredrikNoren: I was thinking we just do the one line changes (that you added on top), more details can be left in the complementing issue/PR

components/graph/git-graph-actions.js
```diff
@@ -100,17 +99,17 @@ GraphActions.Reset.prototype.createHoverGraphic = function() {
...
+  return components.create('yesnodialog', { title: 'Are you sure?', details: 'Resetting to ref: ' + remoteRef.name + ' cannot be undone with ungit.'})
     .show()
     .closeThen(function(diag) {
       if (!diag.result()) return;
-      self.server.postPromise('/reset', { path: self.graph.repoPath(), to: remoteRef.name, mode: 'hard' })
-        .then(function() { context.node(remoteRef.node()); })
-    }).finally(callback)
+      return self.server.postPromise('/reset', { path: self.graph.repoPath(), to: remoteRef.name, mode: 'hard' })
+        .then(function() { context.node(remoteRef.node()); });
+    }).closePromise;
```
campersau: That is why I suggested to return the close promise in `show` https://github.com/FredrikNoren/ungit/pull/852#discussion_r97205602 😉

codingtwinky: how is this bad?

campersau: I haven't said it is bad. It is just more code to write.

codingtwinky: I'd rather have more code to be more explicit when context changes than silently changes context middle of function chain.

codingtwinky:
There are so many more refactoring I want to do but I will merge this now and break the PR apart as this is getting too big as it is...

Issue #851, opened 2017-01-20
Add option to skip version check


yebrahim:
It's useful to add a config to skip version checks, especially when running ungit as part of another project, where every new release has to be tested first.

FredrikNoren:
@yebrahim feel free to submit a PR

codingtwinky:
Can you clarify on what version check are you talking about?  are you talking about ungit version check or git version check?

yebrahim:
@codingtwinky The ungit version, it pops up in the UI when there's a version difference.

Issue #823, opened 2016-12-04
Continuous deployment (npm publish on merge)


FredrikNoren:
Hi all,

Ungit is a great community driven project, and I'm constantly impressed with all the stuff people are contributing back to this project.

However, I've noticed one big bottleneck to progress: Releases. The way they work today is a strictly manual process, so a person (usually me) has to at some point decided that "now seems like a good time to release" and then perform the release. There's a couple of problems with this:

1. Merge to deploy time is very long. Sometimes it's been months, and even if we try to be more diligent about releasing more often, it will still probably be weeks between a change being merged and users actually seeing that change.
2. This means valuable new features and fixes aren't getting out fast enough to users.
3. Master is "untested". Since only a handful of people use the master branch, it's very hard to catch problems. Often we don't discover bugs until we do a release. At that point it's very hard to know exactly what commit introduced the bug.

Therefore, I'd like to propose that we change to a continuous deployment release system instead of a manual one. On each commit or merge to master, the travis build would automatically calculate a new patch number and push the package to NPM.

There are some drawbacks to this approach, most of all users would be getting a lot more "it's time to upgrade" messages. To make this less distracting I'd suggest we update the UI of this message to something slightly more discreet, or perhaps something that highlights more the older your version is (if it's < 5 patches diff, do a small thing, if it's more do a big dialog).

It's a big change in "how we work" so I'd like to hear what people have to say before making this change (or if someone is interested in taking this on, let me know), so I'll leave this ticket open for a few weeks to gather feedback/input before moving forward.

cc @codingtwinky

Ajedi32:
Are you still planning to use SemVer? How will Travis know whether a particular change is SemVer-patch or SemVer-minor when incrementing the version number?

FredrikNoren:
@Ajedi32 This is how I've done it in other projects: Major and minor versions are manual, to bump them you have to change the actual number in package.json, and create a tag for them. All regular commits are patches, and the patch number is calculated as the number of commits since the last version git-tag.

Maybe there's a better way where a PR could be tagged major/minor version change and all of that would happen automatically though.. not sure.

codingtwinky:
I think it would be good as long as minor version publish requires manual intervention as that is important, which seems to be the direction we are going.

So I'm for this idea.

FredrikNoren:
[...]

PR #854, opened 2017-01-25
Add option to override ungit version check

yebrahim:
This PR adds a config `--ungitVersionCheckOverride` to ignore ungit version checks. It hides the notification message from the client side, and doesn't touch the version check functionality on the server side.

I didn't think this change warranted a new entry in the change log and bumping the patch version, please let me know if you think otherwise.

Fixes https://github.com/FredrikNoren/ungit/issues/851


FredrikNoren:
@yebrahim Sweet! Please increase the minor version in `package.json` (we're on continuous delivery so merges to master automatically publishes a new version, see https://github.com/FredrikNoren/ungit/issues/823) and just add a small line in `CHANGELOG.md` about the change.

yebrahim:
Awesome!

PR #860, opened 2017-02-03
Update "Toggle all" button after commit or changing selected files

### LINKED ISSUE #859: Update "Toggle all" button after commit or changing selected files
If the `Toggle all` button is used to uncheck all files before committing it switches to the unchecked state. After the commit it stays in that stage but all remaining files are automatically checked. This requires to press the button twice in order to unselect all files again.
This is also the case when manually (un)selecting all files.

Proposal: Update the `Toggle all` button after each commit and make it resemble the state "all files are checked". So when you manually unselect one file the `Toggle all` button should switch to the unchecked state.

I know this is only a minor thing but i found it really annoying.

schrej:
Compute the allStageFlag of the staging component by comparing the total and selected staging files.

Improves the behavior of the `Toggle all` button:
When all files are selected it will unselect all files.
If one or more files are unselected it will select all files.

This is my first time using knockout.js so if there are better ways of doing things, please point them out!

resolves #859

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 Use the following format for additions: ` - VERSION: [feature/patch (if applicable)] Short description of change. Links to relevant issues/PRs.`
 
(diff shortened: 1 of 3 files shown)


FredrikNoren:
Sweet!

PR #864, opened 2017-02-09
credential dialog fix

codingtwinky:


codingtwinky:
I have another PR coming soon so I'm intentionally not triggering the CD

Issue #871, opened 2017-02-16
--cliconfigonly doesn't affect


bes-internal:
--cliconfigonly not work. If I understand correctly rc must not be taken from ~/.ungitrc if agrument is specified
ungit 1.1.7

codingtwinky:
Can you provide command and .ungitrc you have tried?

bes-internal:
~/.ungitrc:
```javascript
{
    "logLevel": "debug",
    "port": 10222
}
```
Run ungit:
```
$ ungit 
Setting log level to debug
2017-02-21T09:12:50.364Z - info: Listening on port 10222
...
```

Then try this:
```
$ ungit --cliconfigonly
Setting log level to debug
2017-02-21T09:13:08.826Z - info: Listening on port 10222
...
```

PR #880, opened 2017-03-02
Fix cli args; Ignore ~/.ungitrc with syntax error

bes-internal:
fixes: #871
--cliconfigonly works now. Transparent override configs: default -> rc -> argv

Post to /api/userconfig writes in JSON, therefore support ini format can be confusing. 

It would be good if yargs started to support an arbitrary configuration and all of the standard paths as in rc. Then it would be possible to remove the rc dependency

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 Use the following format for additions: ` - VERSION: [feature/patch (if applicable)] Short description of change. Links to relevant issues/PRs.`
 
+- 1.1.11:
+    - Fix cli arguments [#871](https://github.com/FredrikNoren/ungit/pull/871)
+    - Stop if ~/.ungitrc contains syntax error
+    - Removed official support ini format of ~/.ungitrc, because internal API supports only JSON
 - 1.1.10: Fix broken diff out in some cases when diff contains table. [#881](https://github.com/FredrikNoren/ungit/pull/881)
 - 1.1.9: Fix around ubuntu's inability to cache promises. [#877](https://github.com/FredrikNoren/ungit/pull/878)
 - 1.1.8:
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "ungit",
   "author": "Fredrik Norén <fredrik.jw.noren@gmail.com>",
   "description": "Git made easy",
-  "version": "1.1.10",
+  "version": "1.1.11",
   "ungitPluginApiVersion": "0.2.0",
   "scripts": {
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 Use the following format for additions: ` - VERSION: [feature/patch (if applicable)] Short description of change. Links to relevant issues/PRs.`
 
+- 1.2.0:
+    - Fix cli arguments [#871](https://github.com/FredrikNoren/ungit/pull/871)
+    - Ignore ~/.ungitrc if contains syntax error
+    - Removed official support ini format of ~/.ungitrc, because internal API supports only JSON
```
codingtwinky: Ah, can we do 1.1.10?  I would like to reserve minor version bump to when stability is confirmed and code is more mature after code freeze.  I'd imagined @FredrikNoren wants that.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "ungit",
   "author": "Fredrik Norén <fredrik.jw.noren@gmail.com>",
   "description": "Git made easy",
-  "version": "1.1.9",
+  "version": "1.2.0",
```
codingtwinky: ditto

source/config.js
```diff
@@ -190,23 +191,36 @@ let argv = yargs
...
-} else if (argv.cliconfigonly) {
-  module.exports = argv.default(defaultConfig).argv;
-} else {
-  module.exports = rc('ungit', argv.default(defaultConfig).argv);
 }
+
+var rcConfig = {};
+if (!argvConfig.cliconfigonly) {
+  try {
+    var rcConfig = rc('ungit');
```
codingtwinky: is this intended to have `var` infront?

source/server.js
```diff
@@ -305,11 +305,11 @@ app.get('/api/gitversion', (req, res) => {
...
   return fs.isExists(userConfigPath).then((hasConfig) => {
-      if (!hasConfig) return {};
-      return fs.readFileAsync(userConfigPath, { encoding: 'utf8' }).then((content) => {
-          return JSON.parse(content.toString());
-        });
-    });
+    if (!hasConfig) return {};
+    return fs.readFileAsync(userConfigPath, { encoding: 'utf8' })
+      .then((content) => { return JSON.parse(content.toString()); })
+      .catch((err) => { winston.error(`Ignore .ungitrc because ${err}`); return {}; });
```
codingtwinky: If `.ungitrc` is malformed or errored, should we throw exception here to halt the entire system rather than starting the system?

codingtwinky:
👍 

Thank you for fixing this!!!  sorry it took so long to try it out...

Issue #887, opened 2017-03-22
Ungit stopped working on Chrome 57


leonadler:
The recent version of Chrome gives me the "Whooops - Something went wrong, reload the page to start over." message.
The same server & directory works in Firefox. Chrome private mode does not fix it either.
Accessing both via localhost or 127.0.0.1 do not work in chrome, both work in Firefox.

Versions:
* OS: Arch Linux
* chrome: 57.0.2987.110 (64-bit)
* firefox: 52.0 (64-bit)
* node: 6.10.1
* npm: 4.4.1
* ungit: 1.1.11

[My server.log](https://github.com/FredrikNoren/ungit/files/860942/server.log.txt) (via `ungit --logRESTRequests --logGitCommands --logGitOutput`)
[Chrome's browser console output](https://github.com/FredrikNoren/ungit/files/860943/console-output.log.txt).
Any other Information I can provide?

###### by the way: very nice work on ungit!

bes-internal:
Can't confirm on same chrome version on linux.

From your console output:
```
server.js:174 Unhandled Promise ERROR:  Object {errorCode: "cross-domain-error", error: Object} Promise {_bitField: 18087936, _fulfillmentHandler0: Object, _rejectionHandler0: undefined, _promise0: undefined, _receiver0: undefined…}
```
cast @codingtwinky

leonadler:
The error thrown seems to be caused by an XHR returning status 0 for `GET /api/log` in `server.js`:
```JavaScript
Server.prototype._httpJsonRequest = function(request, callback) {
  var httpRequest = new XMLHttpRequest();
  httpRequest.onreadystatechange = function() {
    // It seems like you can get both readyState == 0, and readyState == 4 && status == 0 when you lose connection to the server
    if (httpRequest.readyState === 0) {
      callback({ error: 'connection-lost' });
    } else if (httpRequest.readyState === 4) {
      var body;
      try {
        body = JSON.parse(httpRequest.responseText);
      } catch(ex) { body = null; }
      if (httpRequest.status == 0) {
      	callback({ error: 'connection-lost' });  // <--- error is thrown here
      }
      else if (httpRequest.status != 200) callback({ status: httpRequest.status, body: body, httpRequest: httpRequest });
      else callback(null, body);
    }
  }
  var url = request.url;
  if (request.query) {
    url += '?' + this._queryToString(request.query);
  }
  httpRequest.open(request.method, url, true);
[...]

Issue #874, opened 2017-02-22
Doesn't work with Chrome on MacOS


JRosanowski:
I guess real Mac users wouldn't have noticed this since they would use Safari but I've got Chrome installed on a Mac at work and couldn't get ungit to work. Chrome just comes up with the error "Network error - the requested URL could not be retrieved". After trying a whole bunch of things I finally thought maybe it was the browser. Sure enough it works fine with Safari. If it can't be fixed it might be worth a note somewhere that it doesn't work to save any other pseudo Mac users.

lfx:
Can't reproduce - running latest MacOS with the latest Chrome. All works fine.

codingtwinky:
@JRosanowski It works fine for me. Can you please share the URL you are trying along with the client console log?

JRosanowski:
I assume you mean the console log from Chrome?

http://localhost:8448/favicon.ico Failed to load resource: the server
responded with a status of 403 (Forbidden)
http://localhost:8448/#/repository?path=%2FUsers%2Fjrosano%2FDocuments%2FGit%2FGitTest
Failed to load resource: the server responded with a status of 403
(Forbidden)

codingtwinky:
Perhaps it's [ipv6 problem](http://stackoverflow.com/questions/32453984/cant-access-localhost-with-wamp-and-chrome)?  For mac, it is located in `/etc/apache2/httpd.conf`.

ylecuyer:
I have the same issue with chrome but not only on macos also on chromium linux.

The error seems to be a cors issue with keen.js and the /api/log

kaizvn:
I got this error : 
```
Unhandled rejection TypeError: First argument must be a string or Buffer
    at ServerResponse.OutgoingMessage.end (_http_outgoing.js:555:11)
    at /usr/local/lib/node_modules/ungit/src/server.js:191:9
    at tryCatcher (/usr/local/lib/node_modules/ungit/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/usr/local/lib/node_modules/ungit/node_modules/bluebird/js/release/promise.js:510:31)
    at Promise._settlePromise (/usr/local/lib/node_modules/ungit/node_modules/bluebird/js/release/promise.js:567:18)
    at Promise._settlePromise0 (/usr/local/lib/node_modules/ungit/node_modules/bluebird/js/release/promise.js:612:10)
[...]

Issue #892, opened 2017-03-29
Ungit crash on Firefox


dregad:
Greetings,

I am running Ungit on a Cloud9 workspace (Ubuntu 14.04). This has been working just fine for many months, but today I am getting _Whooops Something went wrong, reload the page to start over._  There haven't been any changes on the system . I tried upgrading Ungit (from 1.10.3 to 1.10.11), but the problem persists.

It's worth mentioning that the error occurs only with Firefox. When using Chrome Ungit loads just fine, no idea why.

I am not seeing any errors on the server side (with --logLevel=debug)
```
2017-03-29T14:45:06.406Z - info: Start watching /home/ubuntu/workspace/adodb recursively
2017-03-29T14:45:06.406Z - info: Start watching with .git and .git/refs/[heads|remotes|tags]
2017-03-29T14:45:09.447Z - info: emitting git-directory-changed to sockets, manually triggered
```

On the client, the console shows
```
App version: 1.1.11 mantis-dregad.c9users.io:8081:32:2
Initing raven mantis-dregad.c9users.io:8081:34:3
unreachable code after return statement ungit.js:19770:4
mutating the [[Prototype]] of an object will cause your code to run very slowly; instead create the object with the correct initial [[Prototype]] value using Object.create ungit.js:22079:9
unreachable code after return statement ungit.js:19770:4
Google analytics inited mantis-dregad.c9users.io:8081:914:3
Keen.io inited mantis-dregad.c9users.io:8081:921:3
Event: Object { event: "navigated-to-path", path: "/home/ubuntu/workspace/adodb" } ungit.js:50666:3
Event: Object { event: "navigation-changed", path: "repository?path=%2Fhome%2Fubuntu%2F…", oldPath: undefined } ungit.js:50666:3
Event: Object { event: "connected" } ungit.js:50666:3
Unhandled Promise ERROR:  Exception { message: "", result: 2153644038, name: "", filename: "https://mantis-dregad.c9users.io:80…", lineNumber: 5569, columnNumber: 0, data: null, stack: "require<[42]</Server.prototype._htt…", 2 de plus… } Object { _bitField: 16777216, _fulfillmentHandler0: Exception, _rejectionHandler0: undefined, _promise0: undefined, _receiver0: undefined, _trace: Object } ungit.js:5667:5

Event: Object { event: "git-crash-error" } ungit.js:50666:3
Event: Object { event: "current-remote-changed", newRemote: "origin" } ungit.js:50666:3
[...]

PR #893, opened 2017-03-30
Fix cached promised values not resolving properly in some platforms.  #874 #878 #887

### LINKED ISSUE #882: Commit message lost when git returns error code
When committing something using ungit, and the commit fails in some way with git returning a non-zero error code, the commit message (title and body) entered into ungit is removed.  

This is particularly frustrating when ungit is used with pre-commit hooks.  These hooks modify code and use the non-zero git return code to allow you to inspect and re-commit the automatic changes.  

In an older version of ungit (not sure which, I could bisect), when git / pre-commit hooks returned non-zero code, the commit message was preserved in the ungit UI.  This allowed me to simply press the commit button and see if the pre-commit hooks did anything and press commit again if they did.  

In the current version, if the pre-commit hooks fires, I have to retype my commit message.  This is a bit cumbersome, and I have worked around it at the moment by copying my message before committing in case the hook fires and I have to paste in the text again.

codingtwinky:
* add readme for #887
* fix rare edge case where remote ref node is gone on reset operation
* rescursively resolve all promises before caching them.  #878 #874
* Retain commit messages when commit fails fix #882


README.md
```diff
@@ -99,6 +99,7 @@ Known issues
 * If you're running MacOSX Mavericks and Ungit crashes after a few seconds; try updating npm and node. See [#259](https://github.com/FredrikNoren/ungit/issues/259) and [#249](https://github.com/FredrikNoren/ungit/issues/249) for details.
 * Ubuntu users may have trouble installing because the node executable is named differently on Ubuntu, see [#401](https://github.com/FredrikNoren/ungit/issues/401) for details.
 * Debian Wheezy's supported git and nodejs packages are too old, therefore download newest [git](https://github.com/git/git/releases) and [nodejs](https://nodejs.org/download/) tarballs and [build from source](http://www.control-escape.com/linux/lx-swinstall-tar.html).
+* Adblocker may block Ungit! Some ad blockers, such as uBlock, doesn't like localhost api calls and assumes that it is a cross domain attack.  Please whitelist `localhost:{ungit port number}`. [#887](https://github.com/FredrikNoren/ungit/issues/887)
```
dregad: Suggested changes:
- rewording:`Some ad blockers, such as [Adblock plus](https://adblockplus.org) and [uBlock](https://www.ublock.org/), don't like localhost api calls and assume that [...]`
- add reference to `[#892](https://github.com/FredrikNoren/ungit/issues/892)`

Also, note that the host name it is not necessarily *localhost* (in my case, running ungit on Cloud9 workspace, I have a *workspace-name.c9users.io* hostname), so your hint could be misleading.

source/utils/cache.js
```diff
@@ -24,6 +23,21 @@ cache.resolveFunc = (key) => {
...
+ * @description In Linux, or certain settings, it seems that cached promises
+ *   are not able to resolved and we need to cache raw result of promieses.
+ * @param {prom} - raw value or promise to be returned or resolved
+ * @param {promise} - a promise where next "then" will result in raw value.
+ */
+const getHardValue = (prom) => {
+  if (prom.then) {
+    return prom.then(getHardValue);
+  } else {
+    return Bluebird.resolve(prom);
```
campersau: Strange... normally [`.resolve`](http://bluebirdjs.com/docs/api/promise.resolve.html) should already do this:
> If value is a thenable (Promise-like object, like those returned by jQuery's $.ajax), returns a trusted Promise that assimilates the state of the thenable.

codingtwinky: yeah this is a terrifying find and i need more confirmation but behavior was replicate-able in ubuntu.

README.md
```diff
@@ -99,6 +99,7 @@ Known issues
 * If you're running MacOSX Mavericks and Ungit crashes after a few seconds; try updating npm and node. See [#259](https://github.com/FredrikNoren/ungit/issues/259) and [#249](https://github.com/FredrikNoren/ungit/issues/249) for details.
 * Ubuntu users may have trouble installing because the node executable is named differently on Ubuntu, see [#401](https://github.com/FredrikNoren/ungit/issues/401) for details.
 * Debian Wheezy's supported git and nodejs packages are too old, therefore download newest [git](https://github.com/git/git/releases) and [nodejs](https://nodejs.org/download/) tarballs and [build from source](http://www.control-escape.com/linux/lx-swinstall-tar.html).
+* Adblocker may block Ungit! Some ad blockers, such as [Adblock plus](https://adblockplus.org) and [uBlock](https://www.ublock.org/), doesn't like localhost api calls and assumes that it is a cross domain attack.  Please whitelist `{localhost|127.0.0.1|$UngitURL}:{ungit port number}`. [#887](https://github.com/FredrikNoren/ungit/issues/887) [#892](https://github.com/FredrikNoren/ungit/issues/892)
```
dregad: @codingtwinky thanks for updating the README. If I may pick nits, you missed one little bit `doesn't` --> `don't`.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ Use the following format for additions: ` - VERSION: [feature/patch (if applicab
     - Fix cli arguments [#871](https://github.com/FredrikNoren/ungit/pull/871)
     - Stop if ~/.ungitrc contains syntax error
     - Removed official support ini format of ~/.ungitrc, because internal API supports only JSON
+    - Retain commit messages when commit fails [#882](https://github.com/FredrikNoren/ungit/pull/882)
```
FredrikNoren: 1.1.11 is already out right? Think you ment to cut a new version here

PR #904, opened 2017-04-17
Randomize click test port

codingtwinky:
Randomize click test port

### DIFF
--- a/clicktests/helpers.js
+++ b/clicktests/helpers.js
@@ -1,4 +1,6 @@
 var helpers = exports;
+var Bluebird = require('bluebird');
+var serverPort = 45062; // it's like between port side and starboard...s
 
 helpers.log = function(text) {
   console.log((new Date()).toISOString(), text);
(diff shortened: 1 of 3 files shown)


clicktests/helpers.js
```diff
@@ -105,3 +107,7 @@ helpers.selectAllText = function(page) {
   helpers.log('Trying to select all in focused element (ctrl-A)');
   page.sendEvent('keypress', page.event.key.A, null, null, 0x04000000 );
 }
+helpers.getPort = function() {
```
FredrikNoren: https://www.npmjs.com/package/get-port might be a simpler way to do this

codingtwinky: That is very similar to what I have tried: https://gist.github.com/mikeal/1840641, which I liked it better because it's less dependancy.   However within phantomjs client code, `net` package is unavailable.  phantom has [webserver](http://phantomjs.org/api/webserver/) module but its has "experimental" tag.  

Thus I've went with try and retry with random port.

FredrikNoren: Oh right, forgot this is in phantom. We should really consider trying nightmare...

PR #907, opened 2017-04-21
reporting commit error in the UI

CurtisHumphrey:
It helpful for the error to be reported in the UI vs having to view the error in the console.  Pre-commit hooks are a common source of these type of errors


codingtwinky:
There was a reason why this was the case.  I can't remember it from top of my head right now.  But I will be testing this.  It may very well be the case that this is one of the old behavior that needs revisit. 

Thanks for pointing this oddity out.

codingtwinky:
For this PR, I think it would be better if we delete the catch block and let the [global unhandled promise rejection function](https://github.com/FredrikNoren/ungit/blob/master/public/source/server.js#L158-L178) to handle it instead.  With this change, if git error, output the git error message same as you have.  But will crash if non git error is returned, which is not a bad idea.

jung-kim:
close for inactivity

PR #908, opened 2017-04-23
Clicktest refactoring with Nightmare

codingtwinky:
As some may know current click test has several problems.

* it is bit hard to write and confusing 
* there are some intermittent test failures

I doubt this is the solution but I think it addresses or helps addresses above pain points.  to run new click test run `grunt nmclicktest`.  

here are some notes:

1. No longer using phantomJS and custom spun test runners.  Uses nightamre and mocha
2. Using promises.  (we talked about using es6 async await but moch framework had nice implementation that works well with promises)
3. Did consider [`nightwatch.js`](http://nightwatchjs.org/) and had a working version but it was bit more complex than I was hoping for.  But should we want more comprehensive testing in different browser flavors, versions and etc with selenium we can do that instead of simplistic approach in this PR)
4. I haven't done research but I think mocha can do file level test parallelization.  This is not a big deal but it helps a lot to do run tests in parallel during dev.

built on top of #905


nmclicktests/spec.authentication.js
```diff
@@ -0,0 +1,39 @@
...
+});
+
+describe('test authentication', () => {
+  before('Environment init', () => {
+    return environment.init();
+  });
+
+  it('Open home screen should show authentication dialog', () => {
+    return environment.goto(environment.url)
+      .wait('[data-ta-container="login-page"]');
```
FredrikNoren: I know I used to have a different position, but I'm wondering if we shouldn't just move a away from the data-ta-container stuff and just use css classes instead?

codingtwinky:
@FredrikNoren It is my plan to do exactly that, a file by file conversion.  I'm having some hard time with promises not being resolved in ubuntu and once that is fixed I can do exactly that.

codingtwinky:
Welp, [this](https://github.com/electron/electron/issues/9323) is annoying...

codingtwinky:
@FredrikNoren wow this PR has going on long enough and I figure I will post my discoveries so far.

1. Test fully passes on mac.  
2. nightmare js doesn't work on ubuntu.  https://github.com/segmentio/nightmare/issues/224
3. WIndows seems to be having issue with branch test and I think this is due to one of the test state setting tests failed silently and git state does not match the expected for `cherrypick fail case` test.
4. as far as removing `data-ta` tags, I will push it for later days as it would touch all tests for some cases.  (since I don't have windows machine, its really hard for me to capture screenshots between steps and confirm/fix)
5. tests are so much faster now...
6. there are some nightmare and es6 specific quirks such as "this" context is different when you use arrow or normal functions, `then` will return a promise rather than nightmare object and etc.  But I like nightmarejs 1000x better than other solutions.

Until bullet 2 and 3 is resolved I guess this PR will be dormant...

FredrikNoren:
Very nice!! Comments:

1. 👍
2. It should work on ubuntu, I'm running it currently :) We have to add some stuff to our travis file though, see this: https://github.com/electron/electron/blob/master/docs/tutorial/testing-on-headless-ci.md
3. Which test is failing on windows? Can we disable it on windows for now so we can get this merged and create a separate ticket for it?
4. Yup for sure, I think let's have a policy of moving away from them and eventually we'll get there.
5. 🎉
6. Not sure I follow, are you saying arrow functions are or aren't working according to [spec](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions)? Either way happy to hear it feels like a step forward. I've been using nightmare myself for the past few months, and to be honest it does have a lot of quirks so def agree with the assessment that this doesn't feel like the "solution", but at least a step forward.

FredrikNoren:
Oh and noticed you removed the auth test but not the other ones you migrated. Any reason to keep them around? Otherwise I think we should just remove them too.

nmclicktests/spec.branches.js
```diff
@@ -0,0 +1,109 @@
...
+      .ug.click('[data-ta-clickable="checkoutbranch-1"]')
+      .ug.waitForElementNotVisible('[data-ta-clickable="branch"][data-ta-element="progress-bar"]')
+  });
+
+  it('cherrypick fail case', () => {
+    return environment.nm.ug.click('[data-ta-clickable="node-clickable-0"]')
+      .ug.log("clicked first node")
+      .ug.click('[data-ta-action="cherry-pick"][data-ta-visible="true"] [role=button]')
+      .ug.log("cherrypick button clicked")
+      .ug.click('[data-ta-action="abort"]')
```
campersau: If you wait before this then this will pass on windows. (I know you already wait inside of click but it looks like windows needs more time.)
e.g. `.wait(500)`

codingtwinky: Thanks for the tip! looks like its working now!  😀

codingtwinky:
@FredrikNoren There might be intermittent test failures but I will fix it later.  And here are the test config changes that you might need to be aware of.

1. Dropped support for node 4.x, (lets get out of stone age)
2. changes node 6 test to 6.11.x test as it is the latest of the 6 and one should really use the latest
3. Added node 7.10 and the latest node version, currently at 8.1.2.
4. Doing git edge test with the latest node version only.  

Please let me know once review is done!

nmclicktests/spec.branches.js
```diff
@@ -0,0 +1,109 @@
...
+  it('cherrypick success case', () => {
+    return environment.nm.ug.click('[data-ta-clickable="node-clickable-1"]')
+      .ug.click('[data-ta-action="cherry-pick"][data-ta-visible="true"] [role=button]')
+      .wait(500)
+      .visible('[data-ta-action="abort"]')
+      .then((isVisible) => {
+        if (isVisible) {
+          throw new Error("Cherry pick errored when success was expected.");
+        }
+        return;
```
campersau: Nit: not needed.

nmclicktests/environment.js
```diff
@@ -0,0 +1,269 @@
...
+      `--autoShutdownTimeout=${this.config.serverTimeout}`,
+      '--logLevel=debug',
+      '--maxNAutoRestartOnCrash=0',
+      '--no-autoCheckoutOnBranchCreate',
+      '--alwaysLoadActiveBranch',
+      '--logGitCommands']
+      .concat(this.config.serverStartupOptions);
+    const ungitServer = child_process.spawn('node', options);
+    ungitServer.stdout.on('data', (data) => {
+      if (this.config.showServerOutput) console.log(prependLines('[server] ', data.toString()));
```
campersau: Nit: you are calling `data.toString()` three times in this method, maybe introduce a new variable.

nmclicktests/environment.js
```diff
@@ -0,0 +1,269 @@
...
+          console.log('Ungit started twice, probably crashed.');
+        } else {
+          this.hasStarted = true;
+          console.log('Ungit server started.');
+        }
+      }
+    });
+    ungitServer.stderr.on("data", (data) => {
+      console.log(prependLines('[server ERROR] ', data.toString()));
+      if (data.indexOf("EADDRINUSE") > -1) {
```
campersau: In the method above you are always calling `toString` before calling `indexOf` maybe you should do it here too? (And use a separate variable.)

nmclicktests/environment.js
```diff
@@ -0,0 +1,269 @@
...
+        if (this.hasStarted) {
+          console.log('Ungit started twice, probably crashed.');
+        } else {
+          this.hasStarted = true;
+          console.log('Ungit server started.');
+        }
+      }
+    });
+    ungitServer.stderr.on("data", (data) => {
+      console.log(prependLines('[server ERROR] ', data.toString()));
```
campersau: `console.error`?

nmclicktests/environment.js
```diff
@@ -0,0 +1,269 @@
...
+      .then(() => {
+        if (!this.hasStarted) {
+          return Bluebird.resolve()
+            .delay(50)
+            .then(() => { return this.ensureStarted(); });
+        }
+      });
+  }
+
+  init(isSkipTempCreate) {
```
campersau: Unused.

nmclicktests/spec.authentication.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+const expect = require('expect.js');
+const testuser = { username: 'testuser', password: 'testpassword' }
+const environment = require('./environment')({
+  serverStartupOptions: ['--authentication', `--users.${testuser.username}=${testuser.password}`],
+  showServerOutput: true
+});
+
+describe('test authentication', () => {
+  before('Environment init without temp folder', () => environment.init(true));
```
campersau: `true` in `init(true)` is unused.

nmclicktests/environment.js
```diff
@@ -0,0 +1,269 @@
...
+    });
+  }
+
+  ensureStarted() {
+    return Bluebird.resolve()
+      .then(() => {
+        if (!this.hasStarted) {
+          return Bluebird.resolve()
+            .delay(50)
+            .then(() => { return this.ensureStarted(); });
```
campersau: Nit: `return` is not needed if brackets are omitted.

nmclicktests/spec.bare.js
```diff
@@ -0,0 +1,23 @@
...
+
+describe('test bare repo', () => {
+  before('Environment init', () => {
+    return environment.init()
+      .then(() => environment.createRepos([{ bare: true }]))
+      .then((repos) => testRepoPaths = repos);
+  });
+  after('Environment stop', () => environment.shutdown());
+
+  it('Open path screenz', () => {
```
campersau: `screenz`? In the other file you are calling it `screen`.

nmclicktests/spec.branches.js
```diff
@@ -0,0 +1,109 @@
...
+      .ug.waitForElementNotVisible('[data-ta-clickable="branch"] [data-ta-element="progress-bar"]')
+      .wait(500);
+  });
+
+  it('add a commit', () => {
+    return environment.nm.ug.createTestFile(`${testRepoPaths[0]}/testfile2.txt`)
+      .ug.commit('commit-3');
+  });
+
+  it('checkout chery pick base', () => {
```
campersau: Nit: `cherrypick`

nmclicktests/spec.commands.js
```diff
@@ -0,0 +1,72 @@
...
+  it('add a branch-2', () =>{
+    return environment.nm.ug.createTestFile(`${testRepoPaths[0]}/testfile.txt`)
+      .ug.commit('commit-1')
+      .wait('.commit')
+      .ug.createBranch('branch-2');
+  });
+
+  it('test branch create from command line', () => {
+    return environment.nm.ug.gitCommand({ command: ["branch", "gitCommandBranch"], repo: testRepoPaths[0] })
+      .then((cc) => environment.nm.wait('[data-ta-name="gitCommandBranch"]'));
```
campersau: Nit: argument `cc` is unused.

nmclicktests/environment.js
```diff
@@ -0,0 +1,269 @@
...
+    });
+  }
+
+  shutdown(doNotClose) {
+    this.shuttinDown = true;
+    return this.nm.ug.backgroundAction('POST', `${rootUrl}/api/testing/cleanup`, undefined)
+      .ug.shutdownServer()
+      .then(() => {
+        if (!doNotClose) {
+          this.nm.end();
```
campersau: `end()` returns a promise so you should add an `return` here.

nmclicktests/environment.js
```diff
@@ -255,9 +256,10 @@ class Environment {
         }
       }
     });
-    ungitServer.stderr.on("data", (data) => {
-      console.log(prependLines('[server ERROR] ', data.toString()));
-      if (data.indexOf("EADDRINUSE") > -1) {
+    ungitServer.stderr.on("data", (stderr) => {
+      const stderrStr = stderr.toString();
+      console.err(prependLines('[server ERROR] ', stderrStr));
```
campersau: `console.error`

`console.err is not a function`

codingtwinky: Yup!, I was too haste and it is now fixed in `d130a7a`.

Thanks for catching my sloppiness and doing code review!

FredrikNoren:
@codingtwinky Yeah that sounds good.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ Use the following format for additions: ` - VERSION: [feature/patch (if applicab
 - 1.1.19:
   - Ask before deleting a stash
   - Initial refactoring of click test using nightmare and mocha
+  - **Dropping support for node 4.x.**
+  - **Adding tests with node 7.x and the latest node. (currently 8.x)**
```
FredrikNoren: Feels wrong to add changes to something that's already out :) Maybe either bump the version or do a line of `- Internal/No release` with the changes?

.travis.yml
```diff
@@ -1,9 +1,10 @@
 sudo: required
 language: node_js
 node_js:
-  - "4.1"
   - "5.3"
```
FredrikNoren: Let's drop 5.3 too. LTS feels like our minimum.

codingtwinky:
Windows test failed on intermittent test failure within existing click test frame work not nightmare click test one. merging.... yay~~

Gruntfile.js
```diff
@@ -77,7 +77,7 @@ module.exports = function(grunt) {
           reporter: 'spec',
           timeout: 15000
         },
-        src: 'nmclicktests/*.js'
+        src: 'nmclicktests/spec.commands.js'
```
campersau: Why no more `*`?

codingtwinky: It was a mistake by me that slipped through.  It will be fixed in next round of nm clicktests coming in.  😭

FredrikNoren:
Looks like the version in package.json wasn't bumped. I'll fix

PR #917, opened 2017-05-10
repo list along config

bes-internal:
- [x] remove deprecated visitedRepositories
- [x] show home screen repo list from config and local storage
<del>read/write repo list from/to config</del>
<del>initial protected, readonly config repo list for server (shared) installation</del>
- [ ] tests
 
I gather opinions


FredrikNoren:
Seems reasonable, only reason it was stored in localStorage was for the "multiuser" use case, (many users logging in to the same server with multiple git repos) but I'm not sure if anyone is doing that, or if they are if they aren't sharing repos anyway on that machine.

bes-internal:
This is just my case. There is a set of repositories that it is difficult to type each time on the way. These are the configs on the servers. And also the QA department, which does not want to add repositories as remote each time

PR #921, opened 2017-05-22
Ask before deleting stash

alterratz:
I find it very irritating that you are not asked before deleting a stash ....
However I did not test this with auto-stashing since I never used this feature.
Also I only used yesno and not yesnomute since I am no JS expert and did not know there to store the mute timestamp.


FredrikNoren:
Can you bump the patch version in package.json and add a line in the CHANGELOG.md? Every merge to master is deployed (see https://github.com/FredrikNoren/ungit/blob/master/CONTRIBUTING.md#pull-requests).

alterratz:
done.

PR #924, opened 2017-05-28
Make 'git' command configurable via .ungitrc

yukoff:
It especially useful on windows, when git (eg. portable) is not on `PATH`

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,9 @@ Ungit brings user friendliness to git without sacrificing the versatility of git
 
  * Clean and intuitive UI that makes it easy to _understand_ git.
  * Runs on any platform that node.js & git supports.
- * Web-based, meaning you can run it on your cloud/pure shell machine and use the ui from your browser (just browse to http://your-cloud-machine.com:8448).
+ * Web-based, meaning you can run it on your cloud/pure shell machine and use the ui from your browser (just browse to <http://your-cloud-machine.com:8448>).
  * Works well with GitHub.
- * [Gerrit](https://code.google.com/p/gerrit/) integration through plugin: https://github.com/FredrikNoren/ungit-gerrit
+ * [Gerrit](https://code.google.com/p/gerrit/) integration through plugin: <https://github.com/FredrikNoren/ungit-gerrit>
 
 [Follow @ungitui on twitter](https://twitter.com/ungitui)
 
@@ -52,10 +52,33 @@ Example of `~/.ungitrc` configuration file to change default port and enable bug
 
 ```json
 {
-	"port": 8080,
-	"bugtracking": true
+  "port": 8080,
+  "bugtracking": true,
+  "gitCommand": null
 }
 ```
+
+Normally you shouldn't configure much, but in case your `git` command is not on `$PATH` you could set `gitCommand` parameter in `~/.ungitrc` (or use `--gitCommand` runtime parameter). If set to null or empty string - the default `git` command will be used.
+```json
+{
+  "gitCommand": "/path/to/your/git"
+}
+```
+
+On Windows `gitCommand` should be set like the following (remember of escaping backslashes):
+```json
+{
+  "gitCommand": "D:\\Apps\\Git\\bin\\git"
+}
+```
+or this way
+```json
+{
+  "gitCommand": "D:/Apps/Git/bin/git"
+}
+```
+
+You should be **warned**, however, that usage of this setting **is not recommended**. You should better **consider having `git` on your `$PATH`**.
 
 External Merge Tools
 --------------------
@@ -63,7 +86,7 @@ If you have your own merge tool that you would like to use, such as Kaleidoscope
 
 Auto Refresh
 ------------
-Ungit will watch git directory recursively upon page view and automatically refresh contents on git operations or changes on files that are not configured to be ignored in `.gitignore`.  
+Ungit will watch git directory recursively upon page view and automatically refresh contents on git operations or changes on files that are not configured to be ignored in `.gitignore`.
 
 One caveat is that node's [`fs.watch()`](https://nodejs.org/docs/latest/api/fs.html#fs_fs_watch_filename_options_listener) with `recursive: true` option is only available in Mac and Windows.  For non Mac and Windows machines, git operations will be automatically refreshed but file changes may require manual refreshes as `fs.watch()` is unable to detect changes within nested directory hierarchy.
 
--- a/source/config.js
+++ b/source/config.js
@@ -8,6 +8,8 @@ const homedir = require('os-homedir')();
 const winston = require('winston');
 const child_process = require('child_process');
 
+const isWindows = /^win/.test(process.platform);
+
 const defaultConfig = {
 
   // The port ungit is exposed on.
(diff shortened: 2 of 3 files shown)


source/config.js
```diff
@@ -235,6 +239,25 @@ if (typeof currentRootPath !== 'string') {
...
+let currentGitCommand = module.exports.gitCommand;
+if (typeof currentGitCommand !== 'string') {
+  currentGitCommand = '';
+} else if (currentGitCommand !== '') {
+  // must start with a slash
+  if (currentGitCommand.charAt(0) !== '/') {
+    currentGitCommand = '/' + currentGitCommand;
+  }
+  // can not end with a trailing slash - it should be a command
+  if (currentGitCommand.charAt(currentGitCommand.length - 1) === '/') {
```
codingtwinky: instead of checking for this.  Can we literally concatenate "/git" at the end?  Also, it would be `gitPath` instead of `gitCommand`.  I would like to limit the possibility that ungit can execute non git commands.  

Furthermore, considering this is a crux of all operations, I would like to reject on invalid inputs rather than adjust on invalid input as well to make the intent explicit.

source/config.js
```diff
@@ -235,6 +239,25 @@ if (typeof currentRootPath !== 'string') {
...
+    currentGitCommand = currentGitCommand.substring(0, currentGitCommand.length - 1);
+  }
+}
+// finally set default if empty
+if (currentGitCommand === '') {
```
codingtwinky: This will never be hit right due to if claus at the line 247 above?  if null or blank is passed in, it would be '/' and this will never be hit right?

Also, despite the documentation, I'm not sure where window's pathing is being accepted.  Your example will result in `/D:\\Apps\\Git\\bin\\git` right?  Am I missing somehting?

README.md
```diff
@@ -52,8 +52,25 @@ Example of `~/.ungitrc` configuration file to change default port and enable bug
 
 ```json
 {
-	"port": 8080,
-	"bugtracking": true
+  "port": 8080,
+  "bugtracking": true,
+  "gitCommand": "D:\\Apps\\Git\\bin\\git"
```
codingtwinky: Can we make it explicit that `gitCommand` or `gitPath` being null is a valid config?

source/config.js
```diff
@@ -249,16 +272,17 @@ Object.defineProperty(Error.prototype, 'toJSON', {
...
+  winston.error('Can\'t run "' + module.exports.gitCommand + ' --version". Is git installed and available in your path?', e.stderr);
+  winston.warn('Alternatively you can set full path to git command via "gitCommand" setting in .ungitrc', e.stderr);
   throw e;
 }
 
 module.exports.ungitPackageVersion = require('../package.json').version;
 
 if (fs.existsSync(path.join(__dirname, '..', '.git'))){
-  const revision = child_process.execSync('git rev-parse --short HEAD', { cwd: path.join(__dirname, '..') })
+  const revision = child_process.execSync(module.exports.gitCommand + ' rev-parse --short HEAD', { cwd: path.join(__dirname, '..') })
```
codingtwinky: Inorder for this PR to use custom pathed `git` command, https://github.com/FredrikNoren/ungit/blob/master/source/git-promise.js#L20 also needs to be updated.

yukoff: Overlooked it :) Thanks for pointing - somehow it worked when I tested, seems I started ungit with modifed PATH. Confirmed today it doesn't work without fixing that.

source/config.js
yukoff: I see the point. Well, initially it was `gitCommandPath`, but then I recalled a [JGit](https://wiki.eclipse.org/JGit/User_Guide#Running_the_JGit_CLI) - and decided to go with the `gitCommand` instead.

source/config.js
yukoff: Why? If set to `null` or empty string it will get right there - maybe I'm missing something, but this is the case I've added that final check. It is actually outside of that `if` (on lines 243..254).

source/config.js
```diff
@@ -220,21 +226,42 @@ module.exports = argv.default(defaultConfig).default(rcConfig).argv;
...
+
 let currentRootPath = module.exports.rootPath;
 if (typeof currentRootPath !== 'string') {
   currentRootPath = '';
 } else if (currentRootPath !== '') {
-  // must start with a slash
-  if (currentRootPath.charAt(0) !== '/') {
-    currentRootPath = '/' + currentRootPath;
+  // must start with a slash (on unix-like systems)
+  if (currentRootPath.charAt(0) !== fileSeparator && isWindows !== true) {
```
yukoff: This change is not mandatory, everything works fine without it. On the other hand, there's no point in adding root slash on windows.

BTW, as ungit requires node >= 4 this could be replaced by
```javascript
if (!path.isAbsolute(currentRootPath) { .. }
```
As far as I know prior node 4 `path.isAbsolute()` was not always returning boolean (behaved differently on win and unix).

source/config.js
yukoff: As for paths - on windows node accepts either `D:\Some\Path` or `D:/Some/Path` (double backslashes are to escape special char, which backslash is), also AFAIK on windows node will strip leading slash.

> Your example will result in `/D:\\Apps\\Git\\bin\\git` ...

Actually it will be `/D:\Apps\Git\bin\git` inside the code - if not escaped in config string will become `D:AppsGiingit`

source/config.js
yukoff: > Furthermore, considering this is a crux of all operations, I would like to reject on invalid inputs rather than adjust on invalid input as well to make the intent explicit.

In current patch set I left this part as is - how the setting should be checked? I see it like:
* check if a file exists
* check if it is a file, not a directory
* throw exception otherwise

And surely get rid of those adjustments.

source/config.js
```diff
@@ -104,6 +106,10 @@ const defaultConfig = {
   // "pluginConfigs": { "gerrit": { "disabled": true } }
   pluginConfigs: {},
 
+  // full path to git executable. If not set, system default will be used
+  // (in this case, git should be available on PATH)
+  gitCommand: null,
```
yukoff: It might be also useful to add it as an option - for example, in [Brackets](http://brackets.io) editor one could use setting from `brackets-git` extension config and pass it over to ungit, instead of setting it via `.ungitrc`. Should we have that added?

yukoff: Probably it might also be useful to set `gitSshCommand` - I'm currently on windows and it is not very convenient to set system-wide environment variables :(

Simple use-case: on windows `putty` is widely used ssh client, which is a GUI. It also has `plink` command, which is a cli command and meant to be a drop-in replacement for `ssh` on windows (supports same options as latter). And the last - the `pageant` or the putty authentication agent, which allows to add your keys and get them delivered to either `putty` or `plink` on request (nothing new here - actually they just reflect standard ssh client features).

So despite I have bash/git/ssh/etc on windows (thanks to [msys2](http://www.msys2.org/)), it's more convenient to use `plink` instead of standard `ssh` (which just doesn't work with pageant).

codingtwinky:
I'm beginning to think having `git` included in path as a prerequisit and document it in the readme, as it should be anyway.  

I'm going to relay this to @FredrikNoren...

jung-kim:
close for inactivity

PR #931, opened 2017-07-04
Nightmare

codingtwinky:
* Finalize nightmare
* remove old clicktests and phantomjs


package.json
```diff
@@ -85,7 +85,6 @@
     "istanbul": "~0.4.5",
     "mocha": "~3.3.0",
     "nightmare": "~2.10.0",
-    "phantomjs-prebuilt": "~2.1.14",
```
campersau: Also remove
```js
var phantomjs = require('phantomjs-prebuilt');
```
from Gruntfile.js

codingtwinky:
Merging, did a couple of reruns and it seems to be passing well and I'm not seeing intermittent test failures yet.  

Couple of things to note and some of this maybe be converted to issues later after more considerations:

* `data-ta*` tag removal to be replace by css selector is to follow
* Using `nightmare.insert` on typing navigation path on some of the tests.  This should be replaced with `nightmare.type` as it is more realistic and doesn't require passing in `\u000d` as "enter" char.
* Parallel click test execution to speed up dev process is missing.  (Should be easy as ports are not shared and had that in mind)
* this is bit minor but, one of the terrible things about `clicktests` is that a test is often depends on the success of previous test.  Previously we were able to terminate the test on failure, but mocha doesn't have and one test fail can cause massive slew of other false negative failures.
* Really should have gave more thought for deferred, but due to my own lack of experience with it and javascript in general didn't give it a faire shot during this refactoring... :(

PR #930, opened 2017-06-24
fixes #929 - Added a config setting to allow setting the default diff type

srfrnk:
Updated the readme as well.

PR #934, opened 2017-07-07
Fix #930

FredrikNoren:


srfrnk:
sorry i thought the pulll request would automatically update with those :) 10x.

PR #949, opened 2017-08-09
Parallel click test

codingtwinky:
Add parallel click test for speeding up dev efforts.

**Tests may time out depends on the parallelization level and resource allocation.** tried to minimize those variables via removing some of the unnecessary heavy network IOs but it is not gurantted.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -399,7 +399,7 @@ module.exports = function(grunt) {
           return file.startsWith("spec.");
         });
       }).then(function(tests) {
-        grunt.log.writeln('Running click tests in parallel... (this will take a while...) \t');
+        grunt.log.writeln('Running click tests in parallel... (this will take a while...)');
         return Bluebird.map(tests, function(file) {
           var output = "";
           var outStream = function(data) { output += data; }
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -381,6 +381,59 @@ module.exports = function(grunt) {
     })
   });
 
+  var maxConcurrency = 5;
+  /**
+   * Run clicktest in parallel at test suite level.
+   * This test does intermittently fails depends on the maxConcurrency level set
+   * above and the capacity of the computer as sometimes lack of resource allocation
+   * triggers timeouts.
+   * Use at own discretion.
+   */
+  grunt.registerTask('clickParallel', 'Parallelized click tests.', function() {
+    var done = this.async();
+
+    fs.readdirAsync('./nmclicktests')
+      .then(function(files) {
+        return files.filter(function(file) {
+          return file.startsWith("spec.");
+        });
+      }).then(function(tests) {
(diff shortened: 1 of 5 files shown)


components/commit/commit.js
```diff
@@ -32,9 +32,7 @@ function CommitViewModel(gitNode) {
   this.numberOfAddedLines = ko.observable();
   this.numberOfRemovedLines = ko.observable();
   this.authorGravatar = ko.computed(function() {
-    var lowerCaseEmail = self.authorEmail();
-    lowerCaseEmail = lowerCaseEmail ? lowerCaseEmail.trim().toLowerCase() : lowerCaseEmail;
-    return md5(lowerCaseEmail);
+    return md5((self.authorEmail() || "").trim().toLowerCase);
```
campersau: `.toLowerCase()`

Gruntfile.js
```diff
@@ -382,6 +382,59 @@ module.exports = function(grunt) {
...
+  grunt.registerTask('clickParallel', 'Parallelized click tests.', function() {
+    var done = this.async();
+
+    fs.readdirAsync('./nmclicktests')
+      .then(function(files) {
+        return files.filter(function(file) {
+          return file.startsWith("spec.");
+        });
+      }).then(function(tests) {
+        grunt.log.writeln('Running click tests in parallel... (this will take a while...) \t');
```
campersau: I think `\t` is not necessary here?

components/commit/commit.js
codingtwinky: somebody has been working on scala too much... >.<

codingtwinky:
Thanks for the review!!  you are a life saver!  

Also, I don't know if you do read some of my other PRs, could you please comment even if you don't catch anything?  Although I'm sure all of my PRs are littered with mistakes.  But if you could confirm however informally, it would make me feel so much better.

campersau:
@codingtwinky I have reviewed all of your PRs. In the future I will comment even if I can't find something. <sub>Which does not necessarily mean that there are no bugs :smirk:</sub>

codingtwinky:
Oh for sure, I'd just be happy knowing that there is some I can share ~~blame~ joy of programming with.

nmclicktests/environment.js
```diff
@@ -218,7 +218,7 @@ class Environment {
     return this.getPort()
       .then(() => this.startServer())
       .then(() => this.ensureStarted())
-      .catch((err) => { throw new Error("Cannot confirm ungit start!!"); })
+      .catch((err) => { console.log(err); throw new Error("Cannot confirm ungit start!!", err); })
```
campersau: In JavaScript there is no `Error` constructor which takes the "inner" / "previous" error.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error
https://nodejs.org/api/errors.html#errors_new_error_message

Issue #838, opened 2016-12-21
Credentials prompt for remote repository not appearing on web interface


viharm:
- Ungit version 0.10.3
- Debian Jessie 8.6 kernel 3.16.36-1+deb8u2
- NodeJS 6.9.2-1nodesource1~jessie1
- npm 3.10.9
- Apache 2.4.10-10+deb8u7 (not running Ungit, but probably relevant)

I am experiencing a similar issue to #107 where prompts for remote repository credentials do not appear on the web interface but on the command line, so have to keep an eye on the command line to enter these details. I run Ungit on a server, so have to SSH in to enter these details - very difficult :-(

Please help.

CurtisHumphrey:
Same issue but on a Mac running it locally. Push entering the password on the command line doesn't seem to work at all.

dimitrovs:
Same problem.

FredrikNoren:
Is this the same issue as #102 ? If so let's move the discussion into that one

dimitrovs:
I don't think it's the same, here we are talking about the credentials for the Git repo not a password for a private key. Ultimately the implementation may be the same though. For instance, check Eclipse Orion, it's a web IDE which comes with Git integration, there the UI handles the Git credentials prompt by throwing up a modal and the credentials are not stored.

molda:
I'm having similar issue and i'm unable to type in my credentials
When i try to go to a repo folder in browser and check terminal i see two prompts for username 
`Username for 'https://github.com': Username for 'https://github.com':` typing in `molda`
then i get prompt for password
`Password for 'https://molda@github.com':` 
but when i type in my password i get another prompt 
`Password for 'https://<my-password-here>@github.com':` 
I never see any prompt on the web interface.

wonson:
Same issue, so i close mine

wonson:
one temporary means to save your day is to set your remote path like this:
http://username:password@remotepath.com/repo.git

though i think this may be a bit insecure.

y3nly:
I have the same issue as well. Ungit 1.1.0 works fine.

viharm:
@y3nly ,
Does 1.1.0 prompt you for remote repo credentials on the web interface?

@FredrikNoren ,
This appears to be a popular issue, and a key requirement for working with remote repositories. Is this on the roadmap?

codingtwinky:
as this being open source I've been simply worked on "what is most interesting right now" and I understand that is not the best prioritization.  

[...]

PR #950, opened 2017-08-12
Fix cred helper for remotes

codingtwinky:
fix: #838

Basically due to how the [config.js](https://github.com/FredrikNoren/ungit/blob/master/source/config.js) is doing initialization validation, spawned [credential-helpr](https://github.com/codingtwinky/ungit/blob/master/bin/credentials-helper) by git process to trigger credential collection died silently without erros being properly logged.  

Solution is to remove those dependencies.  (All deps were for logging and error usage tracking, but the value of those are questionable at best and considering this is a specialized process that is to be spawned by git, we really should keep it skinny as possible.)

FYI, because I forgot how cred helpers work and me simply assuming this bug was due to recent git promise upgrade, this was one of the most exhausting bug that I've worked on while solution being so simple.

@viharm please let me know if this doesn't work for you.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,12 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 Use the following format for additions: ` - VERSION: [feature/patch (if applicable)] Short description of change. Links to relevant issues/PRs.`
 
-- 1.1.28: **UNPUBLISHED** please append your changes to here instead of new version.
+- 1.1.28: please append your changes to here instead of new version.
   - Refactoring to remove static data-ta tags from tests
   - `grunt nmclicktest` -> `grunt clicktest`
   - Stabilize ungit open test of clicktest via using a tag that is guaranteed to be generated
   - Add click test bailout on tes failure
+  - Remove deps to fix config init bug for the `credentials-helper`. [#838](https://github.com/FredrikNoren/ungit/issues/838)
 - 1.1.27: Add alert when moving back in time. [#914](https://github.com/FredrikNoren/ungit/issues/914)
 - 1.1.26:
   - fix invalid path input for autocomplete causing front end crash [#942](https://github.com/FredrikNoren/ungit/issues/942)
--- a/bin/credentials-helper
+++ b/bin/credentials-helper
@@ -1,33 +1,10 @@
 #!/usr/bin/env node
-var config = require('../src/config');
 var winston = require('winston');
-var path = require('path');
-var async = require('async');
-
-winston.remove(winston.transports.Console);
-
-var BugTracker = require('../src/bugtracker');
-var bugtracker = new BugTracker('credentials-helper');
-var usageStatistics = require('../src/usage-statistics');
-
-process.on('uncaughtException', function(err) {
-  console.error(err.stack.toString());
-  async.parallel([
-    bugtracker.notify.bind(bugtracker, err, 'credentials-helper'),
-    usageStatistics.addEvent.bind(usageStatistics, 'credentials-helper-exception')
-  ], function() {
-    process.exit();
-  });
-});
-
-if (config.logDirectory)
-  winston.add(winston.transports.File, { filename: path.join(config.logDirectory, 'credentials-helper.log'), maxsize: 100*1024, maxFiles: 2 });
-
 var socketId = process.argv[2];
 var port = process.argv[3];
-winston.info('Credentials helper invoked; port ' + port + ', socketId ' + socketId);
-
 var http = require('http');
+winston.remove(winston.transports.Console);
+winston.info('Credentials helper invoked; port ' + port + ', socketId ' + socketId);
 
 if (process.argv[4] == 'get') {
   winston.info('Getting credentials');
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "ungit",
   "author": "Fredrik Norén <fredrik.jw.noren@gmail.com>",
   "description": "Git made easy",
-  "version": "1.1.27",
+  "version": "1.1.28",
   "ungitPluginApiVersion": "0.2.0",
   "scripts": {
     "start": "node ./bin/ungit",


viharm:
@codingtwinky ,

Thanks. What's the best way to try this? Will this be on the npm registry?

codingtwinky:
Since this is a completely separate process, easiest way would be following:
1. go to the directory where ungit is installed.  (if you did `npm install -g ungit` then it would be likely be `/usr/local/lib/node_modules`)
2. modify `.../bin/credentials-helper` with the changes here.  
3. restart ungit

But if you would like to me to push this and give new version for you to download I can do that.

viharm:
Hi @codingtwinky , thanks for the advice. I will try these changes manually. Please don't go through the effort of packaging before confirmation of successfull testing. Will provide feedback in a day or two.

y3nly:
This fix didn't work for me. It's still failing silently and nothing gets logged.

I'm currently using git 1.9.1.

codingtwinky:
@y3nly thanks for the report.  Could you possibly help me debug this issue by running below steps and capturing the errors?

// 1~3 is same as above, including just to be verbose
1. go to the directory where ungit is installed. (if you did npm install -g ungit then it would be likely be /usr/local/lib/node_modules)
2. modify .../bin/credentials-helper with the changes here.
3. restart ungit

4. navigate to your git directory in browser, which is something like `http://127.0.0.1:8080/#/repository?path=%2F${your troubled git path}`
5. open up a terimnal
6. run `git -c color.ui=false -c core.quotepath=false -c core.pager=cat -c credential.helper=${UNGIT_INSTALL_ABS_PATH/bin/credentials-helper} 1 ${UNGIT_PORT} fetch ${NAME_OF_PASSWORDED_REMOTE} --prune`  (this should literally same as git remote fetch command you should see in ungit log if `logGitCommands` is enabled via `ungit --logGitCommands=true`.  In fact you can capture this in ungit log and run it
7. This should trigger user name and password dialog for the remote in the opened browser.  If it fails, logs should be observable in terminal.

Thanks,

y3nly:
@codingtwinky Sorry, looks like I had made some changed to the code that broke the credentials helper. After a  fresh install of ungit and your fix applied, the dialog box properly appears in the browser.

codingtwinky:
@y3nly thanks!

@viharm please let me know if this doesn't work

viharm:
I modified the file `/usr/lib/node_modules/ungit/bin/credentials-helper` based on 8ce0218

Used *Ungit* v1.1.27

I am still not getting the prompt; but getting the error on the console
```
Wed, 16 Aug 2017 16:27:17 GMT express deprecated res.send(status): Use res.sendStatus(status) instead at ../usr/lib/node_modules/ungit/src/server.js:76:7
```
This is followed by a prompt onthe console

I will try to debug based on your instructions originally for @y3nly .

viharm:
Just to be sure, the file looks like this
```javascript
!/usr/bin/env node
var winston = require('winston');
var socketId = process.argv[2];
var port = process.argv[3];
var http = require('http');
winston.remove(winston.transports.Console);
winston.info('Credentials helper invoked; port ' + port + ', socketId ' + socketId);
                                                                                                                                                                                                                                                                                       
if (process.argv[4] == 'get') {
  winston.info('Getting credentials');
  http.get('http://localhost:' + port + '/api/credentials?socketId=' + socketId, function(res) {
    winston.info('Got credentials');
    res.on('data', function(body) {
      var data = JSON.parse(body);
      console.log('username=' + data.username);
      console.log('password=' + (data.password ? data.password : ''));
    });
  }).on('error', function(err) {
    winston.error('Error getting credentials, couldn\'t query server', err);
  });
} else {
  winston.info('Unhandled param: ' + process.argv[4]);
}
```

viharm:
I did the following

1. go to the directory where ungit is installed. (`/usr/lib/node_modules`)
1. modify `/usr/lib/node_modules/ungit/bin/credentials-helper` as below
```javascript
!/usr/bin/env node
var winston = require('winston');
var socketId = process.argv[2];
var port = process.argv[3];
var http = require('http');
winston.remove(winston.transports.Console);
winston.info('Credentials helper invoked; port ' + port + ', socketId ' + socketId);
                                                                                                                                                                                                                                                                                       
if (process.argv[4] == 'get') {
  winston.info('Getting credentials');
  http.get('http://localhost:' + port + '/api/credentials?socketId=' + socketId, function(res) {
    winston.info('Got credentials');
    res.on('data', function(body) {
      var data = JSON.parse(body);
      console.log('username=' + data.username);
      console.log('password=' + (data.password ? data.password : ''));
    });
  }).on('error', function(err) {
    winston.error('Error getting credentials, couldn\'t query server', err);
  });
} else {
  winston.info('Unhandled param: ' + process.argv[4]);
}
```
3. Kill and restart *ungit*
1. navigate to git directory in browser, which is something like http://127.0.0.1:8080/#/repository?path=%2F${your troubled git path}
1. open up a terminal and navigate to the same directory on the console
1. run the command `git -c color.ui=false -c core.quotepath=false -c core.pager=cat -c credential.helper=${UNGIT_INSTALL_ABS_PATH/bin/credentials-helper} 1 ${UNGIT_PORT} fetch ${NAME_OF_PASSWORDED_REMOTE} --prune`

The `git` command produces an error on the command line
```
git: '1' is not a git command. See 'git --help'.
                                                                                                                                                                                                                                                                                       
Did you mean one of these?
       am
       gc
       mv
       rm
```

I tried replacing the `1` with a `0`, as this is the command line from the *Ungit* log when `--logGitCommands=true`

Still the same console error but with a `0` instead of `1`.

y3nly:
@viharm try git -c color.ui=false -c core.quotepath=false -c core.pager=cat -c credential.helper="${UNGIT_INSTALL_ABS_PATH/bin/credentials-helper} 1 ${UNGIT_PORT}" fetch ${NAME_OF_PASSWORDED_REMOTE} --prune

viharm:
@y3nly ,
The quotes are a good spot, thanks.

@codingtwinky ,
I tried with the quotes as suggested by @y3nly . The command runs in the console, with no unusual console output or prompt in the web browser. Am I doing something wrong? My *Git* version is `2.1.4-2.1+deb8u3` (`2.1.4`)

codingtwinky:
hmmm...  Can we make sure one thing.  So no error message at terminal means `credential-helper` it self ran without error.  However, if no dialog popup at browser means that we might have mismatched socket Id.   or there is something that I'm completely missing.  

Although I could ask you to do more testing with more steps but considering this is fix at least addresses a subset of potential problems I'm going to push this.  

@viharm @y3nly could you kindly upgrade ungit, and do this test one more time and let me know if problem persists?

codingtwinky:
Please feel free to reopen #838 if needed be

y3nly:
@codingtwinky I was able to `sudo -H npm install -g ungit@1.1.27` and confirm the previous version breaks the dialog box, then install 1.1.28 to resolve the issue. Thank you!

wonson:
fixed confirmed. Thanks

Though when i goes into a submodule, it keeps prompt me about username/password in UI (username/password not saved for submodule?).
But because it pops in browser UI, so I still can use chrome's "save password" to overcome that :)

viharm:
I'm still having issues with v1.1.28

@y3nly , @wonson , may I know your environments please? OS, Git, etc.?

codingtwinky:
@wonson submodule is a tricky concept as there are some difficulties around handling git within a git.  I may try to fix this along with several "todos" around submodule but it may require significant over haul on how it works...

@viharm I'm sorry to hear that and could you possibly open another ticket with more specific details such as below along with other details you may or maynot have already provided??
* console logs in terminal, ungit, and browser when you do steps above with 1.1.28 and manual git `credential-helper` command
* are there any git plugins you are using?


Thanks,

viharm:
@codingtwinky ,

Thanks for looking further into this. With respect, I would normally prefer to utilise the same issue #838 , as this was originally created by me, but also understand that others were having similar issues. So will open a new issue.

viharm:
Created #951 as suggested by @codingtwinky

wonson:
Git 2.7.4
Ubuntu 16.04

Another problem caused by this credential issue for submodule is: "file change" always show that it wants to remove my submodule, and I have no way to discard that change ._.

I have to `git submodule update --init -r` to reset the submodule state

y3nly:
@viharm I'm on Ubuntu 14.04.5 LTS, git 1.9.1, node.js 8.1.4, npm 5.0.3

wonson:
Update: I went every submodule directory and run `git config credential.helper store` seems solved the problem. can ungit run that for us?

codingtwinky:
I will try that over the weekends, but could you possibly open a new ticket so we can track this issue separately? 

Also, I'm assuming that you mean applying `git config credential.helper store` in submodules of a git repo will fix an issue where login prompts frequently being displayed.  if you could clarify that in the ticket it would be great.  

as always, thanks for the help!

wonson:
#954

Issue #176, opened 2013-08-31
Error initializing a repository


diegoferigo:
On ArchLinux i get a Whooops when i initialize a repo. Then, pressing F5 brings me always to the error page, to do anything i need to clear the entire folder.

I followed these steps:

```
diego@eeepc ~>mkdir /tmp/testrepo
diego@eeepc ~>cd /tmp/testrepo/
diego@eeepc testrepo>ungit 
info: App version: 0.1.7
info: App version: 0.1.7
info: socket.io started
info: Listening on port 8484
Browse to http://localhost:8484/#/repository?path=%2Ftmp%2Ftestrepo
## Ungit started ##
info: GET /
info: GET /
info: GET /css/styles.css
info: GET /config.js
info: GET /version.js
info: GET /js/bugsense.js
info: GET /js/ungit.js
info: handshake authorized
info: GET /api/latestversion
info: GET /images/logo.png
info: GET /images/repositoryBackgroundArrowUp.png
info: GET /images/logoLarge.png
info: GET /api/status?path=%2Ftmp%2Ftestrepo

# Here i press the init button

info: POST /api/init
info: GET /api/status?path=%2Ftmp%2Ftestrepo
info: GET /api/status?path=%2Ftmp%2Ftestrepo
info: GET /images/remoteIcon.png
info: GET /images/branchIcon.png
info: GET /images/tagIcon.png
info: GET /api/status?path=%2Ftmp%2Ftestrepo
info: GET /api/log?path=%2Ftmp%2Ftestrepo&limit=25
info: Start watching /tmp/testrepo
info: GET /api/checkout?path=%2Ftmp%2Ftestrepo
info: GET /api/remotes?path=%2Ftmp%2Ftestrepo
info: POST /api/fetch
```

I'm on nodejs v0.10.17.
Btw, if I clone a github repo all seems fine. Let me know if you need others info.

PR #963, opened 2017-09-16
Force git lang to english for ungit

### LINKED ISSUE #959: Don't auto-apply stash when switching branches
I may get something wrong here, but it is not clear to me why ungit automatically applies the stash I created on one branch to another branch on checkout. At first, I thought I might have some customized setting triggering that behavior, but I could not find anything like that.

From my understanding of git (and what I use as workflow), the idea is that you can use ´git stash´ to save your uncommited changes on Branch A, checkout another branch B, do some work, get back to Branch A, apply the stash and find everything like it was the moment you left.

Maybe also relevant: If I don't have any stashes and change branches, ungit will throw an error, reporting that theres no stash available.

codingtwinky:
fixes #959

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 Use the following format for additions: ` - VERSION: [feature/patch (if applicable)] Short description of change. Links to relevant issues/PRs.`
 
+- 1.1.30: add `LANG=en_US.UTF-8` to force git lang to english for ungit [#959](https://github.com/FredrikNoren/ungit/issues/959)
 - 1.1.29: move `Gruntfile.js` to es6
 - 1.1.28: please append your changes to here instead of new version.
   - Refactoring to remove static data-ta tags from tests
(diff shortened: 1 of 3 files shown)


campersau:
Maybe you can try `LANG=en_US.UTF-8 && git`?
On windows the syntax will be: `set LANG=en_US.UTF-8 && git`
But I haven't tested it because I use english git.

codingtwinky:
On  second thought, this is actually more complicated than my original thought.  I think even in different linux versions language specification is handled differently and it is not always guaranteed. 

Not to mention I can't even test this in Windows machine.

I would need more thought around this but for now we would have to simply document this.

campersau:
Just for reference: ungit already sets the env variables here https://github.com/FredrikNoren/ungit/blob/master/bin/ungit#L20

See also:
https://github.com/FredrikNoren/ungit/pull/694
https://github.com/FredrikNoren/ungit/issues/676
https://github.com/FredrikNoren/ungit/issues/176

codingtwinky:
So I've saw that and read some doc on node child process but this means that that env variable is not working.  If it was applied as one would have thought we wouldn't have #959 issue to begin with.

It is possible to add `LC_ALL`, `LANG` or any other keys but for whatever reason I couldn't successfully change git lang in my mac book and thus I couldn't really validate the fixes.

Issue #961, opened 2017-09-12
Unable to install ungit in node version 8.4.0


TitaneBoy:
Hi guys..

I'm trying to install ungit with yarn (on Windows) with node version 8.4.0 and I got the following error:

`error forever-monitor@1.1.0: The engine "node" is incompatible with this module. Expected version "0.8.x"`
`error Found incompatible module`

With investigation, I found in the package.json of your module that you are using the version 1.1.0 of forever-monitor module (as the error described before). And within its (forever-monitor) package.json, it use only the version 0.8.x of node, and not higher..

From the version 1.2.1 of forever-monitor module, It uses a version of node higher than 0.8.x.

Is there a reason that you use the version 1.1.0 of this module dependencie ?

Thank you for your help

codingtwinky:
When node 8 came out, packages NPM has some very nasty bugs that prevented installation of some of the c lib dependent lower level libraries, such as phantom, electron and etc.  

Will try to bump the node version and dependencies and give another go at it later.

codingtwinky:
Correction! for this ticket it has nothing to do with npm 5 issue we saw.  Reason why `forever-mointor`  was due to and old bug #531.  Thus we've locked `forever-monitor@1.1.0`.  

This actually has been causing lots of warnings and concerns but it didn't break so we largely left it alone as I don't have a window machine to verify #531 is indeed fixed for us.

I'm going to assume that it is fixed and will do a push.  Can you please verify it is functional in windows?

TitaneBoy:
@codingtwinky It will be a pleasure to test your fixes in Windows but I guess that your fixes is not yet released in npm because the last release of ungit is about 3 weeks.  I will wait until it will be available on npm..

Thanks :-)

campersau:
This is now fixed in ungit version 1.1.33 (see #971).

PR #964, opened 2017-09-17
Bump dependencies including forever-monitor

codingtwinky:
* bump dep
* add sleep in nmclick test as op is going through too fast in nightmare
* remove forever monitor version lock down

cc: #961 @TitaneBoy


campersau:
Sadly the cmd windows are poping up again on windows...

campersau:
See https://github.com/foreverjs/forever-monitor/issues/147 if I comment out this line it works...

codingtwinky:
I'm beginning to think that we should just drop forever monitor...  Have process die and let it be spawned if needed be considering ungit isn't mission critical process that needs high availability

campersau:
We can update forever-monitor up to 1.2.4 which is the last version without the cmd windows bug. (1.3.0 and higher are affected.)
As @TitaneBoy already mentioned it will then work with [node >= 0.8.x](https://github.com/foreverjs/forever-monitor/blob/v1.2.4/package.json#L40)
I have successfully tested it on windows.

TitaneBoy:
@campersau Great... Thank you for your help and support... One last question if possible: Does the fix is yet published in npm ? Because as I can see, the last release of ungit is since 3 weeks ..

Thanks :-)

campersau:
@codingtwinky also update `package-lock.json`.

@TitaneBoy No, it is not published. We will close your issue #961 once it is fixed.

campersau:
**Sorry** I have found another bug with forever-monitor 1.2.x which means we are stuck with 1.1.0...

After the 1.1.0 release they tried to be very clever by [allowing "complex" commands](https://github.com/foreverjs/forever-monitor/commit/d4a35488a679e146d0d22c577f7bf1b2dfc42a35#diff-8281e414f02aa1e5596b72f556bd4221R224). As you can see they are splitting the command by spaces to separate the executable from the args. But this also means that if the path to the executable (in our case nodejs) contains spaces the path will be broken.
E.g.: `C:\Program Files\nodejs\node.exe` becomes `C:\Program` `Files\nodejs\node.exe`

So running `bin/ungit` will not work because of the invalid executable path.

Maybe @codingtwinky is right by just dropping it. <sub>Or we can fork it...</sub>

---

forever-monitor versions:
- 1.1.0: works but is [restricted to node 0.8.x](https://github.com/foreverjs/forever-monitor/blob/ff6ff22cfefb56607a506e2b2c924a5cbdf4b41b/package.json#L40)
- 1.2.x: is not restricted to node 0.8.x but will not work with nodejs paths which include spaces
- 1.3.x and higher: does not really work on windows because [cmd windows are detached](https://github.com/foreverjs/forever-monitor/issues/147)

😢

codingtwinky:
@campersau I'm going to downgrade it to 1.1.0 for now and I might send a different PR to remove forever monitor.

PR #971, opened 2017-10-10
Removing Forever dependency

codingtwinky:
I will try to refactor it to es6 without async.


bin/ungit
```diff
@@ -11,19 +10,10 @@ var BugTracker = require('../src/bugtracker');
...
-  env: { LANG: 'en_US.UTF-8' }
-});
-
-process.on('exit', function() {
-  child.kill();
-  console.log('Exit: killed children')
-});
+var child = child_process.spawn('node',
+  [path.join(__dirname, '..', 'src', 'server.js')].concat(process.argv.slice(2)),
+  { cwd: path.join(process.cwd(), '..') }
```
campersau: What about the `env` setting?

bin/ungit
```diff
@@ -11,19 +10,10 @@ var BugTracker = require('../src/bugtracker');
 var bugtracker = new BugTracker('launcher');
 var usageStatistics = require('../src/usage-statistics');
 
-var child = new (forever.Monitor)(path.join(__dirname, '..', 'src', 'server.js'), {
-  silent: false,
-  minUptime: 2000,
-  max: config.maxNAutoRestartOnCrash,
```
campersau: Is `maxNAutoRestartOnCrash` used anywhere else?

bin/ungit
codingtwinky: It didn't seems to be respected as apparent by #959.  I may leave it in with a note though...

bin/ungit
codingtwinky: Not really...  From my tests forever monitor actually never kicked in and process didn't die during errors.  Errors seems to be captured and handled so..

campersau: Ok, so that means the config setting can be removed.

CHANGELOG.md
```diff
@@ -3,7 +3,9 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 Use the following format for additions: ` - VERSION: [feature/patch (if applicable)] Short description of change. Links to relevant issues/PRs.`
 
-- 1.1.31: Bump dependencies
+- 1.1.31:
```
campersau: Also bump version in `package.json`.

PR #980, opened 2017-10-23
Resolve given path at the server side.

### LINKED ISSUE #975: Relative path support
Hi there. Currently ungit getting some errors when i'm using path starting with ~/. Would be great if you can fix it, because writing absolute path is pain in the ass

codingtwinky:
fix #975


public/source/main.js
```diff
@@ -53,6 +53,7 @@ ko.bindingHandlers.autocomplete = {
       var lastChar = value.slice(-1);
       if (lastChar == '/' || lastChar == '\\') {  // When "/" or "\"
         server.getPromise('/fs/listDirectories', {term: value}).then(function(directoryList) {
+          $(element).val(directoryList.shift() + ungit.config.fileSeparator);
```
campersau: I think the `ungit.config.fileSeparator` should only be added if it isn't already the last character.
Because if I enter on windows `C:\` it autocompletes to `C:\\`.

source/server.js
```diff
@@ -332,18 +332,18 @@ app.get('/api/fs/exists', ensureAuthenticated, (req, res) => {
 });
 
 app.get('/api/fs/listDirectories', ensureAuthenticated, (req, res) => {
-  const dir = req.query.term.trim();
+  const dir = path.resolve(req.query.term.trim()).replace("/~", "");
```
campersau: Wouldn't it be better to check if it starts with `/~` and replace it with [`os.homedir()`](https://nodejs.org/api/os.html#os_os_homedir).
Then this would also work on windows.

Example:
```js
var dir = req.query.term.trim();
if (dir.startsWith("/~")) {
    dir = path.join(os.homedir(), dir.substring(2));
}
```

codingtwinky: Well, for linux it should work with "~".  My thought behind using `path.resolve` is that it has more considerations around path resolving then I can think of and thus more trust worthier

public/source/main.js
codingtwinky: will do

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ Use the following format for additions: ` - VERSION: [feature/patch (if applicab
 
 - 1.1.34: **NOT RELEASED YET**
   - fix potential memory leak with `express-session`[#977](https://github.com/FredrikNoren/ungit/issues/977)
+  - Fix document title on windows [#983](https://github.com/FredrikNoren/ungit/pull/983)
+  - parse local storage as json instead of regex[#981](https://github.com/FredrikNoren/ungit/pull/981)
```
campersau: Thank you!

Nit: can you add a space between the text and the link?

PR #981, opened 2017-10-23
autocomplete: parse repositories from local storage as json

campersau:
also fix labels on windows

The repositories were already parsed as json in [app.js](https://github.com/FredrikNoren/ungit/blob/master/components/app/app.js#L19)

### DIFF
--- a/public/source/main.js
+++ b/public/source/main.js
@@ -72,12 +72,11 @@ ko.bindingHandlers.autocomplete = {
         event.preventDefault();
         navigation.browseTo('repository?path=' + encodeURIComponent(value));
       } else if (localStorage.repositories && value.indexOf("/") === -1 && value.indexOf("\\") === -1) {
-        var folderNames = localStorage.repositories.replace(/("|\[|\])/g, "")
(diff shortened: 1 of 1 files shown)


public/source/main.js
```diff
@@ -72,12 +72,16 @@ ko.bindingHandlers.autocomplete = {
         event.preventDefault();
         navigation.browseTo('repository?path=' + encodeURIComponent(value));
       } else if (localStorage.repositories && value.indexOf("/") === -1 && value.indexOf("\\") === -1) {
-        var folderNames = localStorage.repositories.replace(/("|\[|\])/g, "")
-          .split(",")
+        var folderNames = JSON.parse(localStorage.repositories)
           .map(function(value) {
+            var index = value.lastIndexOf("/");
+            if (index == -1) {
+              index = value.lastIndexOf("\\");
```
codingtwinky: You could use `ungit.config.fileSeparator` instead of these if statements

campersau: Fixed.

Issue #973, opened 2017-10-10
UI support for tons of tags on the same commit


bakkerthehacker:
Some of the repos I use ungit in have automated systems that will attach a lot of tags to commits.  Some of these commits can end up with 10-100 tags on them, each 50 to 100 characters long.  

The ungit UI puts the commit buttons (merge, checkout, move, rebase) on the far right side of all of these tags.  Scrolling over very very far, or zooming out, is a huge hassle when dealing with these commits and makes ungit essentially unusable.  

Some solutions may be to:
- have a customizable filter for tags
- limit the number of tags on a commit
- move the commit-buttons to some other location that is not dependant on the tags

Thatkookooguy:
@bakkerthehacker is there an example of a project that has this problem? I might look into this a little and a test case would help greatly

campersau:
@Thatkookooguy for your information @codingtwinky is already working on this see #984
Or is this related to https://github.com/FredrikNoren/ungit/issues/974#issuecomment-341961804 ?

PR #984, opened 2017-10-31
When too many refs, allow users to search and cycle through branches and tags

codingtwinky:
fix: #973

* Will always display ref search as long as each nodes has at least one ref available
* Selecting hidden ref will bring it to the left most order and pop off the last displayed ref
* Selecting already displayed ref will bring existing ref to the left most order
* Number of refs to display is configurable by `numRefsToShow`
  * Distribution between branches and tags is not straight forward.  2/5 of the `numRefsToShow` will be used to show tags and `numRefsToShow` - ${length of the tags} will be used to show branches.  This means if `numRefsToShow` is 5 and a git node has 0 tags, ungit will display 5 branches no matter how many branches exists.  If `numRefsToShow` is 5 and git node has 1 tag, ungit will display 4 branches no matter how many branches exists.


components/graph/graph.html
```diff
@@ -43,7 +43,7 @@
         <!-- /ko -->
 
         <!-- ko if: showNewRefAction -->
-        <span class="newRef" data-bind="css: { editing: branchingFormVisible }">
+        <span class="newRef" data-bind="click: function() {console.log(3333333)}, css: { editing: branchingFormVisible }">
```
campersau: Remove `console.log`

components/graph/graph.less
```diff
@@ -167,6 +167,20 @@
         padding: 0px;
         margin: 0px;
         margin-top: 9px;
+        -webkit-text-stroke-width: 1px;
+        color: rgba(255, 255, 255, 0.3);
+        &:hover {
+          color: rgba(255, 255, 255, 0.8);
+        }
+      }
+      .showSearchForm{
```
campersau: Nit: add space

components/graph/git-node.js
```diff
@@ -62,13 +89,21 @@ var GitNodeViewModel = function(graph, sha1) {
...
   });
   this.branchOrder = ko.observable();
   this.aboveNode = undefined;
   this.belowNode = undefined;
+  this.refSearchFormVisible = ko.observable(false);
+  this.refSearchFormVisible.subscribe((newValue) => {
+    if (!newValue) {
+      // Small timeout because in ff the form is hidden before the submit click event is registered otherwise
+      setTimeout(() => {
+        this.refSearchFormVisible(false);
```
campersau: I tested it in Firefox too. But for me it worked also without this code.
Also I don't quite understand why setting `refSearchFormVisible` to `false` after it is already set to `false` should do anything. Doesn't knockout check if the value is the same as before and does nothing?

components/graph/git-node.js
```diff
@@ -24,12 +27,33 @@ var GitNodeViewModel = function(graph, sha1) {
...
+    if (newValue) {
+      this.branches(newValue.filter((r) => r.isBranch));
+      this.tags(newValue.filter((r) => r.isTag));
+      this.tagsToDisplay(this.tags.slice(0, maxTagsToDisplay));
+      this.branchesToDisplay(this.branches.slice(0, ungit.config.numRefsToShow - this.tagsToDisplay().length));
+    } else {
+      this.branches.removeAll();
+      this.tags.removeAll();
+      this.branchesToDisplay.removeAll();
+      this.tags.removeAll();
```
campersau: Too late but I think this should be `this.tagsToDisplay.removeAll();`.

codingtwinky: thanks for the catch, will submit one soon if no one has

PR #1012, opened 2018-01-17
Add remote branch listing feature

### LINKED ISSUE #966: [Feature Request] List of remote branches?
I hope ungit can show a list of remote branches(&tags). If i hvnt checkout a remote branch, and the branch has sunk for a period of time. It seems that I need to scroll and scroll down to find that remote branch from the tree now :(

jung-kim:
fixes #966

* ability to list all branches including remote branches within branch list bar.
* lots of refactoring to recycle and share git-ref objects and checkout logic for simplified code path
* ability to checkout remote branches[1]

[1] Current implementation of remote branch logic is something that I'm not sure if it is a good idea always.
Currently, when remote branch checkout is performed, ungit will FORCE move local branch of the remote branch to the remote branch's location and checkout that local branch.  

This is not always ideal and needs to be discussed more and is demonstrated below.

screen shot 2018-01-15 at 13 46 02

* you can see that remote branch `remot/imgtest` and local `imgtest` branch is in different locations.
* when `remot/imgtest` is checked out, local `imgtest` is force moved to `remot/imgtest`.

Considering this was an existing behavior, perhaps users are okay with this but I may fix this so to create tmp branches as briefly discussed in #966.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "ungit",
   "author": "Fredrik Norén <fredrik.jw.noren@gmail.com>",
   "description": "Git made easy",
-  "version": "1.4.6",
+  "version": "1.4.7",
   "ungitPluginApiVersion": "0.2.0",
   "scripts": {
     "start": "node ./bin/ungit",
--- a/public/less/styles.less
+++ b/public/less/styles.less
@@ -85,6 +85,19 @@
   max-height: 500px;
   height: auto;
   overflow-y: scroll;
+  min-width: 205px;
+  widht: auto;
+  .option {
+    padding-left: 20px;
+    font-size: 120%;
+    color: black;
+  }
+  .dropdown-divider {
+    height: 0;
+    margin: .5rem 0;
+    overflow: hidden;
+    border-top: 1px solid #e9ecef;
+  }
   li {
     a {
       &.list-url {
--- a/public/less/styles.less
+++ b/public/less/styles.less
@@ -100,6 +100,7 @@
   }
   li {
     a {
+      padding-right: 30px;
       &.list-url {
         padding: 0px 4px 0px 4px;
         margin-right: 30px;
--- a/public/less/styles.less
+++ b/public/less/styles.less
@@ -82,6 +82,9 @@
 }
 
 .dropdown-menu {
+  max-height: 500px;
+  height: auto;
+  overflow-y: scroll;
   li {
     a {
       &.list-url {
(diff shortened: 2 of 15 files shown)


public/less/styles.less
```diff
@@ -82,6 +82,22 @@
 }
 
 .dropdown-menu {
+  max-height: 500px;
+  height: auto;
+  overflow-y: scroll;
+  min-width: 205px;
+  widht: auto;
```
campersau: `width`

source/git-api.js
```diff
@@ -401,11 +403,30 @@ exports.registerApi = (env) => {
       .finally(emitGitDirectoryChanged.bind(null, req.query.path));
   });
 
+  const createBranchIfPossible = (branchName, path, retry) => {
+    retry = retry || 3;
```
campersau: If `retry` is `0` then `retry` would be set to `3` again.

source/git-api.js
```diff
@@ -401,11 +403,30 @@ exports.registerApi = (env) => {
...
+    return gitPromise(['branch', branchName], path).catch((e) => {
+        if (retry > -1) {
+          return createBranchIfPossible(`ungit-${crypto.randomBytes(4).toString('hex')}`, path, retry - 1);
```
campersau: Where is `crypto` defined?

source/git-api.js
```diff
@@ -401,11 +404,29 @@ exports.registerApi = (env) => {
       .finally(emitGitDirectoryChanged.bind(null, req.query.path));
   });
 
+  const createBranchIfPossible = (branchName, path, retry) => {
+    return gitPromise(['branch', branchName], path).catch((e) => {
+        if (retry > -1) {
```
campersau: Just to be sure this is intended: this will retry `4` times. 3, 2, 1, 0

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "ungit",
   "author": "Fredrik Norén <fredrik.jw.noren@gmail.com>",
   "description": "Git made easy",
-  "version": "1.4.6",
+  "version": "1.4.7",
```
campersau: also update `package-lock.json`

jung-kim: I don't think they changes `package-lock.json` for version up for the actual app.  I have no changes for it.

source/git-api.js
jung-kim: On grand scheme of things... Meh... but yeah will fix it for the consistency sake.

package.json
campersau: Hmm if I run `npm install` it shows up as a change. But it might not be important.

image

jung-kim: you are absolutely right and it is now fixed.  will do `npm i` onwards even if no dep changes happens.

Issue #1016, opened 2018-01-22
Differentiate local/remote Tags?


LukeDWood:
Could there be a way to differentiate between a local and remote tag? 
Currently it's not possible to tell whether a tag has been pushed to the repo or is just local.

PR #1024, opened 2018-01-27
Add fixes to prevent remote tag removal

jung-kim:
fixes: #1016

* Added api end point `/refs` - it will return local branch, remote branch and local tags.
* Continue to rely on existing `components/remotes/remotes.js#fetch()` for refreshing remote tags.
* on proper git events, trigger refresh using __BOTH__ functions above.
* no longer rely on parsing git log result for updating any refs.

----
_(outdated comments)_

This one is an odd on.  It's not that we are not differentiating between remote and local tags.  It's that it was being removed later by the process that parses `/log` and detects any remote branch or tag changes and removes them. 

For whatever reason, `/log` is not displaying remote tags any more and process thinks that it was deleted by another source and deletes from ungit UI.

I'm currently trying to figure out why `/log` is not showing remote tags but current fix means that remote tag deletion or changes will not be picked up in the UI without refresh.

screen shot 2018-01-27 at 13 30 36

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@ Use the following format for additions: ` - VERSION: [feature/patch (if applicab
   - fix remote tag push not creating remote tag
   - change ref refresh logic
   - show error on incorrect credentials [#1042](https://github.com/FredrikNoren/ungit/pull/1042)
-  - allow credential handling for remotes [#1041](https://github.com/FredrikNoren/ungit/pull/1041)
-  - add cancel button for empty commits and amends [#1036](https://github.com/FredrikNoren/ungit/pull/1036)
+  - allow credential handling for remotes [#1039](https://github.com/FredrikNoren/ungit/issues/1039)
+  - add cancel button for empty commits and amends [#1029](https://github.com/FredrikNoren/ungit/issues/1029)
 - 1.4.7: add remote branches to the branch list. [#966](https://github.com/FredrikNoren/ungit/issues/966)
 - 1.4.6:
   - dependency bump to fix dependency's security problem.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 Use the following format for additions: ` - VERSION: [feature/patch (if applicable)] Short description of change. Links to relevant issues/PRs.`
 
-- 1.4.8: differentiate remote vs local tag. [#1016](https://github.com/FredrikNoren/ungit/issues/1016)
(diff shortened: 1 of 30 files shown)


campersau:
This was added in https://github.com/jung-kim/ungit/commit/5d66c2a2c4713c455edd75c968e65da30bba5e10 to detect external changes https://github.com/FredrikNoren/ungit/pull/809

Will this still work with this change?

jung-kim:
So without this PR, external changes for remote tag never worked as it was not gathered from `/log` api call.

This PR is explicitly ignoring that logic to skip remote tag so remote tags wouldn't be delete by external change logic.

----

Right now we are relying on `/log` to update nodes, remote branch, and remote tags. we would have to separate out to different api calls but I'm conducting more tests

CHANGELOG.md
```diff
@@ -3,11 +3,14 @@ All notable changes to this project will be documented in this file.
...
-  - empty commit [#1028](https://github.com/FredrikNoren/ungit/issues/1028)
-  - fix commit detail layout while hovering over commit node [#1025](https://github.com/FredrikNoren/ungit/issues/1025)
-- 1.4.8: fix remote branches display name and delete action [#1032](https://github.com/FredrikNoren/ungit/issues/1032), [#1031](https://github.com/FredrikNoren/ungit/issues/1031)
+- 1.4.10:
+  - differentiate remote vs local tag. [#1016](https://github.com/FredrikNoren/ungit/issues/1016)
+  - fix push not throwing giterror
+  - fix remote tag push not creating remote tag
+  - change ref refresh logic
+  - show error on incorrect credentials [#1042](https://github.com/FredrikNoren/ungit/pull/1042)
+  - allow credential handling for remotes [#1041](https://github.com/FredrikNoren/ungit/pull/1041)
```
campersau: Nit: We always link to the issues instead of the PR if one exists.

#1042 > https://github.com/FredrikNoren/ungit/issues/1039
#1036 > https://github.com/FredrikNoren/ungit/issues/1029

jung-kim:
@campersau @ylecuyer 

I feel that I should bring [this](https://github.com/FredrikNoren/ungit/pull/1024/files#diff-7cc6d113be128a4ba2e032f8db376ebaR342) to you guys' attention and would love to hear your thoughts.


One of the point of this PR is to have better remote ref fetching and have repos be refreshed real time as repo is altered.  However, to do this correctly, we would have to fetch credentials for every single remotes everytime this new api route is called.  

Which may get annoying.

Right now, I'm completely ignoring credential refresh at this route and do blind fetch.  I solemnly believe that people should go to passwordless management solution via .ssh credentials or keychain management or whatever.  But I'm don't want force my belief and would love to hear you guy's thoughts on this.

ylecuyer:
I don't think forcing passwordless is a good thing. Can't you just pass remotes that require a password ? and maybe displaying a warning showing that some remotes were omited

PR #1026, opened 2018-01-27
fix commit detail layout while hovering over commit node

### LINKED ISSUE #1025: Bug in commit details layout when hovering over commit node
image

campersau:
Returning an empty object will remove the inline styles completly. No need to set everything back to 0 or inherit.
Not sure when this issue occured the first time.

fixes #1025


jung-kim:
can we add change logs as well?

campersau:
@jung-kim yes, you can piggi back it later. less conflicts 😄

PR #1033, opened 2018-01-30
Remove heads from display name

### LINKED ISSUE #1032: Remove heads from branch dropdown
branch-delete

As the dropdown only display local and remote branches (ie no tags) it is not needed to display heads/ before local branch names.

ylecuyer:
Fixes #1032 

new-branch-dpdown


campersau:
I have written the exact same fix for this 😄 

<sub>(It would be great to also use the new `displayName` in the `graph` component)</sub>

PR #1035, opened 2018-01-31
Cleanup clicktest output

ylecuyer:
The clicktest output is too bloated. This is an attempt to clean it


jung-kim:
I think this is fine but I was hoping to use more proper login tools such as `winston` we are already using for this one.

ylecuyer:
I updated the PR to use winston

PR #1036, opened 2018-01-31
Cancel button

### LINKED ISSUE #1029: Add cancel button next to commit (amend, empty-commit)
Add a cancel button next to the commit button in the staging view.
Currently, when you're amending a commit or committing an empty commit you have to refresh to cancel.

ylecuyer:
Fixes #1029

### DIFF
--- a/components/staging/staging.js
+++ b/components/staging/staging.js
@@ -81,6 +81,9 @@ var StagingViewModel = function(server, repoPath) {
   this.showNux = ko.computed(function() {
     return self.files().length == 0 && !self.amend() && !self.inRebase() && !self.emptyCommit();
   });
+  this.showCancelButton = ko.computed(function() {
+    return self.amend() || self.emptyCommit();
+  });
   this.commitValidationError = ko.computed(function() {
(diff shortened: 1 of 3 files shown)


jung-kim:
Can we uptick npm packge version and add to change log so we can keep track of this?

ylecuyer:
@jung-kim can we bundle this PR and #1041 in one release ? How do we do that ?

jung-kim:
some have difference of opinion on how it should be managed, some are more strict on saying that a commit should be released with change log and uped package version.  

But don't feel so strict about it and feel free to make comments per PR and I can bundle the releases together for you on subsequent releases.

And yes I will do it for this one

components/staging/staging.js
```diff
@@ -263,6 +266,10 @@ StagingViewModel.prototype.invalidateFilesDiffs = function() {
     file.diff().invalidateDiff();
   });
 }
+StagingViewModel.prototype.cancelAmendEmpty = function() {
```
jung-kim: Instead of `cancelAmendEmpty()` is a literally a proxy function for `resetMessages()`.  

Should we just call `resetMessages()` directly?

Issue #1039, opened 2018-02-04
Diferentiate remotes authentication


ylecuyer:
If two remotes require authentication, the pop asking for credentials doesn't say for which remote.

campersau:
fixed by #1041

PR #1040, opened 2018-02-04
display the remote which requires authentification

campersau:
fixes https://github.com/FredrikNoren/ungit/issues/1039

image


campersau:
closing in favor of #1041

PR #1042, opened 2018-02-05
Show error if incorrect login/password

ylecuyer:


jung-kim:
Thanks for catching this one.  I know I have found one and I'm sure there are several places where I have missed it.

Issue #1038, opened 2018-02-02
Add a commit & push button


ylecuyer:
I often find myself doing the combo commit & push.

I think it may be useful to have a commit & push button next to the existing commit button

campersau:
Just to clarify: this should push to the currently active remote right?

Visual Studio uses this design:
![7838 image_042e259b](https://user-images.githubusercontent.com/4009570/35727789-92d6a49c-0809-11e8-95c3-58029c810691.png)

ylecuyer:
Right. What does commit and sync does ?

campersau:
> Commit and Sync does three things. First, it will commit. Second, it will perform a pull (grabs the updated information from the remote repo). Finally, it will push.

https://stackoverflow.com/a/30039179

jung-kim:
Should one implement this, can we make this feature a simple flag that alters behavior of the commit button rather than have it as new button?  I think that would be better

ylecuyer:
I think it would be better to have a drop down button as in campersau
screenshot.

jung-kim:
So when `commit and push` is clicked does `commit` button switch to `commit and push` button and state is managed that way?  or when you want o do `commit and push` do you have to do two clicks all the time?

Honestly as long state is sticky I'm for it

campersau:
In Visual studio it is not a sticky button but I am not sure if it should be. @jung-kim what is your definition of sticky? Sticky per session? Per repository? Global?
Two clicks are at least one click less than now 😄 

Also one interesting thing is that if there is no remote Visual Studio still shows the commit and push option and displays a screen where you can add a remote after the commit.

jung-kim:
I was thinking per browser should be sufficient.

oh thats nice, we can do that as well.

campersau:
fixed with #1046

PR #1046, opened 2018-02-11
Commit & push

ylecuyer:
See #1038

### DIFF
--- a/components/staging/staging.html
+++ b/components/staging/staging.html
@@ -20,9 +20,17 @@
           </span>
           <span class="commit-message-title-counter" data-bind="text: commitMessageTitleCount"/>
         </div>
-        <button class="btn btn-primary btn-lg commit-btn" data-bind="click: commit, visible: isStageValid, enable: !commitValidationError()">
-          Commit
-        </button>
+        <div class="btn-group commit-grp" data-bind="visible: isStageValid">
+          <button class="btn btn-primary btn-lg commit-btn" data-bind="click: commit, enable: !commitValidationError()">
+            Commit
+          </button>
+          <button type="button" class="btn btn-primary btn-lg dropdown-toggle" data-toggle="dropdown" data-bind="enable: !commitValidationError()">
+            <span class="caret"></span>
+          </button>
+          <ul class="dropdown-menu" role="menu">
+            <li><a href="#" data-bind="click: commitnpush, css: { disabled: commitValidationError() }" class="commitnpush">Commit & Push</a></li>
+          </ul>
+        </div>
         <button class="btn btn-warning btn-lg btn-stg-cancel" data-bind="click: cancelAmendEmpty, visible: showCancelButton">
           Cancel
         </button>
--- a/nmclicktests/environment.js
+++ b/nmclicktests/environment.js
@@ -25,6 +25,13 @@ Nightmare.action('ug', {
       .wait(1000)
       .then(done.bind(null, null), done);
   },
+  'commitnpush': function(commitMessage, done) {
+    this.wait('.files .file .btn-default')
+      .insert('.staging input.form-control', commitMessage)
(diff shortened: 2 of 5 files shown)


ylecuyer:
Currently missing force push if behind

components/staging/staging.html
```diff
@@ -20,9 +20,17 @@
           </span>
           <span class="commit-message-title-counter" data-bind="text: commitMessageTitleCount"/>
         </div>
-        <button class="btn btn-primary btn-lg commit-btn" data-bind="click: commit, visible: isStageValid, enable: !commitValidationError()">
-          Commit
-        </button>
+        <div class="btn-group commit-grp">
+          <button class="btn btn-primary btn-lg commit-btn" data-bind="click: commit, visible: isStageValid, enable: !commitValidationError()">
```
campersau: I think the `visible: isStageValid` should be on the `btn-group` because if there is a conflict the arrow of the dropdown is still visible and clickable.

image

ylecuyer: nice catch will fix it

nmclicktests/environment.js
```diff
@@ -25,6 +25,16 @@ Nightmare.action('ug', {
       .wait(1000)
       .then(done.bind(null, null), done);
   },
+  'commitnpush': function(commitMessage, done) {
+    this.wait('.files .file .btn-default')
+      .insert('.staging input.form-control', commitMessage)
+      .wait(100)
+      .click('.commit-grp .dropdown-toggle')
+      .wait(100)
```
campersau: You can use `ug.click` instead. It already waits.

https://github.com/FredrikNoren/ungit/blob/39a93198f0e53eaf97025ae8d3bb11cf79e161d3/nmclicktests/environment.js#L146-L152

ylecuyer:
GTM if ok for you @campersau

components/staging/staging.html
```diff
@@ -20,9 +20,17 @@
...
           <span class="commit-message-title-counter" data-bind="text: commitMessageTitleCount"/>
         </div>
-        <button class="btn btn-primary btn-lg commit-btn" data-bind="click: commit, visible: isStageValid, enable: !commitValidationError()">
-          Commit
-        </button>
+        <div class="btn-group commit-grp" data-bind="visible: isStageValid">
+          <button class="btn btn-primary btn-lg commit-btn" data-bind="click: commit, enable: !commitValidationError()">
+            Commit
+          </button>
+          <button type="button" class="btn btn-primary btn-lg dropdown-toggle" data-toggle="dropdown">
```
campersau: Hmm.. not sure if this button should also be disabled. It looks strange to have the commit button disabled but the dropdown is enabled which then shows only disabled entries.

image

See Visual Studio for comparison
image

ylecuyer: OK. I updated the PR to disable the dropdown toggle too and squashed the commits to have only one.

ylecuyer:
Is the PR OK for you @campersau?

campersau:
I just remembered this:
> [Also one interesting thing is that if there is no remote Visual Studio still shows the commit and push option and displays a screen where you can add a remote after the commit.](https://github.com/FredrikNoren/ungit/issues/1038#issuecomment-362904172)

I am not sure if we want to do that or just disable the `Commit & Push` button. Currently this exception is shown:
image

It would also be great if you could update the [`CHANGELOG.md` and bump the version in `package.json`](https://github.com/FredrikNoren/ungit/blob/master/CONTRIBUTING.md#pull-requests) and [`package.json.lock`](https://github.com/FredrikNoren/ungit/pull/1012#discussion_r162830763).

Sorry I should have done a complete review from the beginning 😞

ylecuyer:
@campersau this error also appears if you try to push from a repo without any remote.
I think we should merge this PR and open another ticket for the non existing remote problem

PR #1052, opened 2018-02-17
copy nprogress.css to public/css/

campersau:
I got the following network error before:
image

@jung-kim how did it work for your in https://github.com/FredrikNoren/ungit/pull/1003 ?


jung-kim:
Ermm... I'm not sure, I think I might have fixed it by making the manual fix and forgot to add copy statement.


Thanks for catching this!

PR #1056, opened 2018-02-24
Format

atefBB:
formating !

### DIFF
--- a/bin/ungit
+++ b/bin/ungit
@@ -58,11 +58,11 @@ const launch = () => {
   );
 
   child.on('exit', (res) => {
-    console.log('Stopped keeping ungit alive');
+    console.log('Stopped! keeping ungit alive');
   });
 
   const startupListener = (data) => {
-    if (data.toString().indexOf('## Ungit started ##') >= 0) {
+    if (data.toString().indexOf('Ungit started!') >= 0) {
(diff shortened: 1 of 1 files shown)


bin/ungit
```diff
@@ -58,11 +58,11 @@ const launch = () => {
   );
 
   child.on('exit', (res) => {
-    console.log('Stopped keeping ungit alive');
+    console.log('Stopped! keeping ungit alive');
   });
 
   const startupListener = (data) => {
-    if (data.toString().indexOf('## Ungit started ##') >= 0) {
```
jung-kim: Ungit's clicktest capture this console output to detect the service start and this is the reason why all tests are failing.

atefBB: @jung-kim how to change this behavior?

jung-kim: Before that, why do you want to change this behavior?

atefBB: the phrase `## Ungit started ##` with `##` seems ugly for me !

jung-kim: it seems that what is being suggested here is about the appearance of string, I'm not sure that is a strong enough of a reason to change test behaviors.

atefBB: As you want! I think it's not that difficult to change test behavior! 
Have a nice day.

PR #1066, opened 2018-03-21
Fix pathToNavigateTo not set when launchBrowser is false and launchCommand is set

### LINKED ISSUE #1065: %U is null when launchBrowser is false and launchCommand is set
I used to use the following `.ungitrc` to start ungit server and launch chromium without the browser toolbars.

```
{
  "debug": true,
  "launchCommand": "chromium-browser --app=%U",
  "launchBrowser": false
}
```

It was working in the previous versions of ungit (I can't recall the exact version I used before).

With the latest version, when I run `ungit` within a git repository, the output is as follows:

```
## Ungit started ##

Took 635ms to start server.
Running custom launch command: chromium-browser --app=null
```

And an empty chromium window (with toolbars) is launched.

If I set `launchBrowser` to `true`, `%U` is correctly set and chromium is launched with the correct configuration. But the default browser is also launched _-as expected-_, which makes it an unusable workaround.

I think adding `|| config.launchCommand` to this line might fix this issue, but I haven't tested it myself.

https://github.com/FredrikNoren/ungit/blob/b86452903db91c1828a1882abfe794f2fd96fa1b/bin/ungit#L32

I can test & verify it locally and send a PR if my approach is right.

_versions_
ungit: 1.4.13
node: v8.9.4
npm: 5.6.0
os: linux

cgty:
Fixes #1065

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 Use the following format for additions: ` - VERSION: [feature/patch (if applicable)] Short description of change. Links to relevant issues/PRs.`
 
+- 1.4.15:
+  - fix not setting `pathToNavigateTo` properly when `launchBrowser` is false and `launchCommand` is set [#1065](https://github.com/FredrikNoren/ungit/issues/1065)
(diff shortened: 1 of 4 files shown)


jung-kim:
Lovely, thanks!

PR #1068, opened 2018-03-22
Fix: text-area width inside dialog

ylecuyer:
Before

bette-gitignore-width-before

After

bette-gitignore-width-after

No version bump because it's minor


jung-kim:
Thanks for the PR!

However, can you make sure you only bump the patch version and not the minor version when you bump the ungit version number?  

Also, can you be sure to propagate the version number bump to `package.json` and `package-lock.json` via `npm -i`?

PR #1069, opened 2018-03-22
Mover version number below logo with 140px padding

wormeyman:
Just a quickie fix I find this more pleasing to my eyes.

image


jung-kim:
Thanks!

PR #1070, opened 2018-03-24
Misc. typos

luzpaz:
Found via `codespell -q 3`


jung-kim:
What is this sorcery that pinpoints all the embarrassing mistakes... 

Thanks! and will be using it next time.


(Will not be bumping the ungit version and cutting the new version as it's only comment changes)

PR #1073, opened 2018-03-28
Fix branch name conflict when auto checkout and ref filtering

jung-kim:
- fix inaccurate git state issue when creating new branch with name conflict and `autoCheckoutOnBranchCreate` is enabled.
  - Add content refresh on .gitignore file change
  - fix reference filtering


components/branches/branches.html
```diff
@@ -8,8 +8,12 @@
   </button>
   <ul class="dropdown-menu pull-right" role="menu">
     <div class="option" onclick="event.stopPropagation()">
-      Only local branch?
-      <input type="checkbox" data-bind="checked: isLocalBranchOnly, "></input>
+      Remote
```
campersau: A `<label>` around the text and the checkbox would be nice. Then you can click both the text and the checkbox to toggle it.

PR #1075, opened 2018-04-07
Support alternate git directory

jpschewe:
This PR has changes to support a git directory that isn't in the standard location. It leverages the use of `git rev-parse --absolute-git-dir` to do the work of finding the actual git directory.

This fixes FredrikNoren/ungit/#1013.

### DIFF
--- a/source/git-api.js
+++ b/source/git-api.js
@@ -47,7 +47,7 @@ exports.registerApi = (env) => {
             if (!isMac && !isWindows) {
               // recursive fs.watch only works on mac and windows
               const promises = [];
-              const gitDir = gitParser.findGitDir(data.path)
+              const gitDir = gitPromise.findGitDir(data.path)
 
               promises.push(watchPath(socket, path.join(gitDir, 'HEAD')));
               promises.push(watchPath(socket, path.join(gitDir, 'refs', 'heads')));
@@ -559,7 +559,7 @@ exports.registerApi = (env) => {
   });
 
   app.get(`${exports.pathPrefix}/baserepopath`, ensureAuthenticated, ensurePathExists, (req, res) => {
-    const gitDir = gitParser.findGitDir(req.query.path)
+    const gitDir = gitPromise.findGitDir(req.query.path)
 
     jsonResultOrFailProm(res, gitPromise(['rev-parse', '--show-toplevel'], gitDir)
       .then((baseRepoPath) => {
@@ -603,7 +603,7 @@ exports.registerApi = (env) => {
     const task = gitPromise(['submodule', 'deinit', "-f", req.query.submoduleName], req.query.path)
       .then(gitPromise.bind(null, ['rm', '-f', req.query.submoduleName], req.query.path))
       .then(() => {
-        const gitDir = gitParser.findGitDir(req.query.path)
+        const gitDir = gitPromise.findGitDir(req.query.path)
 
         rimraf.sync(path.join(req.query.path, req.query.submodulePath));
         rimraf.sync(path.join(gitDir, 'modules', req.query.submodulePath));
--- a/source/git-api.js
+++ b/source/git-api.js
@@ -47,10 +47,12 @@ exports.registerApi = (env) => {
             if (!isMac && !isWindows) {
               // recursive fs.watch only works on mac and windows
               const promises = [];
-              promises.push(watchPath(socket, path.join('.git', 'HEAD')));
-              promises.push(watchPath(socket, path.join('.git', 'refs', 'heads')));
-              promises.push(watchPath(socket, path.join('.git', 'refs', 'remotes')));
-              promises.push(watchPath(socket, path.join('.git', 'refs', 'tags')));
+              const gitDir = gitParser.findGitDir(data.path)
+
+              promises.push(watchPath(socket, path.join(gitDir, 'HEAD')));
+              promises.push(watchPath(socket, path.join(gitDir, 'refs', 'heads')));
+              promises.push(watchPath(socket, path.join(gitDir, 'refs', 'remotes')));
+              promises.push(watchPath(socket, path.join(gitDir, 'refs', 'tags')));
               return Bluebird.all(promises);
             }
           }).catch((err) => {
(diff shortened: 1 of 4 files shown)


jpschewe:
Note that I have not added any test cases for this functionality. I could use some pointers on how to do that and then I can add a test case for this.

source/git-parser.js
```diff
@@ -336,3 +337,10 @@ exports.parsePatchDiffResult = (patchLineList, text) => {
     return null;
   }
 }
+
+// determine the git directory, always an absolute path
+// TODO: This method does not handle errors, it assumes that repoPath is a valid git repository path
+exports.findGitDir = (repoPath) => {
+  const gitDir = child_process.execSync('git rev-parse --absolute-git-dir', {cwd: repoPath}).toString().trim();
+  return gitDir;
+}
```
jung-kim: `git-parser.js` is not the place for executing git commands.  That is within `git-promise.js`.


`git-parser` is exclusively for parsing the git output

source/git-parser.js
```diff
@@ -336,3 +337,10 @@ exports.parsePatchDiffResult = (patchLineList, text) => {
     return null;
   }
 }
+
+// determine the git directory, always an absolute path
+// TODO: This method does not handle errors, it assumes that repoPath is a valid git repository path
+exports.findGitDir = (repoPath) => {
+  const gitDir = child_process.execSync('git rev-parse --absolute-git-dir', {cwd: repoPath}).toString().trim();
```
jung-kim: Please use async version of `execSync` as it is more performant

source/git-api.js
```diff
@@ -106,12 +108,21 @@ exports.registerApi = (env) => {
       return false;  // ignore files that are in .gitignore
     } else if (filename.endsWith(".lock")) {
       return false;
-    } else if (filename.indexOf(path.join(".git", "refs")) > -1) {
-      return true;   // trigger for all changes under refs
-    } else if (filename == path.join(".git", "HEAD")) {
-      return true;   // Explicitly return true for ".git/HEAD" for branch changes
+    } else if (filename.indexOf(path.join("", "refs")) > -1) {
```
jung-kim: Let's remove empty string as it doesn't needs to be there.

source/git-api.js
```diff
@@ -106,12 +108,21 @@ exports.registerApi = (env) => {
...
+      // trigger for all changes under directory named refs
+      // NOTE: This check will catch some other refs directories, but we can't look for .git
+      // in case the git directory has been relocated
+      return true;
+    } else if (filename.endsWith("HEAD")) {
```
jung-kim: Let's not do this as `HEAD` can be a git directory.  There is no reason to change this

jung-kim:
Let's fix existing tests first and worry about adding tests later.

I'm okay with fundamental idea of finding the git directory via using `git rev-parse --absolute-git-dir` but the way you are proposing requires two things.

1. Adding additional git commands to every single git api request
2. Making changes for every single api end points.

Neither of those two are fun problems and may need to think alternate way with caching or different methods

source/git-api.js
jpschewe: Done

source/git-api.js
jpschewe: I thought the point if this if statement was that ungit needs to watch the HEAD file. Given that the git directory may not be named ".git" the current check isn't going to watch the HEAD file in those git directories.

jpschewe:
If there's a place that ungit knows that the current directory has changed, I could cache the value of gitdir there. I didn't look very hard yet as I just wanted to get something working. Do you know of such a location?

I think all of the end points need to be touched, at least to reference the cached value.

I started looking at the test failures and I tried running the tests locally, but many more failed than what continuous integration said. So I expect I've got something wrong with my environment. 

I also started looking into making the method return a promise rather than being a direct call to git. This may help with the test failures as well. I figured this was more in line with the other methods. However my initial attempt has failed. If you could take a look at the branch promise in my repository, that would be great. 

I will be rather busy over the next week, so may not get many changes in quickly.

source/git-api.js
jpschewe: @jung-kim Does my reasoning make sense here? Or am I not understanding what this piece of code is trying to do?

jpschewe:
@jung-kim do you have any suggestions on my last comment?

source/git-api.js
jpschewe: @jung-kim does this make sense?

jpschewe:
@jung-kim 2 things:
1) Is there anywhere that the current git working directory is stored?
2) Can you look at the branch promise and see what I've done that is wrong there?

Thank you

jung-kim:
1. No, Ungit's design is stateless. That is why git repo value is passed in per every requests
2. I'd rather not because value of this fix is questionable at best fix requires large changes.

jpschewe:
1. The current working directory must be stored in the UI though, otherwise it can't be passed to the API calls, correct?

jung-kim:
No we do not, state of which git repo is at the browser and not at the server.

If we take a look at the simpler API call such as [`/show`](https://github.com/FredrikNoren/ungit/blob/master/source/git-api.js#L325), we can see that `req.query.path` contains the git repo path.

Which is then gets passed through few functions and eventually get's passed in as a [current working directory](https://github.com/FredrikNoren/ungit/blob/master/source/git-promise.js#L34) argument for the spawned child process.

Does this answer your question?  git repo locations are passed in per every single API calls, which allows us to do keep server stateless and simpler

jpschewe:
Yes, I see that the API is stateless. So if it's too show to determine the git directory on every call, then either the UI needs to keep this state little it does the working directory, or the API needs to add a cache making working directory to git directory.

jung-kim:
We could do it that way.  Another way is do it on front end.  On git repo navigation resolve git repo path, and block all other git api callas until git repo path is resolveed.  

Either way it requires significant coordinations.

source/git-api.js
jpschewe: @jung-kim Does this make sense?

jung-kim: No, two reasons:

1. in corrected git repo after resolved via `git rev-parse`, `.../.git/HEAD` is the file and it should exist.
2. This will capture any and all file with names that ends with `HEAD`.  it's a simple file watch so it maybe innocuous but it's not specific enough.

jpschewe: However not all git directories are named ".git". This is the reason for this pull request. So `HEAD` won't be watched at all in these repositories. That's why I made the change to look at all `HEAD` files.

jung-kim: `git rev-parse` that you are suggesting will give us the correct git repo with the `.git` folder.

jung-kim: Okay, I will await your solution.

jpschewe:
I have decided to go with @jung-kim suggestion of not developing inside my submodules. So I won't be working on this merge request. Others may be interested in it if they want to use the alternate git directory functionality in git.

jpschewe:
If someone wants to continue development here I would suggest that there be a cache of working directory to git directory stored in the library to avoid the system call on every operation.

PR #1107, opened 2018-06-25
Use ungit without Internet access

gigi206:
If you have no access to Internet, you have a black page and you cannot use ungit


public/js/raven.min.js
```diff
@@ -0,0 +1,3 @@
...
+//# sourceMappingURL=raven.min.js.map
```
campersau: Maybe remove `sourceMappingURL` or also include `raven.min.js.map` file.

jung-kim:
Ah... instead of checking in the code can we include https://www.npmjs.com/package/raven in  `package.json` and reference `https://github.com/getsentry/raven-js/blob/master/dist/raven.min.js` via browserify?

gigi206:
> Ah... instead of checking in the code can we include https://www.npmjs.com/package/raven in package.json and reference https://github.com/getsentry/raven-js/blob/master/dist/raven.min.js via browserify?

I don't know :) I use ungit for devops (I'm system ingineer, not real developper and I use ungit with [VScode plugin](https://github.com/Hirse/vscode-ungit)). The main goal is that the installation must be done online and if there is no access to the Internet I must be able to use ungit.

I change the code the most simply for it works for me offline :)

jung-kim:
I'm going to merge this for now, but will remove checked in generated code and release it as that is not the best way to go about this.

Issue #1105, opened 2018-06-14
Implement GIT_OPTIONAL_LOCKS


stdedos:
I have been spammed a lot of times with "fatal: Unable to create '[...]/.git/index.lock': File exists." during a `rebase`.

There are many occasions, however, that commands that don't strictly need the lock take it anyway (most notorious command being `git status`)

Since a lot of people are affected, git >2.15.0 added https://github.com/git/git/commit/27344d6a6c8056664966e11acf674e5da6dd7ee3 this.

I can obviously start the command with `env GIT_OPTIONAL_LOCKS=0 ungit`; however, I thought you might be interested to take a more fine-tuned / less crude look at it.

jung-kim:
This this is a big problem for Ungit and we've added in the retry logic to get around it but clearly that doesn't 100% eliminate the problem.

I've added `--no-optional-locks` to the status end point only via #1109 when git version is appropriate.  

Thanks for bringing this issue up!

stdedos:
@jung-kim I am not sure what you would mean by "added in the retry logic"; however, you cannot retry during `git rebase` (or I don't know yet how to continue when that breaks).

_Yeah, `rebase` is of the satan, but I like it 😸_

Issue #1093, opened 2018-05-18
sometimes ungit makes it seem like commits are on a single subtree when they are not


capaj:
I have this tree:

![image](https://user-images.githubusercontent.com/1305378/40223320-e8361758-5a82-11e8-8a90-a69fdb5ccf53.png)

which ungit renders such as it seems like the dev-1032 is continuing from to-asyn-await branch, which it is not.
See this rendering of the same tree when I check out dev-1032: 
![image](https://user-images.githubusercontent.com/1305378/40223396-3a8e77ca-5a83-11e8-9a31-bc8a71321d5d.png)

It would be nice if ungit could render the tree branches more separated. Maybe coloring the connections between commits as well as the commits would help?

jung-kim:
Yes this is an issue and it is bit hard to envision a branch spacing that can remove this issue perfectly and consistently for all cases.  

I'ver personally tried different spacings and rendering logics but none seemed like a clear winner.

Therefore, I'm going to have it so that when a mouse hover over the line would highlight the line and the two nodes at the end of the line so one can easily demystify any confusion in case of doubt with a mouse hover.

How does that sound?

capaj:
That would certainly be a step in the right direction. Even though it's unfortunate that in order to discriminate those branches, I'll have to hover over it, but certainly better than the current state.

Issue #988, opened 2017-11-11
Require authentication on remote clients


KSXGitHub:
Let's say that my computer is connected to a local area network (such as wifi), my own ip address is `192.168.1.5` and `::ffff:192.168.1.5`. I run `ungit --port=8448` so every device in my LAN can go to `http://192.168.1.5:8448` and edit my repo. I could just use `--allowedIPs` but it's more convenient to be able to access my repo from other devices.

### Request

1. The following hosts should be able to freely access ungit without authentication:
  * `127.0.0.1`
  * `::1`
  * `::ffff:127.0.0.1`
  * `192.168.1.5`
  * `::ffff:192.168.1.5`

2. The other hosts must login in order to use ungit in my machine

gigi206:
I agree, for security reasons ungit must only allow local adresses.

How exactly works these options **--allowedIPs** and **--authentication --users** ?

I tried :
```
--allowedIPs='localhost'
--allowedIPs="['localhost']"
```

But I obtain the following message :
_This host is not authorized to connect
You are trying to connect to an Ungit instance from an unathorized host._

jung-kim:
`--allowedIPs="['localhost']"` should be the correct syntax.  However, the reason why that isn't working is because there is because ungit doesn't resolve given [`allowedIPs`](https://github.com/FredrikNoren/ungit/blob/master/source/server.js#L88-L98).  

yes this is very rudimentary and not pedantic at all and we could introduce the [ip](https://www.npmjs.com/package/ip) npm module to do more precise matching but frankly it was a feature that wasn't used quite often so it was bit neglected but will fix it up soon if no body touches it by the time I get to it.

gigi206:
For my point of view the main problem is a security issue : Ungit must bind by default on 127.0.0.1 and not 0.0.0.0

jung-kim:
Oh wow, I thought it would default bind to 127.0.0.1 but I [guess not](https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback).  Has this changed or has it always been like this for the `http`?  

I think there is a better way to start the server with express now but will visit that later.

@gigi206 you are right this did surprise me and I've totally missed your point.  Is there anything else I could add?

gigi206:
No juste bind by default on 127.0.0.1 is good in my point of view :)

You can check with **lsof** on linux : 
```
$ lsof -i:8448
```
[...]

PR #1109, opened 2018-07-01
Force bind to '127.0.0.1' and add `--no-optional-locks` when applicable

### LINKED ISSUE #1110: Update the VSCode extension in README
The VSCode extension listed in README (https://github.com/RomanValihura/vscode-ungit-tab) is now no longer maintained. I suggest listing this extension instead: https://marketplace.visualstudio.com/items?itemName=Hirse.vscode-ungit

jung-kim:
- bind do `127.0.0.1` on default (ref #988)
- utilize `--no-optional-locks` (ref #1105)
- node highlight on path mouse hover (ref #1093)
- fix #1110

### DIFF
--- a/source/config.js
+++ b/source/config.js
@@ -7,6 +7,7 @@ const yargs = require('yargs');
 const homedir = require('os-homedir')();
 const winston = require('winston');
 const child_process = require('child_process');
+const semver = require('semver');
 
 const defaultConfig = {
 
@@ -281,3 +282,5 @@ if (fs.existsSync(path.join(__dirname, '..', '.git'))){
 } else {
   module.exports.ungitDevVersion = module.exports.ungitPackageVersion;
 }
+
+module.exports.isGitOptionalLocks = semver.satisfies(module.exports.gitVersion, '2.15.0');
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ Use the following format for additions: ` - VERSION: [feature/patch (if applicab
 - 1.4.29:
   - Add `--no-optional-locks` if git version is appropriate [#1105](https://github.com/FredrikNoren/ungit/issues/1105)
   - Ensure ungit server to bind to `127.0.0.1` [#988](https://github.com/FredrikNoren/ungit/issues/988)
+  - Add node highlight on mouse hover on relationsip path [#1093](https://github.com/FredrikNoren/ungit/issues/1093)
 - 1.4.28: adding raven locally for offline access. [#1107](https://github.com/FredrikNoren/ungit/pull/1107)
 - 1.4.27: logic change for the merge conflict resolution
 - 1.4.26: add a way to preconfigure repo lists [#1106](https://github.com/FredrikNoren/ungit/issues/1106)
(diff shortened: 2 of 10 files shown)


source/config.js
```diff
@@ -281,3 +282,5 @@ if (fs.existsSync(path.join(__dirname, '..', '.git'))){
 } else {
   module.exports.ungitDevVersion = module.exports.ungitPackageVersion;
 }
+
+module.exports.isGitOptionalLocks = semver.satisfies(module.exports.gitVersion, '2.15.0');
```
stdedos: IMHO you could simply set `GIT_OPTIONAL_LOCKS=0` on the parent's env and be done with it. I don't think that `git` will look twice if it doesn't know what  `GIT_OPTIONAL_LOCKS` is

jung-kim: My original thought is to limit this to `git status` only as that is the biggest offender for this issue.  

If we do that it will be applying to all git operations right?  I maybe too cautious but I would like to limit this to an end point and observe

jung-kim: Hmm so are you suggesting that we set env `GIT_OPTIONAL_LOCKS=0` env variable and override this env variable settings by doing `--optional-locks` on `rebase`, `merge`, and `cherry-pick`?  (I would also add `patch` to list as well)

stdedos: I think I didn't explain myself thoroughly.

If you think the `git` repository as a FSM, all those operations (`rebase`, `merge`, `cherry-pick`, `patch`, `am` ... etc) "change the repository's state" before they return back to the "idle" state when they are finished.

If you run on Unix, and you have git_ps1 set, you will see it.
(Examples [this](https://pagefaultblog.files.wordpress.com/2017/02/git-prompt2.png?w=685) from [here](https://pagefault.blog/2017/02/16/improved-git-cli-with-git-completion-and-git-prompt/), or [that](https://raw.githubusercontent.com/arialdomartini/oh-my-git-gh-pages/master/images/samples/14-action-in-progress.jpg) from [there](https://github.com/arialdomartini/oh-my-git))

_During_ that time, _it would be nice_ to set `GIT_OPTIONAL_LOCKS=0` on the `git`'s environment.

----

In that sense, you can have the best of two worlds: _Not changing_ current behavior and opportunistically still updating index, _unless_ in "a critical state" in the repository.

jung-kim: Considering the doc explicitly calls out git status running on background, let's do this first without setting env for other operations.

https://git-scm.com/docs/git-status#_background_refresh

chrjorgensen:
I'm on the same page as @tsuckow - this change makes it impossible to use ungit on a server, in my case an IBM i server. Ungit ran perfectly on IBM i with each user connecting to a different port on the common external ip address.

Please make this configurable...

jung-kim:
Yes the fix #1113 is out but I'm having problem publishing it as all npm token are invalidated due to recent security incident.

Will try to publish it as soon as I can but once we do you can set `ungitBindIp` to a blank string and it will work as before

jung-kim:
ungit@1.4.31 is cut now and this field is configurable.

Issue #1112, opened 2018-07-11
Cannot use ungit in cloud9


tsuckow:
ungit explicitly binds to the loopback adapter as of https://github.com/FredrikNoren/ungit/commit/d893f3a045caa0eda6c3bfae9ee55625789c8801
Which prevents using ungit in cloud9. The bind ip should be configurable or use the $IP environmental variable.

jung-kim:
Shoot...  `127.0.0.1` binding is preferable in most of the cases and I guess I now see the where this is an exception, cloud9.

Will be pushing a simple config to allow you to do default binding

PR #1113, opened 2018-07-11
Configurable binding ip

jung-kim:
fix: #1112

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 Use the following format for additions: ` - VERSION: [feature/patch (if applicable)] Short description of change. Links to relevant issues/PRs.`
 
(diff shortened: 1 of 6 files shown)


source/config.js
```diff
@@ -148,6 +148,9 @@ const defaultConfig = {
 
   // Array of local git repo paths to display at the ungit home page
   defaultRepositories: [],
+
+  // is force ungit server to bind to `127.0.0.1`
+  isBindToLocalhost: true,
```
campersau: Why not using a string?
If someone has multiple networkinterfaces you can bind to a specific one.

jung-kim: you are right, will touch that up soon.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 Use the following format for additions: ` - VERSION: [feature/patch (if applicable)] Short description of change. Links to relevant issues/PRs.`
 
+- 1.4.30: Add `isBindToLocalhost` config to allow default binding in some cases[#1112](https://github.com/FredrikNoren/ungit/issues/1112)
```
campersau: Also change to `ungitBindIp` and maybe add a space before the issue link?

PR #1126, opened 2018-08-10
Wrap localStorage to not depend on it

Hirse:
This change wraps usages of localStorage with a custom storage object so that ungit still works in environments where localStorage is not available. It is not an ideal solution and does not provide the same features as actual localStorage, but is intended to not break.

Background: I am trying to update the [VSCode Ungit extension](https://github.com/Hirse/vscode-ungit) to use the new, better [VSCode WebView API](https://code.visualstudio.com/docs/extensions/webview), but am blocked due to the [missing localStorage support](https://github.com/Microsoft/vscode/issues/52246).

### DIFF
--- a/public/source/storage.js
+++ b/public/source/storage.js
@@ -1,0 +1,34 @@
+/**
+ * A wrapper aroung LocalStorage to support environments where LocalStorage is not available.
+ * Stores and retrieves items from LocalStorage if available and uses a non-persistent cache otherwise.
+ */
+var storage = {};
+module.exports = storage;
+
+storage.cache = {};
+
+/**
+ * The getItem() method of the Storage interface, when passed a key name, will return that key's value or null if the key does not exist.
+ * @param {string} key A DOMString containing the name of the key you want to retrieve the value of.
+ * @returns {string | null} A DOMString containing the value of the key. If the key does not exist, null is returned.
+ */
+storage.getItem = function (key) {
+    if (localStorage) {
+        return localStorage.getItem(key);
+    } else {
+        return this.cache[key] || null;
+    }
+};
+
+/**
+ * The setItem() method of the Storage interface, when passed a key name and value, will add that key to the storage, or update that key's value if it already exists.
+ * @param {string} key A DOMString containing the name of the key you want to create/update.
+ * @param {string} value A DOMString containing the value you want to give the key you are creating/updating.
+ */
+storage.setItem = function (key, value) {
+    if (localStorage) {
+        localStorage.setItem(key, value);
+    } else {
+        this.cache[key] = value;
+    }
+};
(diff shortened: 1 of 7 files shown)


campersau:
Also what about these?

https://github.com/FredrikNoren/ungit/blob/06af3ae16adb828817a9cb37628f0e570cae8b72/components/stash/stash.js#L51

https://github.com/FredrikNoren/ungit/blob/06af3ae16adb828817a9cb37628f0e570cae8b72/components/stash/stash.js#L87

Hirse:
Thanks for the review. I have updated the implementation and included missing occurrences of `localStorage`.

public/source/storage.js
```diff
@@ -0,0 +1,17 @@
...
+    storage.setItem = localStorage.setItem;
+} catch (e) { /* Ignore Exception, use fallback implementation. */ }
+
+if (!storage) {
+    var cache = Object.create(null);
+    storage.getItem = function (key) { return cache[key] || null; };
+    storage.setItem = function (key, value) { cache[key] = value; };
+}
+
+module.export = storage;
```
campersau: `module.exports`

public/source/storage.js
```diff
@@ -0,0 +1,17 @@
...
+ * A wrapper around LocalStorage to support environments where LocalStorage is not available.
+ * Stores and retrieves items from LocalStorage if available and uses a non-persistent cache otherwise.
+ */
+var storage = {};
+try {
+    storage.getItem = localStorage.getItem;
+    storage.setItem = localStorage.setItem;
+} catch (e) { /* Ignore Exception, use fallback implementation. */ }
+
+if (!storage) {
```
campersau: This will not work because `storage` is already an empty `object` (`{}`) which is `true` when converted to a boolean in JavaScript.

public/source/storage.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * A wrapper around LocalStorage to support environments where LocalStorage is not available.
+ * Stores and retrieves items from LocalStorage if available and uses a non-persistent cache otherwise.
+ */
+var storage = {};
+try {
+    storage.getItem = localStorage.getItem;
```
campersau: Assigning a function to another object will lose the `this` context and results in a `TypeError` upon invocation. That's why I used `bind`. 😉 
`localStorage.getItem.bind(localStorage)`

Hirse:
Using your implementation and fixed the merge conflict.

jung-kim:
Thanks guys!

Hirse:
Great! Thanks, @campersau for your help.

jung-kim:
Absolutely, I'm a clumsy coder so I miss a lot of details and he definitely helps out so much. 

And... speaking of me being clumsy, I've forgot to ask for modification of `CHANGE_LOG.md` and `package.json` files for the release of this.

If you have some free time, an you please send me a PR with updated changelog, package.json and package-lock.json?  

If not I will create one later this weekends.

PR #1127, opened 2018-08-10
Use ESLint and fix easy issues

Hirse:
Replaces JSHint with ESLint.
The ESLint config is declared in its own file instead of the Gruntfile to enable editor support. 

I fixed all easy issues (mostly missing semicolons and unused variables), but there is one issue where I wasn't sure how to fix it:
https://github.com/Hirse/ungit/blob/9c63a0519645fdc170ac83cd0dd3927b42bfd595/test/spec.git-api.diff.js#L22


jung-kim:
Is there a strong compelling reason to replace JSHint with ESLint?  

I know there was an interesting fight about JSHint vs JSLint but don't know much about ESLint.  Unless there is strong compelling reason I'd rather not introduce a big change.

Hirse:
Generally, more rules, plugins, support for new versions of JS, and more comfortable configuration.
But I guess, it is always a matter of preference to some degree.

Honestly, I started by adding ESLint and didn't realize you were using JSHint until I was halfway through with this PR based on the large number of warnings I found.

If you don't want to switch, but are fine with the cleanup changes, I can change the PR/create a new one.

ylecuyer:
I ran into the lack of async/await support for jshint and I think it would be good to switch if eslint is more uptodate with the new js.

@Hirse for the line you pointed out https://github.com/Hirse/ungit/blob/9c63a0519645fdc170ac83cd0dd3927b42bfd595/test/spec.git-api.diff.js#L22, I think it is a copy/paste error. based on the same line before it should be:

```
      .then((dir) => testBareDir = dir);
```

Hirse:
@ylecuyer I think ESLint is more up to date and has better support in general.

That explanation makes sense. Do you want to fix it in another PR?

ylecuyer:
I did it in #1174

Hirse:
Closing this as it is so outdated.

PR #1129, opened 2018-08-17
Update Changelog and version for localStorage wrapper

Hirse:
@jung-kim Is this what you had in mind? If you don't want to release yet, you can just take the first commit.

For PR #1126


jung-kim:
Yes! sorry, bene crazy busy but this is spot on.

Thanks!

PR #1142, opened 2018-09-24
change .stagging background from red to green

yukulele:
red seem like error message

### DIFF
--- a/components/staging/staging.less
+++ b/components/staging/staging.less
@@ -2,7 +2,7 @@
 @import "public/less/variables.less";
 
 .staging {
-  background: #643A44;
+  background: #3A6444;
   box-shadow: 0px -1px 15px #252833;
   color: #B8A5A5;
   z-index: 5;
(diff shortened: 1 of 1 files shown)


jung-kim:
Ah... Is there a particular reasoning behind this change?

yukulele:
this look like an error message:
screenshot_20180924_200540

this look like a simple information:
screenshot_20180924_200628

Hirse:
I agree that red usually looks like an error message, but between the two screenshots the red looks like it fits better into the theme.
The green definitely has some contrast issues with the light-blue link text.

What about using a more neutral blue-gray?
(Green also looks like success and that message doesn't have either connotation.)

yukulele:
yes, i don't chose blue because of link color (i don't wont to change all the styles) but it's probably a better choice

yukulele:
Try with neutral blue-grey:

screenshot_20181024_113610

yukulele:
@Hirse I updated pull request

Hirse:
Could you also post a new image when you update the color, please?

Additionally, does this also affect the background of the commit area (when there are changes)? Should it?

yukulele:
it's the last screenshot https://github.com/FredrikNoren/ungit/pull/1142#issuecomment-432588078

yes, it also affect commit area

yukulele:
screenshot_20181031_111248

Issue #1139, opened 2018-09-15
Ungit don't detect that a tag was pushed on the remote


gigi206:
Git tag **v1.0.0** and **v1.1.0** are pushed on the remote, but only tag **v1.0.0** is detected on the remote.

If I click on tag **v1.1.0** ungit displays the push button.

I have already tried to erase the tag **v1.1.0** from github and click push again with ungit but this is the same result.

```
$ git ls-remote --tags
From https://github.com/xxx/xxx.git
aeab362abe50f5e9ad8cf98b79e7cc522d2e8c97        refs/tags/v1.0.0
f6191bf1b622266a6928d667550f5c3a518357ad        refs/tags/v1.0.0^{}
3db2b40a60395c1944957fd3d345f2d2dbeb8dce        refs/tags/v1.1.0
```

Ungit see only one pushed :
![](https://image.ibb.co/npVjme/1.png)

![](https://image.ibb.co/m7D9Kz/2.png)

jung-kim:
Is that your ungit? `https://github.com/xxx/xxx.git`

Because `https://github.com/FredrikNoren/ungit/tags` doesn't show v1.1.0 and I don't think it was ever published.

I think the wrong remote is targeted when you try to push the tag. When you push tag or branches, make sure correct remote is targeted in the remote section.  

This is a bit confusing part of the current UI and please let me know if you have any suggestions.

gigi206:
I replace the true url with **xxx** but the remote exists. The remote is :
```
$ git remote show origin
* remote origin
  Fetch URL: https://github.com/gigi206/VSCode-Anywhere.git
  Push  URL: https://github.com/gigi206/VSCode-Anywhere.git
  HEAD branch: master
  Remote branches:
    dev    tracked
    master tracked
    old    tracked
  Local branches configured for 'git pull':
    dev    merges with remote dev
    master merges with remote master
  Local refs configured for 'git push':
    dev    pushes to dev    (local out of date)
    master pushes to master (up to date)
```

We can see on previous images that the tag **v1.1.0** is only detected locally by ungit. And the tag **v1.0.0** is detected on the remote and locally by ungit. But we can see that tag **v1.1.0** exists on the remote :
```
$ git ls-remote --tags
From https://github.com/gigi206/VSCode-Anywhere.git
aeab362abe50f5e9ad8cf98b79e7cc522d2e8c97    refs/tags/v1.0.0
[...]

PR #1144, opened 2018-10-07
Ensure addition of remote tags

jung-kim:
fix: #1139 

Ensure all remote tags are added to ungit, even when they don't have the deref tags.  

(which I'm still not sure how it was induced)


campersau:
A test would be nice!

jung-kim:
I would absolutely agree, but I'm not sure how this git state is created normally.  All create remote tags operations I know created two tags, `$TAG_NAME` and `$TAG_NAME^{}` with `^{}` being the direct reference to the actual tag location.

So in order for me to create test for this I would have fake the data in test and that poses different challenges

PR #1152, opened 2018-10-26
Use glyphicon for commit amend checkbox

Hirse:
Using the glyphicons makes it easier to see the state of the checkbox.


components/staging/staging.html
```diff
@@ -15,7 +15,9 @@
         <textarea class="form-control" rows="2" placeholder="Body" data-bind="value: commitMessageBody, valueUpdate: 'afterkeydown', enable: !inRebase(), event: {keypress: onAltEnter}"></textarea>
         <div>
           <span class="amend" data-bind="visible: canAmend">
-            <span class="checkmark" data-bind="css: { checked: amend }">&#10003;</span>
+            <div class="checkmark" data-bind="click: toggleAmend, css: { checked: amend }">
```
ylecuyer: the click handler can be moved up to the `span.amend` element and removed from both `div.checkmark` and `a.amend-link`

jung-kim: Yeah I think that would be a better idea

Hirse:
Moved the click handler and switched the link to a button for accessibility.

PR #1154, opened 2018-10-26
Fix branch search box

Hirse:
Selection from the branch search autocomplete fails because `this` is set to the jQuery object, not the class instance:


jung-kim:
Thanks!

(As you may know this will not be released until the next version bump)

Hirse:
Yes, that's fine.
I didn't actually use that feature, I just found the bug while working on the other changes.

PR #1158, opened 2018-10-30
fix changing remotes in remotes dropdown

zkxs:
When changing remotes using the dropdown in ungit 1.4.35, the following error occurs:

Firefox 64
```
TypeError: this.currentRemote is not a function[Learn More] remotes.js:97
    changeRemote remotes.js:97
    init/</< knockout:99
    dispatch jquery:5182
    add/elemData.handle jquery:4991
```

Chrome 69
```
Uncaught TypeError: this.currentRemote is not a function
    at Object.changeRemote (remotes.js:97)
    at HTMLAnchorElement.<anonymous> (knockout:99)
    at HTMLAnchorElement.dispatch (jquery:5183)
    at HTMLAnchorElement.elemData.handle (jquery:4991)
```

This issue prevents the user from changing remotes.

### DIFF
--- a/components/remotes/remotes.js
+++ b/components/remotes/remotes.js
@@ -92,7 +92,6 @@ class RemotesViewModel {
   updateRemotes() {
     return this.server.getPromise('/remotes', { path: this.repoPath() })
       .then(remotes => {
-        const outerThis = this;
         remotes = remotes.map(remote => ({
           name: remote,
           changeRemote: () => { this.currentRemote(remote) }
--- a/components/remotes/remotes.js
+++ b/components/remotes/remotes.js
@@ -95,7 +95,7 @@ class RemotesViewModel {
         const outerThis = this;
         remotes = remotes.map(remote => ({
           name: remote,
-          changeRemote() { outerThis.currentRemote(remote) }
+          changeRemote: () => { this.currentRemote(remote) }
         }));
         this.remotes(remotes);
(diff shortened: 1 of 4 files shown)


components/remotes/remotes.js
```diff
@@ -92,9 +92,10 @@ class RemotesViewModel {
   updateRemotes() {
     return this.server.getPromise('/remotes', { path: this.repoPath() })
       .then(remotes => {
+        const outerThis = this;
         remotes = remotes.map(remote => ({
           name: remote,
-          changeRemote() { this.currentRemote(remote) }
+          changeRemote() { outerThis.currentRemote(remote) }
```
ylecuyer: This would be better at it is more ES6 compliant
And no need of outerThis

```suggestion
          changeRemote: () => { outerThis.currentRemote(remote) }
```

components/remotes/remotes.js
```diff
@@ -92,9 +92,10 @@ class RemotesViewModel {
   updateRemotes() {
     return this.server.getPromise('/remotes', { path: this.repoPath() })
       .then(remotes => {
+        const outerThis = this;
```
ylecuyer:
```suggestion
```

zkxs:
@ylecuyer's suggestion looks good. I wasn't happy about `outerThis` so I'm glad to see it go.

PR #1162, opened 2018-11-12
Adding more sleep to discard test for windows

jung-kim:


jung-kim:
Consistent click test is bit trickier than I thought...

PR #1172, opened 2019-01-04
Better specs for git-parser + cleaning

ylecuyer:
Line coverage before:

 ```
----------------|----------|----------|----------|----------|-------------------|
File            |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
----------------|----------|----------|----------|----------|-------------------|
All files       |    34.06 |    20.59 |     32.5 |    35.62 |                   |
  src            |    33.33 |       21 |    33.33 |    34.88 |                   |
   git-parser.js |    33.33 |       21 |    33.33 |    34.88 |... 58,360,361,363 |
  src/utils      |       75 |        0 |        0 |       75 |                   |
   file-type.js  |       75 |        0 |        0 |       75 |                 7 |
----------------|----------|----------|----------|----------|-------------------|
```

Line coverage now:

```
----------------|----------|----------|----------|----------|-------------------|
File            |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
----------------|----------|----------|----------|----------|-------------------|
All files       |    98.25 |    91.18 |      100 |    99.09 |                   |
 src            |    98.22 |       92 |      100 |    99.07 |                   |
  git-parser.js |    98.22 |       92 |      100 |    99.07 |           142,143 |
 src/utils      |      100 |       50 |      100 |      100 |                   |
  file-type.js  |      100 |       50 |      100 |      100 |                 7 |
----------------|----------|----------|----------|----------|-------------------|
```
I have cleaned some code too, I'll highlight in the comments

### DIFF
--- a/source/git-parser.js
+++ b/source/git-parser.js
@@ -74,7 +74,7 @@ const gitLogHeaders = {
   'Reflog': (currentCommmit, data) => {
     currentCommmit.reflogId = /\{(.*?)\}/.exec(data)[1];
     currentCommmit.reflogName = data.substring(0, data.indexOf(' ')).replace("refs/", "");
-    const author = data.substring(data.indexOf(' ') + 2, data.length - 1);
+    const author = data.substring(data.indexOf('(') + 1, data.length - 1);
     const capture = authorRegexp.exec(author);
     if (capture) {
       currentCommmit.reflogAuthorName = capture[1].trim();
--- a/source/git-parser.js
+++ b/source/git-parser.js
@@ -291,7 +291,7 @@ const updatePatchHeader = (result, lastHeaderIndex, ignoredDiffCountTotal, ignor
 }
 
 exports.parsePatchDiffResult = (patchLineList, text) => {
-  if (!text) return '';
+  if (!text) return null;
 
   const lines = text.trim().split('\n');
   const result = [];
--- a/source/git-parser.js
+++ b/source/git-parser.js
@@ -291,7 +291,7 @@ const updatePatchHeader = (result, lastHeaderIndex, ignoredDiffCountTotal, ignor
 }
 
 exports.parsePatchDiffResult = (patchLineList, text) => {
-  if (!text) return {};
+  if (!text) return '';
 
   const lines = text.trim().split('\n');
   const result = [];
--- a/test/spec.git-parser.js
+++ b/test/spec.git-parser.js
@@ -165,3 +165,33 @@ describe('parseGitLsRemote', () => {
     ]);
   });
 });
+
+describe('parseGitStatusNumstat', () => {
+  it('parses the git status numstat', () => {
+    var gitStatusNumstat = '1459\t202\tpackage-lock.json\n'
+    gitStatusNumstat += '2\t1\tpackage.json\n'
+    gitStatusNumstat += '13\t0\ttest/spec.git-parser.js'
+    
+
+    expect(gitParser.parseGitStatusNumstat(gitStatusNumstat)).to.eql({
+      "package-lock.json": { additions: "1459", deletions: "202" },
+      "package.json": { additions: "2", deletions: "1" },
+      "test/spec.git-parser.js": { additions: "13", deletions: "0" }
+    });
+  })
+
+  it('skips empty lines', () => {
+    var gitStatusNumstat = '1459\t202\tpackage-lock.json\n'
+    gitStatusNumstat += '\n'
+    gitStatusNumstat += '\n'
+    gitStatusNumstat += '2\t1\tpackage.json\n'
+    gitStatusNumstat += '13\t0\ttest/spec.git-parser.js'
+    
+
+    expect(gitParser.parseGitStatusNumstat(gitStatusNumstat)).to.eql({
+      "package-lock.json": { additions: "1459", deletions: "202" },
+      "package.json": { additions: "2", deletions: "1" },
+      "test/spec.git-parser.js": { additions: "13", deletions: "0" }
+    });
+  })
+});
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: source/git-parser.js [hunk 1]
```diff
@@ -74,7 +74,7 @@ const gitLogHeaders = {
   'Reflog': (currentCommmit, data) => {
     currentCommmit.reflogId = /\{(.*?)\}/.exec(data)[1];
     currentCommmit.reflogName = data.substring(0, data.indexOf(' ')).replace("refs/", "");
-    const author = data.substring(data.indexOf(' ') + 2, data.length - 1);
+    const author = data.substring(data.indexOf('(') + 1, data.length - 1);
```
ylecuyer: My reflog looks like this `Reflog: git-parser-specs@{Fri Jan 4 14:03:56 2019 +0100} (Test ungit)` and this line fails to find the first space as it is in the date

source/git-parser.js
```diff
@@ -291,7 +291,7 @@ const updatePatchHeader = (result, lastHeaderIndex, ignoredDiffCountTotal, ignor
 }
 
 exports.parsePatchDiffResult = (patchLineList, text) => {
-  if (!text) return {};
+  if (!text) return '';
```
ylecuyer: `parsePatchDiffResult` returns a string so if this error case, an empty string should be returned instead of an empty object for consistency

source/git-parser.js
jung-kim: What version of git are you using?

mine looks like `JK Kim <jung.kim.3627@gmail.com>`?

source/git-parser.js
jung-kim: Nice, I think we should either line 352 to `''` or this to `null`.

test/spec.git-parser.js
```diff
@@ -9,42 +8,383 @@ describe('git-parser stash show', () => {
...
-  describe('git-parser parseGitLog', () => {
-    it('should work with branch name with ()', () => {
-      const refs = gitParser.parseGitLog('commit AAA BBB (HEAD, (test), fw(4rw), 5), ((, ()')[0].refs;
+describe('git-parser parseDiffResult', () => {
+  it('all diff selected', () => {
+    var gitDiff = 'diff --git a/package.json b/package.json\n'
+    gitDiff += 'index f71e0064..08964575 100644\n'
+    gitDiff += '--- a/package.json\n'
+    gitDiff += '+++ b/package.json\n'
+    gitDiff += '@@ -87,9 +87,10 @@\n'
```
jung-kim: can we use multiline syntax with back ticks? it can be immutable and much cleaner.

test/spec.git-parser.js
```diff
@@ -9,42 +8,383 @@ describe('git-parser stash show', () => {
...
     expect(submodules.length).to.be(1);
-    expect(submodules[0].name).to.be('test1');
-    expect(submodules[0].path).to.be(path.join(path.sep, 'path', 'to', 'sub1'));
-    expect(submodules[0].url).to.be('http://example1.com');
+    expect(submodules[0]).to.eql({
+      name: "test1",
+      path: "/path/to/sub1",
+      rawUrl: "http://example1.com",
+      url: "http://example1.com"
+    });
```
jung-kim: 👍 I think this is a cleaner comparison!! I didn't know you could compare objects like this. I wonder if order matters or fully recursive...

test/spec.git-parser.js
```diff
@@ -9,42 +8,383 @@ describe('git-parser stash show', () => {
...
   });
 });
 
-describe('git-parse diff on big change', () => {
-  describe('git-parser parseGitLog', () => {
-    it('should work with branch name with ()', () => {
-      const refs = gitParser.parseGitLog('commit AAA BBB (HEAD, (test), fw(4rw), 5), ((, ()')[0].refs;
+describe('git-parser parseDiffResult', () => {
+  it('all diff selected', () => {
+    var gitDiff = 'diff --git a/package.json b/package.json\n'
```
jung-kim: can we stick with es6 `const` and `let`?

source/git-parser.js
ylecuyer: I have the following version:

```
ylecuyer@hawk:~$ git --version
git version 2.14.1
```
I have found the culprit, the command I used is :+1: 
```
git stash list --decorate=full --pretty=fuller --parents --numstat --date=default
```
notice the `--date=default` the git-api doesn't make the call with this argument.

Anyway, no need to revert, the change work for both :
```
Reflog: refs/stash@{0} (Test ungit <test@example.com>)
```

```
Reflog: refs/stash@{Wed Jan 9 21:31:35 2019 +0100} (Test ungit <test@example.com>)
```

source/git-parser.js
ylecuyer: Good catch :+1: I'll go for `null` here too then

test/spec.git-parser.js
ylecuyer: I didn't use multiline syntax because it was messing with indentation. I have just found `dedent` https://github.com/dmnd/dedent/blob/master/dedent.js which solves the issue :+1: Now it is easier to read and edit :)

test/spec.git-parser.js
ylecuyer: done

jung-kim:
Can we also fix the issue of window vs. linux path breaker problem?

Because right now test is permanently broken. You could use

```
const path = require('path');
path.sep
```

ylecuyer:
You're right, sorry for not having updated yet this obvious fail.

Issue #1173, opened 2019-01-05
Electron packaging errors


M0RGaming:
After running the grunt package command (and a bit of version editing in the grunt file), ungit compiles as a proper Darwin .app file. However, when running the ungit app, it then causes an issue stating that the module 'app' was not found:
![screen shot 2019-01-05 at 12 55 00 am](https://user-images.githubusercontent.com/13577339/50721927-897e6400-1084-11e9-976d-1ea4f602eedc.png)

I have no idea what could be causing this, but I do have a feeling that it may be the switch from async to BlueBird, as when building the 1.0 version, it works.

EDIT: After testing a version which did not contain BlueBird, I came to the conclusion that maybe this wasn't something caused by async. What I did realize is that only on the release 1.0 did I not have to manually install electron, and as such that may have something to do with this.

M0RGaming:
After a whole bunch of testing, I have found out the issue (in the main.js we need to change the var app and var browserwindow to require electron, I will be creating a pull request that includes my changes that allows ungit to work through electron)

PR #1177, opened 2019-01-20
Fix electron build

ylecuyer:
Related to https://github.com/FredrikNoren/ungit/issues/1173

Electron changed some apis https://electronjs.org/blog/electron-api-changes 
+ some update + cleaning


package.json
```diff
@@ -97,17 +98,5 @@
...
-    "toolbar": false,
-    "frame": false,
-    "show": false,
-    "width": 1000,
-    "height": 600,
-    "position": "center",
-    "min_width": 400,
-    "min_height": 200
-  }
+  "main": "electron/main.js"
```
jung-kim: I'm not sure about this line, I think it may point npm commands to electron folder and may result difficulty in debugging as execution will not be in the `public`

I will need to read up more and test this

Gruntfile.js
```diff
@@ -496,7 +471,7 @@ module.exports = (grunt) => {
   grunt.registerTask('publishminor', ['default', 'test', 'release:minor']);
 
   // Create electron package
-  grunt.registerTask('package', ['clean:electron', 'clean:babel', 'babel:electron', 'copy:electron', 'electron']);
+  grunt.registerTask('package', ['clean:electron', 'clean:babel', 'babel:electron', 'electron']);
```
jung-kim: We maybe conflating two PRs but we are doing babel for electron and removing babel in #1178.  

We want to be consistent either keep babel and keep wider support or just remove it.

ylecuyer: This PR is made to be merged before #1178 as it is simpler, in #1178 babel for electron is removed alongside babel for server.

package.json
ylecuyer: Good catch :+1: this point is tricky. Electron Application Architecture guide (https://electronjs.org/docs/tutorial/application-architecture) says that electron runs the script present at `main` inside the package.json

I see 2 possibilities : 
1. Keep the configuration as it is with an electron specific file hidden in `public/main.js` (previous configuration)
2. Make an electron folder to name things and mode `public/main.js` here to clearly states it is for electron (what I chose)

The rest of ungit always run the `./bin/ungit` binary and is not affected by this change

jung-kim: I haven't done the packaging node projects and I'm not too certain on this matter but I think previous option makes more sense as how we package our app should not affect how the app should be built.

campersau:
Superseded by https://github.com/FredrikNoren/ungit/pull/1240
I included one of your commits in my PR.

PR #1137, opened 2018-09-09
Bump get mac and allow animation disable

jung-kim:
- Bump getmac version [#1130](https://github.com/FredrikNoren/ungit/issues/1130)
- Add config to disable animation [#1136](https://github.com/FredrikNoren/ungit/issues/1136)
- Remove node6. Add node9 and node 10 from travis

(I know there is so much refactoring and very little fixes, but es6 has been something that has been bothering me for so long and this has been good exercise for me to revisit how the code works.)

PR #1178, opened 2019-01-20
Remove babel requirement for server

ylecuyer:
Remove babel requirement for server files


jung-kim:
I'm okay with removing babel but this means ungit will not work on older node or older browser as some of the client code within `/components` is in es6 now.

It would be nice to make a note of this in readme.

ylecuyer:
I may be wrong but the components assets are already served as es6. If I look at http://localhost:8448/plugins/app/app.bundle.js for example, I can see a fat arrow here : 

```
    this.repoList.subscribe((newValue) => { storage.setItem('repositories', JSON.stringify(newValue)); });
```

And the grunt file calls `browserify-components` directly inside `./components` which are not fed into babel (only source)

campersau:
It looks like @ylecuyer is right, the components where never fed into bable.
And since all tests passed on [node 8](https://travis-ci.org/FredrikNoren/ungit/jobs/484063609?utm_medium=notification&utm_source=github_status) I think it is good.

I know its been a long time but do you mind rebasing this PR or creating a new one?
Otherwise I can do that for you.

Hirse:
Proposal:
Keep babel, but use it with [preset-env](https://babeljs.io/docs/en/babel-preset-env) for both server and components with different targets (and possibly "compile" shared files twice).

Side note, what are the supported browsers? I am getting syntax errors in IE11 from the fat arrows.

campersau:
Its been over a year since https://github.com/FredrikNoren/ungit/pull/1137 got merged and we never got any issues regarding that.
So I would vote for just supporting evergreen browsers and make the build process a bit simpler :)

campersau:
Thanks @ylecuyer !

---

If we still want to support older browsers we could use https://github.com/babel/babelify for example.

It is a bit difficult because [`octicons` switched to a newer syntax](https://github.com/primer/octicons/commit/875bc333ce4dc6a460a116dfd315a56dbb23120f) and [browserify ignores `npm_modules` by default](https://stackoverflow.com/a/39349302) and we use [`Array.prototype.includes`, `String.prototype.includes`](https://github.com/FredrikNoren/ungit/search?q=includes&type=Code) for which we would need a polyfill.

Issue #1170, opened 2019-01-01
Public Analytics


ylecuyer:
Anlytics are sent to Keen.io, Google Analytics, Sentry but there is no public dashboard.

Is there a way to have a public dashboard ? or read access to those services ?

cc @FredrikNoren

ylecuyer:
ping @FredrikNoren

jung-kim:
So there is a dashboard and it is now broken due to unpaid keen.io.  Other data points I don't know about

PR #1180, opened 2019-01-23
Remove Keen.io

ylecuyer:
Related to #1170 

Keen.io is now useless and can be removed


source/server.js
```diff
@@ -22,8 +21,7 @@ const Bluebird = require('bluebird');
 process.on('uncaughtException', (err) => {
   winston.error(err.stack ? err.stack.toString() : err.toString());
   Bluebird.all([
-    new Bluebird((resolve) => { bugtracker.notify.bind(bugtracker, err, 'ungit-launcher'); resolve(); }),
-    new Bluebird((resolve) => { usageStatistics.addEvent.bind(usageStatistics, 'launcher-exception'); resolve(); })
+    new Bluebird((resolve) => { bugtracker.notify.bind(bugtracker, err, 'ungit-launcher'); resolve(); })
```
jung-kim: if it's just one promise, there is no reason to use `Bluebird.all()`

public/main.js
```diff
@@ -13,8 +12,7 @@ var BrowserWindow = require('browser-window');  // Module to create native brows
 process.on('uncaughtException', function(err) {
   console.error(err.stack.toString());
   Bluebird.all([
-    new Bluebird((resolve) => { bugtracker.notify.bind(bugtracker, err, 'ungit-launcher'); resolve(); }),
-    new Bluebird((resolve) => { usageStatistics.addEvent.bind(usageStatistics, 'launcher-exception'); resolve(); })
+    new Bluebird((resolve) => { bugtracker.notify.bind(bugtracker, err, 'ungit-launcher'); resolve(); })
```
jung-kim: if it's just one promise, there is no reason to use Bluebird.all()

PR #1183, opened 2019-01-26
Better gitGetVersion

ylecuyer:
Check the required version from params instead of hardcoded version


jung-kim:
I'm confused to how this is any better because what you are passing in isn't really a param...?  it is a param created from a const in front end.  

Why or how is this any better?

Issue #1190, opened 2019-03-23
broken links to octicons!


avloss:
https://github.com/FredrikNoren/ungit/blob/e9738bae82730ec0c7b3c48ddc1e3fd9f7055939/Gruntfile.js#L186

icons are broken in `datalab`, which is not a big issue for those who know what's going on but could be confusing for beginners (the target audience of this tool)
<img width="1397" alt="Screenshot 2019-03-23 at 19 29 57" src="https://user-images.githubusercontent.com/17853685/54870717-15f61400-4da2-11e9-931f-710467423fb5.png">

links they are broken even in the official google tutorial!
https://cloud.google.com/datalab/docs/how-to/working-with-notebooks

(this is due to some proxying magic which datalabs do). Obviously, this has NOTHING to do with ungit, and you guys are doing everything right, BUT.. could you please consider importing that library from some CDN? I mean it's just a static file, but would look so much better! :) 

Currently, it's looking inside
`http://localhost:8081/css/octicons.ttf`
But for this to work it should be looking here
`http://localhost:8081/_proxy/8083/css/octicons.ttf`

So, not sure how hard is this to set up with all the `npm` packaging, but would be great to have icons fixed. I'm considering introducing this great tool to our team, and this small thing really makes it very confusing for people who aren't familiar with git!

Hirse:
With the PR merged can this be closed?

Sidenote, would being included in Google Cloud be something to brag about in the Readme? 😉

jung-kim:
Ah... hooray!!!!  I had no idea this was happening and it is cool.

I'm all for adding in little excerpts about this

PR #1191, opened 2019-03-24
fixing broken git icons

avloss:
fixing broken git icons, which occurs when ungit is served from proxy with url rewrite (like datalab does).

#1190 is what I'm trying to address here!


jung-kim:
Thanks!

Since this is a small change, I'm not going to worry about bumping up the version and releasing.

Issue #1196, opened 2019-05-02
[Feature Request] Branch search box


ZimGil:
It would be nice to have a search box at the branches list so you can easily find the branch you're looking for in a big project where it can get really annoying to find it by scrolling on 2 axis.

jung-kim:
![Screen Shot 2019-05-02 at 19 14 57](https://user-images.githubusercontent.com/5281068/57117133-b8dd6e80-6d0e-11e9-9d9b-6f870d7245f7.png)

Like some of the features, it is not documented so few people know but there is a drop down of all branches and tags like above.  And you can checkout by click the row, or delete by clicking `x`.  

It is not exactly a "search" but would this be suffice?

ZimGil:
I knew about this but I work on an enormous repo and hard to find sometimes.

I guess it would be nice to have a search box on the top of this list that will filter out whichever branch isn't falling under the search string.

jung-kim:
I can add ref search box within the branch drop down. I think that would be sufficient right?

ZimGil:
Sounds good! 👍

ZimGil:
@jung-kim Can I give it a try?

jung-kim:
Oh shucks, I will try to deliver soon but it's not quite ready yet... But will let you know when it is ready

ZimGil:
I meant to create it :)

jung-kim:
Oh I see, absolutely!  

There is already a similar search box if you are looking for an example, [backend code](https://github.com/FredrikNoren/ungit/blob/master/components/graph/git-node.js#L169-L195) and [frontend code](https://github.com/FredrikNoren/ungit/blob/master/components/graph/graph.html#L60-L63).  But it is at node level and scope is only the refs in that node.  (I may hide this by default and enable this node level ref search to be enabled only when a node has more than N number of refs as it is bit silly in most cases).

Also, I must say code on this is bit sloppy and if you have a cleaner way I would be all for it

jung-kim:
I will be going on very long vacation soon so I might be very limited in availability but feel free to ask me any questions about how to build and etc.

ylecuyer:
Meanwhile you're on vacation I can help too and start the review

ZimGil:
Sorry but I had to close this PR as I'm too busy right now to fix it

JackNeus:
Is anyone working on this? Would love to give it a shot. @jung-kim

PR #1200, opened 2019-05-18
[WIP] Feature/branch search box

ZimGil:
I was able to make the search box operate as expected in #1196 

Some bugs that I would like to get some help with:
 * ~~After choosing a branch for checkout, the search string stays but the branches are added again into the list.~~
 * ~~I still iterate twice over some branches (once for all the available and 2nd time over the ones that was searched)~~
 * I could not figure out how to change the color of the `placeholder` from white, now it looks like an empty search box until user input.
 * I could not find a way to avoid iterating over the available branches (`branchesAndLocalTags`) in order to fill the `filteredBranchesAndLocalTags` when the `searchString` is empty.
    > Tried using: `this.filteredBranchesAndLocalTags(this.branchesAndLocalTags());` but then the search got stuck.
    > All other methods (that I've found) to fill an observable array do it by iterating.

I've never worked with Knockout.js so I would like to see where I can make this better.
@jung-kim @ylecuyer

### DIFF
--- a/components/branches/branches.html
+++ b/components/branches/branches.html
@@ -21,12 +21,10 @@
         <input type="checkbox" data-bind="checked: isShowTag"/>
       </label>
     </div>
-
     <div class="dropdown-divider"></div>
     <div class="form-inline" onclick="event.stopPropagation()">
       <input class="form-control branch-search-box" type="text" placeholder="Title (required)" data-bind="value: branchSearchString, valueUpdate: 'afterkeydown'"/>
     </div>
-
     <div class="dropdown-divider"></div>
     <!-- ko foreach: filteredBranchesAndLocalTags -->
     <li>
--- a/components/branches/branches.html
+++ b/components/branches/branches.html
@@ -21,8 +21,14 @@
         <input type="checkbox" data-bind="checked: isShowTag"/>
       </label>
     </div>
+
     <div class="dropdown-divider"></div>
-    <!-- ko foreach: branchesAndLocalTags -->
+    <div class="form-inline" onclick="event.stopPropagation()">
+      <input class="form-control branch-search-box" type="text" placeholder="Title (required)" data-bind="value: branchSearchString, valueUpdate: 'afterkeydown'"/>
+    </div>
+
+    <div class="dropdown-divider"></div>
+    <!-- ko foreach: filteredBranchesAndLocalTags -->
     <li>
       <a href="#" data-bind="html: displayName, click: $parent.checkoutBranch.bind($parent), attr: { 'data-ta-clickable': 'checkout' + name }"></a>
       <a href="#" class="list-link list-remove" data-bind="click: $parent.branchRemove.bind($parent), attr: { 'data-ta-clickable': name + '-remove' }">X</a>
--- a/components/branches/branches.html
+++ b/components/branches/branches.html
@@ -23,7 +23,7 @@
     </div>
     <div class="dropdown-divider"></div>
     <div class="form-inline" onclick="event.stopPropagation()">
-      <input class="form-control branch-search-box" type="text" placeholder="Title (required)" data-bind="value: branchSearchString, valueUpdate: 'afterkeydown'"/>
+      <input class="form-control branch-search-box" type="text" placeholder="Search Branchs" data-bind="value: branchSearchString, valueUpdate: 'afterkeydown'"/>
     </div>
     <div class="dropdown-divider"></div>
     <!-- ko foreach: filteredBranchesAndLocalTags -->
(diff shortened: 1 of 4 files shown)


components/branches/branches.js
```diff
@@ -17,6 +17,11 @@ class BranchesViewModel {
     this.repoPath = repoPath;
     this.server = server;
     this.branchesAndLocalTags = ko.observableArray();
+    this.filteredBranchesAndLocalTags = ko.observableArray();
+    this.branchesAndLocalTags.subscribe((branches) => {
```
ZimGil: > * After choosing a branch for checkout, the search string stays but the branches are added again into the list.

Oh now I see that its actually happening on every `git pull`

jung-kim:
- Ermmm, I'm not sure on this one either myself and front-end is a never ending mystery for me.  but I thought placeholder will inherit text color no?
- I see what you are trying to do here, and yes you would have to iterate over all branches everytime search string changes.  Infact, if you see the examples I have shared with you before this is all done via [jquery autocomplete](https://jqueryui.com/autocomplete/) library.  Is there a particular reason why we are not using this lib

ylecuyer:
I haven't read the code yet but tested the feature instead:

1. As you said in the description it is all white, white background, white input, white placeholder, maybe you can add a subtle grey as a background and replace the place holder with a search icon on the right maybe?
2. the debounce doesn't feel ok. I understand it may be needed if there is a lot of branches but for me it looked awkward. For instance, with the ungit repo, if you search for `0/` you can have results though there is no match
3. I tried without debounce and to fully clear the search, you have to press the back key twice. eg: search 0.1 then back / back one more time back and you wont have the full list, back once again and you have the full list.
4. Instead of populating the list of branches and remotes, maybe you can add a flag isHidden to each branch/remote ? it will speedup the process and avoid duplication

components/branches/branches.html
```diff
@@ -22,7 +22,11 @@
       </label>
     </div>
     <div class="dropdown-divider"></div>
-    <!-- ko foreach: branchesAndLocalTags -->
+    <div class="form-inline" onclick="event.stopPropagation()">
+      <input class="form-control branch-search-box" type="text" placeholder="Search Branchs" data-bind="value: branchSearchString, valueUpdate: 'afterkeydown'"/>
```
ylecuyer:
```suggestion
      <input class="form-control branch-search-box" type="text" placeholder="Search Branchs" data-bind="value: branchSearchString, valueUpdate: 'afterkeyup'"/>
```

should fix my comment 3.

ZimGil:
* I don't really like the auto complete solution, you'll end-up having the full list and the auto complete coveting it in some way. I like the idea of having an `isHidden` flag on the array items much better.

* I couldn't see what was the problem @ylecuyer ran into with `_.debounce`, I will change to `afterkeyup` and try to see if there's any problem.

ZimGil:
Sorry for the delay had a busy few weeks,

I was able to fix the placeholder and the debounce looks fine after working with this for a few days.

I've tried to add an `isVisible` flag to the array items but I could not remove it from the list.

I've tried add either `ref.isVisialbe = true` or `ref.isVisible = ko.observable(true)`
and change it either by `isVisible = false` or `isVisible(false)` and in both cases I saw the change but it kept on showing on the list when I had `data-bind="if: isVisible` on the `li`

ZimGil:
After working with this for a while I find the ungit process overloads the CPU from time to time

jung-kim:
> I don't really like the auto complete solution, you'll end-up having the full list and the auto complete coveting it in some way. I like the idea of having an isHidden flag on the array items much better.

But your solution isn't using `isHidden`, you basically have copy of the array which you are against...?  I would much rather use already implemented library if possible as this would be reinventing the wheels and lacks some of the features, such as fuzzy matching or case insensitive matching.

> After working with this for a while I find the ungit process overloads the CPU from time to time

is this due to before or after this feature?

ZimGil:
> But your solution isn't using `isHidden`, you basically have copy of the array which you are against...? I would much rather use already implemented library if possible as this would be reinventing the wheels and lacks some of the features, such as fuzzy matching or case insensitive matching.

As mentioned in my [previous comment](https://github.com/FredrikNoren/ungit/pull/1200#issuecomment-498695442), I've tried to go with the visibility flag but could not get it to work well...
I'm not sure I was doing it the correct way.

> > After working with this for a while I find the ungit process overloads the CPU from time to time
> 
> is this due to before or after this feature?

On this branch

jung-kim:
Sorry I have been busy but every single comments adds further reasons to use auto complete library

This PR utilizes the very reason why you didn't wanted to use auto complete library because initial approach didn't work and it even overloads CPU.

ZimGil:
I'm sorry but I won't have the time to implement this as I'm very busy for the next couple of weeks...

PR #1205, opened 2019-07-08
Gitignore Dialog style

Hirse:
Improves the textarea of the gitignore Dialog:
* Use the existing `form-control` visual design
* Use monospace font
* Remove spellchecking
* Only allow vertical resizing to prevent growing outside of the dialog

Before:
image

After:
image


jung-kim:
Woo! lovely, can you also fix up `CHANGELOG.md`, `package.json` and run `npm i` so `package-json` is also updated?

Thanks!

PR #1206, opened 2019-07-08
Remove deprecated os-homedir

Hirse:
See https://www.npmjs.com/package/os-homedir


jung-kim:
Ah... It might have been better to add this to last PR but can you help us resolve this conflicts??

package.json doesn't really bodes well with ungit's deploy strategy... :(

Issue #1187, opened 2019-03-08
ungit should always run git commands in english


yukulele:
If system is not configured in english:
- Ungit show wrong error message
- it sometimes displays an error when there is none (for example when changing branches)

(I suppose he parses git's answers to check that the process went well)

ungit should always run git commands in english

workaround: run ungit itself in english: 
```console
LANG=en_US ungit
```

jung-kim:
https://github.com/FredrikNoren/ungit/pull/963

So I've tried it years ago but it didn't work out for me.  But feel free to have a a crack at it.  

[gitConfigArgument](https://github.com/FredrikNoren/ungit/blob/master/source/git-promise.js#L11) is an argument that get appended to all git commands.

campersau:
Fixed by https://github.com/FredrikNoren/ungit/pull/1208#issuecomment-511027300

PR #1208, opened 2019-07-10
force LC_ALL=C when running git

bcroq:
Ungit parses git output expecting english.

My default locales configuration is not english so I get a lot of warnings, e.g. when moving a branch.

This PR forces LC_ALL=C before running git commands so the output is always in english.

### DIFF
--- a/source/git-promise.js
+++ b/source/git-promise.js
@@ -39,7 +39,9 @@ const gitExecutorProm = (args, retryCount) => {
       let isCompleted = false;
       let stdout = '';
       let stderr = '';
-      const procOpts = { cwd: args.repoPath, maxBuffer: 1024 * 1024 * 100, detached: false }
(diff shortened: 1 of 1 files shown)


jung-kim:
Thanks for this!

Is there a strong argument for or against each git command's inheriting all of node processors environment variables?  I'm especially concerned around security aspect but I can't think of any but just want to spit ball and make sure everybody is happy.

bcroq:
By default (ie. without this patch) each git command inherits from node's environment variables.

jung-kim:
If that is true, is there any reason we are explicitly setting child process's env to parent's env?

bcroq:
Without this PR, there is no `env` option in `procOpts` so by default the child process get its parent's env.

Hirse:
Does this fix #1187?

bcroq:
Indeed, it fixes #1187.

PR #1212, opened 2019-07-25
correct 'src' to 'source' as per pkg dir structure

HarvsG:
It appears as though lines in `bin/ungit` try to require contents ins `src/` which doesn't exist, however there are contents in `source/`


campersau:
The contents in the `src` directory is generated via these grunt tasks:

https://github.com/FredrikNoren/ungit/blob/e234722d7b4c127623f34a3833ddf07578b4a9c4/Gruntfile.js#L243-L268

So everything is fine as it is 👍

PR #1215, opened 2019-08-05
Prevent diff linenumber selection

### LINKED ISSUE #1214: Prevent diff2html line number and prefix selection
I often want to quickly copy and paste text from the diffs shown by ungit, but it is frustrating because the line number and and +/- prefixes are copied too, along with awkward line breaks.  It would be very helpful if these elements could be marked unselectable.

simonwiles:
Applies `user-select: none;` to `.d2h-code-line-prefix` and `.d2h-code-linenumber` classes.  This allows selection text from diffs for copy and paste without getting line numbers and +/- prefixes.

Closes #1214

### DIFF
--- a/public/less/d2h.less
+++ b/public/less/d2h.less
@@ -173,6 +173,7 @@
   text-overflow: ellipsis;
   border-left: 1px solid rgba(0, 0, 0, 0.34);
   border-right: 1px solid rgba(0, 0, 0, 0.34);
+  user-select: none;
 }
 
 .d2h-del {
--- a/public/less/d2h.less
+++ b/public/less/d2h.less
@@ -119,6 +119,7 @@
 .d2h-code-line-prefix {
   display: inline-flex;
   margin-right: 10px;
+  user-select: none;
 }
 
 .line-num1 {
@@ -150,6 +151,7 @@
   text-overflow: ellipsis;
   border-left: 1px solid rgba(0, 0, 0, 0.34);
   border-right: 1px solid rgba(0, 0, 0, 0.34);
+  user-select: none;
 }
 
 .d2h-cntx {
--- a/public/less/d2h.less
+++ b/public/less/d2h.less
@@ -119,6 +119,9 @@
 .d2h-code-line-prefix {
   display: inline-flex;
   margin-right: 10px;
+  -moz-user-select: none;
+  -ms-user-select: none;
+  -webkit-user-select: none;
   user-select: none;
 }
 
@@ -151,6 +154,9 @@
   text-overflow: ellipsis;
   border-left: 1px solid rgba(0, 0, 0, 0.34);
   border-right: 1px solid rgba(0, 0, 0, 0.34);
+  -moz-user-select: none;
+  -ms-user-select: none;
+  -webkit-user-select: none;
   user-select: none;
 }
 
@@ -173,6 +179,9 @@
   text-overflow: ellipsis;
   border-left: 1px solid rgba(0, 0, 0, 0.34);
   border-right: 1px solid rgba(0, 0, 0, 0.34);
+  -moz-user-select: none;
+  -ms-user-select: none;
+  -webkit-user-select: none;
   user-select: none;
 }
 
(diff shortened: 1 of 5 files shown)


public/less/d2h.less
```diff
@@ -119,6 +119,7 @@
 .d2h-code-line-prefix {
   display: inline-flex;
   margin-right: 10px;
+  user-select: none;
```
campersau: This will disable the selection of the whole line (e.g. with the content). Is this intended? I guess for deletions this might be correct but not for changes / additions.

simonwiles: I don't think it will -- it should just disable selection of the +/- prefix (i.e., the element with the class `d2h-code-line-prefix`).

campersau: For me this isn't working. Here you can see that none of the changed lines is now selected:
image

Because the `d2h-code-line-prefix` also includes the `d2h-code-line-ctn` which inherits the `user-select: none;`:
image


Can you also please fix the side-by-side diff? It uses the `d2h-code-side-linenumber` class:
image

---

On GitHub you can also select addition, modifications and deletions so I think that's fine then.

simonwiles: I'm puzzled -- you appear to be looking at something rather different to what I have:
image
--
image

> Because the `d2h-code-line-prefix` also includes the `d2h-code-line-ctn`

If this were the case (and again, I can't reproduce this), that would be a problem anyway, right?

> Can you also please fix the side-by-side diff?

Yes, thanks for catching this!

> On GitHub you can also select addition, modifications and deletions so I think that's fine then.

I agree :)

I've also added all the prefixed versions of the `user-select` property, as I had mistakenly assumed the LESS preprocessor would do this.

campersau: Ahh I was looking at the diffs in the staging area, they are rendered differently because of the `patch` feature I guess.

https://github.com/FredrikNoren/ungit/blob/e234722d7b4c127623f34a3833ddf07578b4a9c4/components/textdiff/textdiff.js#L191-L196

Can you please add the missing `</div>` in this line? Then it should also work there.

```diff
diff --git a/components/textdiff/textdiff.js b/components/textdiff/textdiff.js
index 9698fe1e..bfe11ce2 100644
--- a/components/textdiff/textdiff.js
+++ b/components/textdiff/textdiff.js
@@ -192,7 +192,7 @@ class TextDiffViewModel {
     if (isActive) {
       this.numberOfSelectedPatchLines++;
     }
-    return `<div class="d2h-code-line-prefix"><span data-bind="visible: editState() !== 'patched'">${symbol}</span><input ${isActive ? 'checked' : ''} type="checkbox" data-bind="visible: editState() === 'patched', click: togglePatchLine.bind($data, ${index})"></input>`;
+    return `<div class="d2h-code-line-prefix"><span data-bind="visible: editState() !== 'patched'">${symbol}</span><input ${isActive ? 'checked' : ''} type="checkbox" data-bind="visible: editState() === 'patched', click: togglePatchLine.bind($data, ${index})"></input></div>`;
   }

   togglePatchLine(index) {
```

---

The commit diffs look good though 👍

jung-kim: I feel like patch feature causes more confusions and problems and not sure if it is worth keeping it...

campersau: @simonwiles any update on this? If you want I can make the [last change](https://github.com/FredrikNoren/ungit/pull/1215/commits/29fff80480280433b31a951aba81c657d1a22a30#r311863078) so we can merge the PR.

simonwiles: @campersau apologies, I got distracted because I noticed that this doesn't work perfectly in Firefox, and I couldn't find a solution (and then I got busy with other things).

In Firefox a lot of unnecessary linebreaks are copied when selecting multiple lines from the diffs.  I experimented with a number of different solutions, but I couldn't get a perfect behaviour in FF, and I still haven't got a solution.  However, the behaviour effected in this PR is good in Chrom[e|ium], Opera, Brave etc., and better than it was previously in FF, so in the spirit of not letting the perfect be the enemy of the good, I've made the requested change, and I think it might be ready to merge.

I assume the failing tests are not related to this PR?

campersau: Thanks!

jung-kim: Test is breaking something completely irrelevant and I will try to fix it

PR #1221, opened 2019-10-12
Fix the width value of the header logo

ParkSB:
This PR fixes the header logo width to a constant value.

The width value of the logo image is `15%` now, and it extends the width value of the parent anchor tag (which navigates to the homepage) too big. Eventually, the transparent clickable area is created. 

AS-IS

I fixed the width to 100px to limit the clickable area to the displayed image.

TO-BE


jung-kim:
thanks!!

PR #1225, opened 2019-10-22
Hide diff-buttons on commit hover

Hirse:
* Show the diff buttons only when a commit is selected, but not on hover
* Remove empty space when no commit message body is present
* Align margins in the commit-"card"
* Clean up lint issues in changed files

*Before*
image

*After*
image

### DIFF
--- a/components/commit/commit.html
+++ b/components/commit/commit.html
@@ -29,7 +29,7 @@
       <!-- ko if: selected() || nodeIsMousehover() -->
       <div class="details">
         <div class="body" data-bind="visible: title().length > 72, text: '...' + title().substring(72)"></div>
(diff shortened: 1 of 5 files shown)


campersau:
While validating these changes I have noticed that there are other inconsitencies as well. Maybe not related to the PR but if you are working on it maybe you can fix them as well. Otherwise we can create separate issues for them.

The order and state of the buttons are inconsistent:
image

Clicking on the `No Wrap` button also toggles the visual state of the `Show/Ignore white space diff` button:
image

Hirse:
@campersau I have noticed some of that too, but don't think they are related.
I can change it in a new PR.

btw, do you know why the tests in Travis seem to consistently fail (even before my PRs)?

PR #1228, opened 2019-10-25
Improve git-init experience

Hirse:
* Invalid Path
  * Use `container` to center the content
* No repository
  image
  image
  * Use "git repository" instead of just "repository" to be more precise
  * Fix path display string on windows
  * Unify panel header and button texts
  * Fix checkbox design
* Empty repository
  image
  image
  * Remove divider when list of remotes is empty
  * Disable branch dropdown when list of refs is empty
  * Use default label when there are no commits

For a better view of the changes (without the indentation) use https://github.com/FredrikNoren/ungit/pull/1228/files?w=1

### DIFF
--- a/components/path/path.js
+++ b/components/path/path.js
@@ -1,4 +1,3 @@
-
 const ko = require('knockout');
 const components = require('ungit-components');
 const addressParser = require('ungit-address-parser');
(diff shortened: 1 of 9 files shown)


components/path/path.js
```diff
@@ -13,7 +12,7 @@ class PathViewModel {
   constructor(server, path) {
     this.server = server;
     this.repoPath = ko.observable(path);
-    this.dirName = this.repoPath().replace('\\', '/')
+    this.dirName = this.repoPath().replace(/\\/g, '/')
```
campersau: Thanks! This looked wired on windows before 👍

PR #675, opened 2016-01-16
define g++ spec for npm warnings.

codingtwinky:
Seems like build has been passing even with npm install step in travis was having [many warnings and some errors](https://travis-ci.org/FredrikNoren/ungit/jobs/101333367)...

https://github.com/travis-ci/travis-ci/issues/4771

PR #1231, opened 2019-10-29
Add libgconf-2-4 package for nightmare on travis

campersau:
remove old g++ workarounds (looks like they aren't needed. Introduced in  https://github.com/FredrikNoren/ungit/pull/675)

Looks like `libgconf-2-4` is required for running nightmare (electron) clicktests. See https://github.com/electron/electron/issues/1518
Not sure why this wasn't an issue before.

This just took me [16 tries](https://travis-ci.org/campersau/ungit/builds/604400898) 😄

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ branches:
   only:
     - master
 env:
-  global:
-    - CXX=g++-4.8
   matrix:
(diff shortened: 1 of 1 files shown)


Hirse:
Great job finding the fix for the build. 👍

PR #1234, opened 2019-10-30
remove electron dependency

campersau:
@jung-kim do you know why you had to add it in https://github.com/FredrikNoren/ungit/commit/cd0a12dc8c0ccdc80455d5c95cb2d14834d519cd ?
Maybe it was an intermediate fix for #1231 ?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -35,12 +35,6 @@
       "dev": true,
       "optional": true
     },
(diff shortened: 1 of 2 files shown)


campersau:
Ah never mind it is used in the electron packaged app, but it seems to be broken.

Issue #1222, opened 2019-10-18
Is there a compiled version of ungit?


adelriosantiago:
Is there a way to obtain a compiled version of it? I work under a proxy that won't allow to `npm install` most of the packages.

Thanks.

campersau:
No, there currently aren't any compiled versions. But maybe we can add them, see https://github.com/FredrikNoren/ungit/issues/1218#issuecomment-548980020

campersau:
We are now producing electron builds which don't require `node` or `npm`. They just need `git` to be installed. You can download them from here https://github.com/FredrikNoren/ungit/releases

Issue #1218, opened 2019-09-18
Actual releases?


gerroon:
Hi 

Are there plans to make actual actual releases for platforms for those who are not familiar with node js stuff?

thanks

campersau:
Yes, maybe we can start doing that again. We could even create release tags automatically during the build like the npm publishing .
Then we could fix our electron packaging and publish the resulting artifacts as github release via travis.
Which would also fix https://github.com/FredrikNoren/ungit/issues/1222

I have prepared something like that over here https://github.com/campersau/ungit/tree/testing
And here are example builds/releases with the electron artifacts:
- https://github.com/campersau/ungit/releases/tag/release-1.4.48
- https://ci.appveyor.com/project/campersau/ungit/builds/28557864/job/w5d4v4bci3i2ai6e/artifacts

@jung-kim what do you think?

gerroon:
Thank you, I am testing it now.

campersau:
We are now producing electron builds which don't require `node` or `npm`. They just need `git` to be installed. You can download them from here https://github.com/FredrikNoren/ungit/releases

PR #1240, opened 2019-11-03
Fix electron package

campersau:
use latest electron packager and build them in CI.

fixes https://github.com/FredrikNoren/ungit/issues/593 for the second time :)

Issue #687, opened 2016-02-01
Changelog/Release Notes


daedalus28:
Congrats and excellent work releasing 0.10! I rushed to update as soon as I saw the email.

It would be really nice to have a list of things that have changed each version moving forward - as it is now I am just looking around in the UI for new features and reading all the commit messages since the last version. Any chance we could get a changelog.md or something similar?

FredrikNoren:
Happy to hear :) And yeah agree, a changelog is much needed.

I think what we should do is that we keep a `CHANGELOG.md` file and any PR that introduces new features also needs to add that feature to the changelog. What do you think @codingtwinky ?

jamesjnadeau:
:+1: on the changelog! 

By the way, 
great update and work :clap: , looks fantastic! 
Love this thing! Makes git easier for everyone.

codingtwinky:
Hooray!

I just saw release note are missing [within gihub](https://github.com/blog/1547-release-your-software)/  I thought release notes that was gathered were added in there as well.   or do you guys prefer `CHANGELOG.md`?

codingtwinky:
Or it could be on gh-page or wiki along with #636?  

As a code cruncher I really hate writing human language and it would be nice to have that going...

daedalus28:
`CHANGELOG.md` is probably the most straightforward to manage from the code. If it was in the wiki or gh-pages, wouldn't that mean it would have to be updated independently from the source code? In my experience that usually means it will just never get updated.

FredrikNoren:
Yeah in source code seems like the easiest

codingtwinky:
Do we have any quick sugestions on format of change log?   My very tired brain is imagining something like below.  Just an example...

```
### 0.10.2
feature:
* #.... : removed hover effect #688

bug fix:
* #692: fix duplicate remote tage bug #686 
```

Ajedi32:
@codingtwinky I personally like the format suggested by the [Keep a CHANGELOG](http://keepachangelog.com/) project: https://raw.githubusercontent.com/olivierlacan/keep-a-changelog/master/CHANGELOG.md

FredrikNoren:
@Ajedi32 nice hadn't seen that before. Seems easiest to use that, I'll change the pr

FredrikNoren:
@codingtwinky I like the idea of referencing issues & pr's in the change log too, I'll include that

PR #1241, opened 2019-11-03
Electron publish

campersau:
based on #1240 this will create a `dist` folder which contains all zipped builds from electron.
These zips are uploaded as artifacts on AppVeyor for every build and can be release by Travis if a commit in the `master` branch is tagged.

Maybe we want to automatically create tags during builds on master which then starts another build which creates the actual release.

fixes https://github.com/FredrikNoren/ungit/issues/1218
fixes https://github.com/FredrikNoren/ungit/issues/1222

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,3 +39,15 @@ before_script:
   - DISPLAY= npm run package
 after_success:
   - npm run travisnpmpublish
+before_deploy:
+  - npm run electronpublish
+deploy:
+  provider: releases
+  api_key: ${GITHUB_TOKEN}
+  file_glob: true
+  file: dist/*.zip
+  overwrite: true
+  skip_cleanup: true
+  on:
+    tags: true
+    condition: $GIT_VERSION = edge
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ node_js:
 branches:
   only:
     - master
+    - /^v\d+\.\d+\.\d+$/
 env:
   matrix:
     - GIT_VERSION=edge
@@ -48,6 +49,7 @@ deploy:
   file: dist/*.zip
   overwrite: true
   skip_cleanup: true
+  body: "[Changelog](https://github.com/FredrikNoren/ungit/blob/master/CHANGELOG.md#${TRAVIS_TAG//[v\\.]})"
   on:
     tags: true
     condition: $GIT_VERSION = edge
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,200 +1,541 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
-Use the following format for additions: ` - VERSION: [feature/patch (if applicable)] Short description of change. Links to relevant issues/PRs.`
+We are following the [Keep a Changelog](https://keepachangelog.com/) format.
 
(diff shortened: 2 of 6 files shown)


Hirse:
Is this deploying on every build or only when the version has increased?

campersau:
@Hirse both. The electron packages will be available on every build as artifacts on AppVeyor (they keep them for 6 months).
If a commit gets tagged in master then this would produce a GitHub release with all electron packages as artifacts.
See https://github.com/FredrikNoren/ungit/issues/1218#issuecomment-548980020 for examples.

campersau:
In the second commit I also enabled automatic version tagging.
Here are the different scenarios:

- PR **with** or **without** `package.json` version change
  1. Build runs on travis and appveyor.
  2. Electron packages are uploaded as artifacts on appveyor.

- Commit to master (directly or PR merge) **without** `package.json` version change
  1. Build runs on travis and appveyor.
  2. Electron packages are uploaded as artifacts on appveyor.
  3. Travis tries to publish npm package and fails.

- Commit to master (directly or PR merge) **with** `package.json` version change
  1. Build runs on travis and appveyor.
  2. Electron packages are uploaded as artifacts on appveyor.
  3. Travis publishes npm package.
  4. Travis creates a new tag with the updated version number,
  5. The new tag triggers new builds on travis and appveyor.
  6. Electron packages are uploaded as artifacts on appveyor.
  7. Travis deploys a new release on github including electron packages.

Some notes:
- The deployment of the release and the tag creation are currently using my (`campersau`) GitHub token. So I guess the releases would always be release with my user.
- We could provide a bit more details in the release description (body). Maybe we can link to the [CHANGELOG.md](https://github.com/FredrikNoren/ungit/blob/master/CHANGELOG.md)?

Example: https://github.com/campersau/ungit/releases/tag/1.4.48

cc @jung-kim @FredrikNoren

Hirse:
The current CHANGELOG would only allow linking to the whole file.
For linking to specific versions, which would be nicer, the format would have to be slightly changed. I would propose adopting this one: https://keepachangelog.com

campersau:
@Hirse Yes, that's how it was done before the automatic version updates were introduced.
See https://github.com/FredrikNoren/ungit/issues/687#issuecomment-178632834

I have updated the change log to use the https://keepachangelog.com format and also applied all missing tags. (I used `git log -m --first-parent -L5,5:package.json` to get all the commits and wrote a program for reading the package json on each of the commits and tagging it.)

Hirse:
@campersau I noticed the old format after writing my comment. Looks good after transforming it, though some changes might fit better under a "Changed" headline rather than "Fixed".

campersau:
@Hirse Yeah, maybe. I just selected all and went with "Fixed" as default and pnly changed it to "Added" sometimes.

campersau:
There is also [electron-builder](https://github.com/electron-userland/electron-builder) which can build e.g. portable executables with auto update functionality. Which would make this scenario even easier. But let's see first how much demand there is for that after this PR is merged or should we move now?

(Example project which uses it: https://github.com/mifi/lossless-cut/releases)

campersau:
The macOS electron packages created by appveyor (windows) for each build don't work because of serveral symlink issues on windows. (https://github.com/electron/electron-packager/pull/984#issuecomment-556994986)
The release packages which are build on travis (linux) do work.

Issue #1204, opened 2019-06-05
Can't run ungit, async path unexpected identifier


NathanOrme:
Running on node v6.16.0,npm v3. 10.10, ungit version 1.4.44.

Get error on line 11 of index.js: const wslToWindowsPath = async path... SyntaxError: Unexpected Identifier on path. 
![15597326900156399218011382706702](https://user-images.githubusercontent.com/18661760/58951890-33743080-878a-11e9-82f7-cb0a08146867.jpg)

ZimGil:
Is this on Linux?

NathanOrme:
It is @zimgil

ZimGil:
Than it's a dup of #1203 
I will comment there...

ylecuyer:
IMO readme should be updated to say ungit support node >=8 as we can see in the travis config file. WDYT @jung-kim  ?

tommai78101:
Also have this issue on CentOS 7. It's very likely this issue has an impact on RHEL flavored Linux OSes. Will also comment on #1203.

jung-kim:
Yup, can you share your node version you are running?  
It probably is at the version that is red for the [async func](https://node.green/#ES2017-features-async-functions) so once you upgrade your node you should be fine.

@tommai78101 This particular error is independent from the OS and solely contingent on the node version.  Please open a ticket if you see different errors on CentOS, or doesn't get fixed after node upgrade.

@ylecuyer yes I will update the README.

tommai78101:
@jung-kim I am seeing the same exact error and stack trace as what was taken in the screenshot (or camera capture). Because of this, I don't see the need to open a new ticket.

jung-kim:
README is updated with 8.16.0 as that is the minimal version of node we are running travis tests with.

And yeah once you upgrade node version >= 8.16.0 you should be fine

PR #1243, opened 2019-11-03
Remove Node 8 from build matrix

Hirse:
The node requirement in `package.json` is at least Node 10. Skipping builds for Node 8 should speed up the CI checks.

https://github.com/FredrikNoren/ungit/blob/eac5134c90d9198857fdd46e8d1e88b05f82715a/package.json#L98

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ node_js:
   - node
   - "12"
   - "10"
-  - "8"
 branches:
   only:
     - master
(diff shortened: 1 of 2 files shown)


campersau:
In the README it says [`8.16.0`](https://github.com/FredrikNoren/ungit/blob/master/README.md#installing) which was updated in August (related issue https://github.com/FredrikNoren/ungit/issues/1204#issuecomment-522127506)

Not quite sure why it was updated to `10.14` in the `package.json` last year in https://github.com/FredrikNoren/ungit/commit/866f7e1bd7dbe9b2a2a03739a76f036548c2e4cb by @jung-kim 

I think we can drop the support for `8.16.0` by the end of the year like [node does](https://nodejs.org/en/about/releases/).

Hirse:
Sounds reasonable. 
Should the field in the package.json be changed then?

PR #1244, opened 2019-11-04
More Consistent Diff Options

### LINKED ISSUE #1229: Inconsistent diff options
The order and state of the buttons are inconsistent:
image

Clicking on the `No Wrap` button also toggles the visual state of the `Show/Ignore white space diff` button:
image

https://github.com/FredrikNoren/ungit/pull/1225#issuecomment-545782743

Hirse:
Fixes #1229 

* Fix connected toggling 
* Use [Button Toolbar](https://getbootstrap.com/docs/3.3/components/#btn-groups-toolbar)
  * Fix order of buttons
  * Fix spacing of buttons
  * Fix border radius of buttons in staging
* Unify text of buttons (and tooltips)
* Don't change text on toggle
* Use inactive as default for all buttons

image

### DIFF
--- a/components/commitdiff/commitdiff.html
+++ b/components/commitdiff/commitdiff.html
@@ -1,17 +1,17 @@
 <div class="btn-toolbar" data-bind="visible: showDiffButtons" >
   <div class="btn-group btn-group-xs">
     <button class="btn btn-default bootstrap-tooltip commit-whitespace" data-bind="click: whiteSpace.toggle, css: {active: whiteSpace.isActive}" data-toggle="tooltip" data-placement="bottom" title="Hide whitespace changes in diff" data-delay='{"show":"2000", "hide":"0"}'>
-      Hide Whitespace
+      <span data-bind="text: whiteSpace.text"></span>
     </button>
   </div>
   <div class="btn-group btn-group-xs">
     <button class="btn btn-default bootstrap-tooltip commit-sideBySideDiff" data-bind="click: textDiffType.toggle, css: {active: textDiffType.isActive}" data-toggle="tooltip" data-placement="bottom" title="Show side by side diff view" data-delay='{"show":"2000", "hide":"0"}'>
-      Side by Side
+      <span data-bind="text: textDiffType.text"></span>
     </button>
   </div>
   <div class="btn-group btn-group-xs">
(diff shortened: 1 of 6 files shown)


components/commitdiff/commitdiff.html
```diff
@@ -1,17 +1,19 @@
...
-<div class="btn-group btn-group-xs" data-bind="visible: showDiffButtons">
-  <button class="btn btn-default bootstrap-tooltip pull-right commit-sideBySideDiff" data-bind="click: textDiffType.toggle, css: {active: textDiffType.isActive}" data-toggle="tooltip" data-placement="bottom" title="Show side by side diff view" data-delay='{"show":"2000", "hide":"0"}'>
-    <span data-bind="text: textDiffType.text"></span>
-  </button>
-</div>
-<div class="btn-group btn-group-xs" data-bind="visible: showDiffButtons">
-  <button class="btn btn-default bootstrap-tooltip pull-right commit-wordwrap" data-bind="click: wordWrap.toggle, css: {active: wordWrap.isActive}" data-toggle="tooltip" data-placement="bottom" title="Wrap words per line" data-delay='{"show":"2000", "hide":"0"}'>
-    <span data-bind="text: wordWrap.text"></span>
-  </button>
+<div class="btn-toolbar">
```
campersau: `showDiffButtons` binding can be added here instead of each of the `btn-group`s.

```suggestion
<div class="btn-toolbar" data-bind="visible: showDiffButtons">
```

campersau:
Hm, I am not sure about toggling only the state and not the text.

Currently the `focus` state looks very similar to the `active` state. This is a general issue but with your state only changes it struck me.
Here the `Side by Side` button is active and `Wrap Lines` was toggled from active to inactive but still has focus (not clicked anywhere else). And it looks almost like it is still active. Before at least the text change so you could see that something has happened.
image

Hirse:
I agree about the colors, but I still feel that changing the text is not a good idea.
These toggle-buttons behave like check boxes which also don't change their label depending on state.

Maybe the best solution would be to use a different visual for focus and active states. Bootstrap changed theirs with version 4: https://getbootstrap.com/docs/4.3/components/buttons/#toggle-states

(Actually migrating to Bootstrap 4 is probably a more long term change, but I might have a look at that.)

components/commitdiff/commitdiff.html
Hirse: Thanks, fixed in f799d90.

Hirse:
For now, I have reverted to changing the text on toggle, but have updated the texts compared to before.

PR #1245, opened 2019-11-06
Restore icons on graph actions using Octicons

Hirse:
Even before switching to the latest version of Octicons in #1224 these icons were not showing up for me.
This PR replaces the broken references to the old octicons and glyphicons with the SVG Octicons.
The icons should look the same as before with the exception of the Move icon. Instead of arrows in all directions, the new one points to the target of the move.

image

Also removes CSS classes for the very old image icons used before #605.

### DIFF
--- a/components/graph/git-graph-actions.js
+++ b/components/graph/git-graph-actions.js
@@ -1,4 +1,5 @@
 const ko = require('knockout');
+const octicons = require('octicons');
 const components = require('ungit-components');
 const programEvents = require('ungit-program-events');
 const RefViewModel = require('./git-ref.js');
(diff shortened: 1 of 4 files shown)


campersau:
Yes, they got lost in this big refactoring https://github.com/FredrikNoren/ungit/pull/1137

jung-kim:
Cool thank you!

Are you using prettifier?  Can you share which one?  I will try to use it as well so we can be more consistent on the code style, admittedly this was lacking on my part.

Hirse:
@jung-kim I didn't use a prettifier for this one, but I sometimes run the one included in VSCode (which I think would be [js-beautify](expandedIcon)).

In this PR, I used ESLint to find and add the semicolons and added the newlines manually.

PR #1246, opened 2019-11-07
Replace glyphicons with Octicons where possible

Hirse:
Use [octicons](https://octicons.github.com/) where possible to be more consistent. The glyphicons are only used for the checkboxes now.

Same icon
* Git error alert icon
* Repository list arrow-right
* Header back icon icon
* Stash apply pencil
* Image diff arrow
* Staging discard all (Note, this is a trashcan while the individual discard action uses `x`)

Changed icon
* Staging, stash button: paperclip -> pin
* Fetch from remote: download -> cloud download
* Submodules: four squares -> submodule file

Image diff has a right and a down arrow depending on the layout used for the screen width.

image


campersau:
Looks great! On my screen the back bottom at the top left looks a little bit off center:

image

Hirse:
What OS and browser are you using?

I think it looks fine on Windows Chrome and Firefox:
image

campersau:
In your chrome screenshot you can see that the arrow is a bit to high. Its just one/two pixel though.

Hirse:
Is that what you meant before? In your screenshot it also looks shifted to the right by a few px.

campersau:
Here is another screenshot from my other windows pc using chrome 78:
image

Reducing the height and with a little looks a bit better.
```css
height: 34px;
width: 35px;
```
image

PR #1248, opened 2019-11-11
fix electron portcheck to also include ungitBindIp as hostname

campersau:
also show error message to user

image

### DIFF
--- a/public/main.js
+++ b/public/main.js
@@ -60,6 +60,7 @@ app.on('ready', function() {
     if (ungitRunning) {
       dialog.showMessageBoxSync({
         type: 'error',
+        title: 'Ungit',
         message: 'Ungit instance is already running'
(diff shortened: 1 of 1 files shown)


public/main.js
```diff
@@ -37,29 +36,32 @@ function launch(callback) {
...
-    });
-  });
+  var server = net.createServer();
   server.on('error', function (e) {
     if (e.code == 'EADDRINUSE') {
-      callback(null, true);
+      callback(true);
     }
   });
+  server.listen(config.port, config.ungitBindIp, function() {
```
campersau: This is now the same as here: https://github.com/FredrikNoren/ungit/blob/bdde44bde09221b392842c703b8458c2687eb5d5/source/server.js#L362

PR #1256, opened 2020-01-01
Remove Node 8 from build matrix

campersau:
It is no longer supported. See https://github.com/FredrikNoren/ungit/pull/1243#issuecomment-549249550

@Hirse: I cherry-picked your commit from the previous PR.


jung-kim:
👍 Can we make change log changes as well?

Issue #1259, opened 2020-01-08
`--no-launchBrowser` is ignored when github already running


yukulele:
If ungit is already running and I exec:
```
ungit --no-launchBrowser
```
it open the browser.

It shouldn't

PR #1260, opened 2020-01-09
check for launchBrowser if ungit is already running

campersau:
the logic was also different in `bin/ungit` and `public/main.js` (electron)

- `--no-launchBrowser`: browser is never launched
- `--forcedLaunchPath=`: navigates to home screen
- `--forcedLaunchPath=path`: navigates to `/#/repository?path=path`

fixes https://github.com/FredrikNoren/ungit/issues/1259

### DIFF
--- a/bin/ungit
+++ b/bin/ungit
@@ -29,21 +29,16 @@ const openUngitBrowser = (pathToNavigateTo) => {
     .catch(err => console.log(`failed to navigate to ${pathToNavigateTo}`, err))
 }
 
-const getPathToNavigateTo= (override) => {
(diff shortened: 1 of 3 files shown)


bin/ungit
```diff
@@ -90,7 +85,6 @@ server.listen(config.port, config.ungitBindIp, (err) => {
 server.on('error', (e) => {
   if (e.code == 'EADDRINUSE') {
     console.log('Ungit server already running');
-    navigate(true);
-    cleanExit();
```
campersau: `cleanExit` prevented the browser from beeing launched on my machine (windows) because `open` now returns a promise. Letting node decide when to exit the process worked for me.
Could also return the promise from `navigate` but that would require more code.

Issue #1124, opened 2018-08-07
diff doesn't work for first commit


ylecuyer:
Because of this line: https://github.com/FredrikNoren/ungit/blob/2aae26ada4f010cf6d42568018351075564a5663/source/git-promise.js#L288

If you want to see the diff of the very first commit of a git repo you will see the current file content instead of the original one.

jung-kim:
I remember having an issue with the very first commit of a new repo as ungit try to compare with previous commit that doesn't exist.

I don't remember if I ever fixed it, but will fix it

jung-kim:
It seems to be working... Can you tell me how I can replicate?

<img width="794" alt="screen shot 2018-08-11 at 16 43 54" src="https://user-images.githubusercontent.com/5281068/43997035-cb117b40-9d85-11e8-828a-2cafbd994712.png">

ylecuyer:
You have to add another commit and modify the file to reproduce the error

jung-kim:
Ah Okay, will retest later and reevaluate

ylecuyer:
This also happens if you look for an old commit and locally modify the same file as one in the commit. Then the diff for the select file in the commit will show the current file content

jung-kim:
![screen shot 2018-10-15 at 20 16 51](https://user-images.githubusercontent.com/5281068/46990663-47faa000-d0b7-11e8-8bf6-9ff5f4a451ca.png)

It seems that I can see the 2nd and 1st commit...? are you at a particular screen dimension?

ylecuyer:
I have found a way to reproduce:

```
cd ~
mkdir ungit-temp
cd ungit-temp
git init
echo "a" > text.txt
git add -A
git commit -m "commit"
```
![first_commit](https://user-images.githubusercontent.com/1866809/48547590-46491580-e8cb-11e8-8c2e-2b2ca78f03e3.png)
Everything is OK

Revert the commit

![second](https://user-images.githubusercontent.com/1866809/48547605-5660f500-e8cb-11e8-9f35-54b6c33255e1.png)
Everything is OK

Check the first commit

![second_b](https://user-images.githubusercontent.com/1866809/48547627-624cb700-e8cb-11e8-977e-303f1c05cfdd.png)
NOT OK should be +a (though the +1/-0 is ok on top)

```
echo "b" > text.txt
``` 

Stage 

[...]

PR #1261, opened 2020-01-11
fix first commit diff

campersau:
also show correct commit diffs if same file is currently modified

fixes https://github.com/FredrikNoren/ungit/issues/1124

### DIFF
--- a/source/git-promise.js
+++ b/source/git-promise.js
@@ -8,6 +8,7 @@ const _ = require('lodash');
 const isWindows = /^win/.test(process.platform);
 const Bluebird = require('bluebird');
 const fs = require('./utils/fs-async');
+const gitEmptyReproSha1 = '4b825dc642cb6eb9a060e54bf8d69288fbee4904'; // https://stackoverflow.com/q/9765453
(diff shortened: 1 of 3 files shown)


source/git-promise.js
```diff
@@ -282,34 +282,30 @@ git.binaryFileContent = (repoPath, filename, version, outPipe) => {
 }
 
 git.diffFile = (repoPath, filename, sha1, ignoreWhiteSpace) => {
-  const newFileDiffArgs = ['diff', '--no-index', isWindows ? 'NUL' : '/dev/null', filename.trim()];
+  if (sha1) {
+    return git(['rev-list', '--max-parents=0', sha1], repoPath).then((initialCommitSha1) => {
+      let prevSha1 = sha1 == initialCommitSha1.trim() ? '4b825dc642cb6eb9a060e54bf8d69288fbee4904' : `${sha1}^`;
```
campersau: "empty tree" commit sha1. See https://stackoverflow.com/a/40884093/9663233

Hirse: Could you extract that sha into a constant and add this comment there too?

PR #1262, opened 2020-01-14
Fix diffs not appearing for renamed files

### LINKED ISSUE #1175: Rename + changes only show rename
It is not possible to see the diff when renaming and making change to the same file : 

```
Author: Ungit Ungit <ungit@example.com>
Date:   Wed Jan 16 15:11:19 2019 +0100

    rename + comment

1       0       lib/{check_middleware.rb => health_check.rb}
```

For example:

```
git init
echo "puts 'coucou'" > a.rb
git add a.rb
git commit -m "initial commit"
mv a.rb b.rb
echo "#com" >> b.rb
git add a.rb
git add b.rb
git commit -m "rename + change"
```
screenshot_2019-01-16_15-25-45

bakkerthehacker:
Fixes #1175

### DIFF
--- a/components/commit/commit.js
+++ b/components/commit/commit.js
@@ -55,8 +55,8 @@ class CommitViewModel {
     this.authorDateFromNow(this.authorDate().fromNow());
     this.authorName(args.authorName);
     this.authorEmail(args.authorEmail);
-    this.numberOfAddedLines(args.fileLineDiffs.length > 0 ? args.fileLineDiffs[0][0] : 0);
-    this.numberOfRemovedLines(args.fileLineDiffs.length > 0 ? args.fileLineDiffs[0][1] : 0);
+    this.numberOfAddedLines(args.total.additions);
+    this.numberOfRemovedLines(args.total.deletions);
     this.fileLineDiffs(args.fileLineDiffs);
     this.isInited = true;
     this.commitDiff = ko.observable(components.create('commitDiff', {
--- a/components/commitdiff/commitdiff.html
+++ b/components/commitdiff/commitdiff.html
@@ -19,7 +19,7 @@
 <div data-bind="foreach: commitLineDiffs" class="commitdiff">
   <div class="file">
     <div class="head commit-diff-filename" data-bind="click: fileNameClick">
-      <span data-bind="text: fileName"></span>
+      <span data-bind="text: displayName"></span>
       <span class="file-stats" data-bind="visible: added() != '-'">
         (
         <span data-bind="text: added"></span>
--- a/components/commitdiff/commitdiff.js
+++ b/components/commitdiff/commitdiff.js
@@ -14,7 +14,6 @@ class CommitDiff {
     this.wordWrap = args.wordWrap = args.wordWrap || components.create('textdiff.wordwrap');
     this.whiteSpace = args.whiteSpace = args.whiteSpace || components.create('textdiff.whitespace');
 
-    args.fileLineDiffs.shift();  // remove first line that has "total"
     this.loadFileLineDiffs(args);
   }
 
--- a/components/commitdiff/commitlinediff.js
+++ b/components/commitdiff/commitlinediff.js
@@ -5,10 +5,12 @@ const programEvents = require('ungit-program-events');
 
 class CommitLineDiff {
   constructor(args, fileLineDiff) {
-    this.added = ko.observable(fileLineDiff[0]);
-    this.removed = ko.observable(fileLineDiff[1]);
-    this.fileName = ko.observable(fileLineDiff[2]);
-    this.fileType = fileLineDiff[3];
+    this.added = ko.observable(fileLineDiff.additions);
+    this.removed = ko.observable(fileLineDiff.deletions);
+    this.fileName = ko.observable(fileLineDiff.fileName);
+    this.oldFileName = ko.observable(fileLineDiff.oldFileName);
+    this.displayName = ko.observable(fileLineDiff.displayName);
+    this.fileType = fileLineDiff.type;
     this.isShowingDiffs = ko.observable(false);
     this.repoPath = args.repoPath;
     this.server = args.server;
@@ -22,6 +24,7 @@ class CommitLineDiff {
   getSpecificDiff() {
     return components.create(`${this.fileType}diff`, {
       filename: this.fileName(),
+      oldFilename: this.oldFileName(),
       repoPath: this.repoPath,
       server: this.server,
       sha1: this.sha1,
--- a/components/imagediff/imagediff.js
+++ b/components/imagediff/imagediff.js
@@ -7,6 +7,7 @@ components.register('imagediff', args => new ImageDiffViewModel(args));
 class ImageDiffViewModel {
   constructor(args) {
     this.filename = args.filename;
+    this.oldFilename = args.oldFilename;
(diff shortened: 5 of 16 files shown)


campersau:
Thanks for the PR @bakkerthehacker !
It would be great if you could also add some unit tests in [spec.git-api.diff.js](https://github.com/FredrikNoren/ungit/blob/master/test/spec.git-api.diff.js).
Maybe for renaming a file and renaming a folder.

Does this also work in the staging area?

bakkerthehacker:
It seems the staging area does not show any renaming of files.  This only appears in the commit graph.  

As for writing tests, the tests in `spec.git-api.diff.js` seem to focus on the staging area with only one test that calls `/diff` with a sha and does no output verification.  I can flesh out the commit graph style of `/diff` calls and add in a rename test as well.

campersau:
If you use `git mv` on the command line then the rename is also visible in the staging area of ungit.
E.g. change the PLUGINS.md file and use `git mv PLUGINS.md PLUGIN.md` in the unigt repro.
image


For testing commit diffs you can use something similar to this https://github.com/FredrikNoren/ungit/pull/1261/files#diff-26d04449e2c60fa8035b4867ce9ca890R65-R75

bakkerthehacker:
I did a bit more tinkering and thinking about how to fix this the best way, and I'm not really happy with my solution.  It will fail on any non-renamed file that matches the regex.  

It seems that all this headache about the renamed files having the weird arrow syntax `{ => }` stems from `git log` being used instead of `git rev-parse`.  `git log` is a porcelain, or user facing command, while `git rev-parse` is the plumbing command, so `git log` will pretty print the renames for us, but that's not very useful for parsing commit data.  

I'm not super familiar with the differences between these commands.  I'm guessing replacing `git log` with the correct plumbing commands might be a difficult task.  Perhaps the `git log` formatting string could be customized to give the data in the format we want to fix this.  

What are your thoughts on this?  I would like to this this without having to munge the filename string after git log messes it up.

bakkerthehacker:
Yeah looks like the `--raw` and `-z` options together fix exactly this.  https://git-scm.com/docs/git-log#Documentation/git-log.txt--z

campersau:
Nice, this also works for `git diff` and `git status` so we could fix the staging area as well.

bakkerthehacker:
While, this change now fixes the missing diffs, the older filenames are discarded and the rename itself now looks at the new filename.  More work can be done to bring this old filename data to the UI.

campersau:
Yes, it would be great to also see the old filename of renames. Maybe we can also change the `fileLineDiff` to be an object instead of an array so it is similar to [`parseGitStatus`](https://github.com/FredrikNoren/ungit/blob/68d69bb150aa200dbf619c10adb9a6cdc4852131/source/git-parser.js#L7-L46). Like:
```js
{
  additions: 0,
  deletions: 0,
  fileName: "new.txt",
  type: "text",
  displayName: "old.txt → new.txt"
}
```

bakkerthehacker:
I'm much happier with this new version.  Stash, staging and commit graph now all support renamed files.  Hopefully I didn't miss any corner cases.  Tests still need updating. 

Also since the displayName is now completely controlled by ungit, we can use fancy arrows but lose out on the filepath-aware git renaming strategy that would put `{old => new}` inside the paths.

bakkerthehacker:
> I have no idea why all the tests have passed since they shouldn't have because of #1265

Looks like Travis and AppVeyor are pointed at a merge commit that puts my branch on top of current master.  I'm guessing that as soon as #1263 was pushed to master, they started working again.

ylecuyer:
Thanks for the fix, this is something I want to try as soon as it is available :)

ylecuyer:
I'm late to the party but I have an error when running master now :(

```
2020-01-29T09:07:12.418Z - warn: Responding with ERROR:  {"stack":"TypeError: lines.slice(...).values is not a function\n    at exports.parseGitStatus (/home/ylecuyer/Projects/ungit/source/git-parser.js:9:37)\n    at tryCatcher (/home/ylecuyer/Projects/ungit/node_modules/bluebird/js/release/util.js:16:23)\n    at Promise._settlePromiseFromHandler (/home/ylecuyer/Projects/ungit/node_modules/bluebird/js/release/promise.js:517:31)\n    at Promise._settlePromise (/home/ylecuyer/Projects/ungit/node_modules/bluebird/js/release/promise.js:574:18)\n    at Promise._settlePromise0 (/home/ylecuyer/Projects/ungit/node_modules/bluebird/js/release/promise.js:619:10)\n    at Promise._settlePromises (/home/ylecuyer/Projects/ungit/node_modules/bluebird/js/release/promise.js:699:18)\n    at _drainQueueStep (/home/ylecuyer/Projects/ungit/node_modules/bluebird/js/release/async.js:138:12)\n    at _drainQueue (/home/ylecuyer/Projects/ungit/node_modules/bluebird/js/release/async.js:131:9)\n    at Async._drainQueues (/home/ylecuyer/Projects/ungit/node_modules/bluebird/js/release/async.js:147:5)\n    at Immediate.Async.drainQueues [as _onImmediate] (/home/ylecuyer/Projects/ungit/node_modules/bluebird/js/release/async.js:17:14)\n    at runCallback (timers.js:696:18)\n    at tryOnImmediate (timers.js:667:5)\n    at processImmediate (timers.js:649:5)","message":"lines.slice(...).values is not a function"}
```

Is it related to my version of git ?

```
ylecuyer@hawk:~/Projects/ungit$ git --version
git version 2.17.1
```

I hope I'm not the only one and this wont be an issue for the release

ylecuyer:
Looks like it is more related to my version of node not supporting the `Array.values()`

I'm running:

```
ylecuyer@hawk:~/Projects/ungit$ node --version
v10.3.0
```

That's strange the CI didn't catch that

campersau:
@ylecuyer since https://github.com/FredrikNoren/ungit/pull/1256 we only support node >= 10.18 because that is the [latest version of this release](https://nodejs.org/about/releases/) and is tested by CI.
https://travis-ci.org/FredrikNoren/ungit/jobs/642151218#L1030
https://ci.appveyor.com/project/FredrikNoren/ungit/builds/30427522/job/uu0jij3qc3a7l5sh#L10

I am not quite sure yet if we should fix it. Looking at https://github.com/ylecuyer/ungit/commit/25148d4f65c53446491755ee5735855661d1de05 it seems easy...

ylecuyer:
I'll have to dig a little more but it looks like the `-z` option breaks some commit titles (mainly merge commits)
I came to the commit `9b44b2d875a6480b5b4489ea0e40a6344fc19983` with a git bisect

```
9b44b2d875a6480b5b4489ea0e40a6344fc19983 is the first bad commit
commit 9b44b2d875a6480b5b4489ea0e40a6344fc19983
Author: Grant Bakker <grant@bakker.pw>
Date:   Mon Jan 20 14:46:28 2020 -0500

    Use -z for parsing log data to fix renamed files getting munged

:040000 040000 53b91c7502870ce9675c001c00a205f4610b3dd6 0d3e95282f180de065a4e088f38f16407a396c76 M	source
:040000 040000 b1d1e85f6784d7f775c0f1fab4e1ba8d7c8a0751 91276e20fa030643e0c1ccba6febcd61bd30a305 M	test
```

campersau:
@ylecuyer can you please open a new issue?

PR #1267, opened 2020-01-28
Add merge squash action

ylecuyer:
This PR adds the merge squash action to the graph. I'm waiting for your feedback before writing tests :)

merge-squash

### DIFF
--- a/components/graph/git-graph-actions.js
+++ b/components/graph/git-graph-actions.js
@@ -10,6 +10,7 @@ const MergeViewModel = HoverActions.MergeViewModel;
 const ResetViewModel = HoverActions.ResetViewModel;
 const PushViewModel = HoverActions.PushViewModel;
 const SquashViewModel = HoverActions.SquashViewModel;
+const MergeSquashViewModel = HoverActions.MergeSquashViewModel;
 
 class ActionBase {
   constructor(graph, text, style, icon) {
--- a/components/graph/git-graph-actions.js
+++ b/components/graph/git-graph-actions.js
@@ -174,6 +174,11 @@ class MergeSquash extends Merge {
     this.text = "Merge Squash";
     this.style("merge-squash");
   }
+
+  perform() {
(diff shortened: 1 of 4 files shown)


components/graph/git-graph-actions.js
```diff
@@ -168,6 +169,26 @@ class Merge extends ActionBase {
...
+  createHoverGraphic() {
+    let onto = this.graph.currentActionContext();
+    if (!onto) return;
+    if (onto instanceof RefViewModel) onto = onto.node();
+    const path = onto.getPathToCommonAncestor(this.node);
+    return new MergeSquashViewModel(this.node, path);
+  }
+
+  perform() {
+    return this.server.postPromise('/squash', { path: this.graph.repoPath(), target: this.graph.currentActionContext().localRefName, squash: true })
```
campersau: `squash: true` is not used on the server and can be removed.

Hirse:
What is the difference in the result between "Squash" and "Merge Squash"?

If the outcome is the same, I would be in favor of having the name, color, and icon of the button also be the same.

ylecuyer:
> Looks good and maybe easier for some to understand then doing it the other way around via `Squash`.
> Is it possible to highlight the commits which are going to be merge squashed like in `Squash`?
> 
> Squash:
> image
> 
> Merge Squash:
> image

done 

> What is the difference in the result between "Squash" and "Merge Squash"?

It is the same if you select the right nodes. But I prefer to have the proposed named in the PR as this is the common wording used to describe this action, see for example the github button :

select-squash-and-merge-from-drop-down-menu

Hirse:
So, if I understand correctly, the new "Merge Squash" does the same as the existing "Squash" on the newest (topmost) commit?

1. Should we change it to "Squash Merge" then?
2. Should we change the previous "Squash" to "Squash Merge" too then?

campersau:
After thinking more about the UI I also think @Hirse renames suggestions would make sense. Also it might be good to unify the two squash actions even further by using the same color and icon.

Here is my suggestion:
image

image

I have picked the `Squash & Merge` spelling because we already have `Commit & Push` elsewhere.

What do you/others think?

PR #1269, opened 2020-01-31
Remove dependency on npm

Hirse:
Currently, ungit has a dependency on the whole `npm` package, which is only used to check latest versions on the npm registry.

This change replaces `npm` with more light-weight solutions to provide the same functionality.

### DIFF
--- a/test/spec.git-api.submodule.js
+++ b/test/spec.git-api.submodule.js
@@ -1,4 +1,3 @@
-
 const expect = require('expect.js');
 const request = require('supertest');
 const express = require('express');
@@ -21,9 +20,9 @@ describe('git-api submodule', function () {
 
   before(() => {
     return common.createSmallRepo(req)
-      .then((dir) => { testDirMain = dir })
+      .then((dir) => { testDirMain = dir; })
       .then(() => common.createSmallRepo(req))
-      .then((dir) => { testDirSecondary = dir });
+      .then((dir) => { testDirSecondary = dir; });
   });
 
   after(() => common.post(req, '/testing/cleanup'));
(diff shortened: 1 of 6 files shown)


campersau:
Quick review:
- All resolved urls in the package lock now point to `https://domoreexp.pkgs.visualstudio.com/_packaging/npm-mirror/npm/registry` instead of `https://registry.npmjs.org`
- I think `pkg-versions` can be a `devDependency`

Will look into it more during the weekend.

Hirse:
Fixed both of those.
I had messed up my registry from work so it was pointing to a mirror.

campersau:
The build is currently failing because `npm` was previously required in [`sysinfo.js`](https://github.com/FredrikNoren/ungit/pull/1269/files#diff-fca91a28999b88422fcd756a69215278L4) which did load [`graceful-fs`](https://github.com/npm/cli/blob/f533d61eb27c484fae84571fbea8bc1d83acb42b/lib/npm.js#L19-L20) which then replaced some `fs` functions and changed the behavior for having to specify callbacks.

We currently have a lot of `fs` calls for testing which don't specify a callback and maybe more: https://github.com/FredrikNoren/ungit/blob/621a04f931ea9007ac826c04a1a02832e20aa470/source/git-api.js#L699-L720
We could also change them to use their `*Async` versions instead because we promisify `fs` here
https://github.com/FredrikNoren/ungit/blob/621a04f931ea9007ac826c04a1a02832e20aa470/source/utils/fs-async.js#L4

campersau:
This will also reduce the electron package size by a few MB. (ungit-linux-x64.zip 98.9MB → 91.6MB in my environment.)

test/spec.git-api.submodule.js
```diff
@@ -35,31 +34,36 @@ describe('git-api submodule', function () {
...
-        displayName: '.gitmodules',
-        isNew: true,
-        staged: true,
-        removed: false,
-        conflict: false,
-        renamed: false,
-        type: 'text',
-        additions: '3',
-        deletions: '0'
+    return common.get(req, '/status', { path: testDirMain })
```
campersau: Nice catch!

PR #1273, opened 2020-02-06
Update diff2html to version 3

campersau:
just to keep up with the API changes.

See https://github.com/rtfpessoa/diff2html/pull/297#issuecomment-576030591


Hirse:
Are there no changes required to the CSS?

public/less/d2h.less
```diff
@@ -1,218 +1,434 @@
...
+.selecting-left .d2h-code-line *::selection,
+.selecting-right td.d2h-code-linenumber::selection,
+.selecting-left .d2h-code-side-line::selection,
+.selecting-left .d2h-code-side-line *::selection,
+.selecting-right td.d2h-code-side-linenumber::selection,
+.selecting-right td.d2h-code-side-linenumber *::selection {
+  background: transparent;
+}
+
+/*** adjustments for ungit ***/
```
campersau: Everything above this line is copied from https://github.com/rtfpessoa/diff2html/blob/4e28c5c3275f6f703a6923d23d001bfb676331a6/src/ui/css/diff2html.css
Everything below this line is customized for ungit.
Maybe we should split these into separate files.

public/less/d2h.less
```diff
@@ -1,218 +1,434 @@
...
-.d2h-wrapper {
-  display: block;
-  margin: 0 auto;
-  text-align: left;
-  width: 100%;
-}
+ *
+ * Diff to HTML (diff2html.css)
+ * Author: rtfpessoa
+ *
```
Hirse: Maybe add the version number of diff2html here.

campersau: Not sure, this part is copied 1:1

campersau:
NEW | OLD

image

image

public/less/d2h.less
campersau: Opted to include the css file from npm package so this file now only contains the adjustments.

Hirse:
Thanks for the comparison images.
Where does the difference in height come from?
Also, in the bottom case (red) the line number color changes.

campersau:
The height difference came from the different font (monospace) and font-size (13px) which I have changed back to the previous values here https://github.com/FredrikNoren/ungit/pull/1273/commits/14a6e9be86f6f7b4e01b6c8eb9f81789b3434af0

Some other things which changed:
- The lines which contain changes are no longer indented a little bit more than the lines without changes
- The line numbers are now always aligned to the right
- The line numbers are now the same color in `line by line` vs `side by side` diffs
- Also fixed the checkbox alignment for patches

Updated screenshots from the new version:
image

image

image

image

image

image

public/less/d2h.less
```diff
@@ -1,218 +1,64 @@
...
-  .word-wrap & {
-    white-space: -moz-pre-wrap;  /* Mozilla, since 1999 */
-    white-space: -webkit-pre-wrap; /*Chrome & Safari */
-    white-space: -pre-wrap;      /* Opera 4-6 */
-    white-space: -o-pre-wrap;    /* Opera 7 */
-    white-space: pre-wrap;       /* css-3 */
-  }
+.d2h-code-linenumber, .d2h-code-side-linenumber, .d2h-code-line-prefix {
+  -moz-user-select: none;
+  -ms-user-select: none;
```
Hirse: These don't seem to be necessary any more:
https://developer.mozilla.org/en-US/docs/Web/CSS/user-select#Browser_compatibility

campersau: Only in the [diff2html UI](https://github.com/rtfpessoa/diff2html#diff2htmlui-usage) which we are not using. Also it seems a bit buggy. I have opened https://github.com/rtfpessoa/diff2html/issues/303 for that.

PR #1275, opened 2020-02-09
[fix #1091] performance boost for big repos

### LINKED ISSUE #1091: Two slow for big repository.
I suggest use in-memory database for log history:_)

jung-kim:
Few things are done here.

- ignore 'rename' filewatch event as it can cause constant update and refresh loop
- Prevent full gitlog history from server to client and load only what is needed
- Prevent redundant ref and node calculations per each `/gitlog` api call

### ignore 'rename' filewatch event as it can cause constant update and refresh loop
This one is bit tricky as some OS, git and node version behave differently and the [`fs.watch()` doc](https://nodejs.org/docs/latest/api/fs.html#fs_fs_watch_filename_options_listener) itself does describe very inconsistent behavior.  

> On most platforms, 'rename' is emitted whenever a filename appears or disappears in the directory.

Here is a problem with that for some combination of OS, git and node versions:

1.  with `/refs` git api call, git operations would continuously delete and create some files such as `.git/refs/remotes/origin/master`
2. Which triggers `fs.watch()` remote event
3. Which emits `git-directory-changed` socket io event
4. Which front end picks up and does many things but most importantly makes two api calls, `/gitlog` and `/refs`
    - `/gitlog` will seriously bog down entire performance both and back end and frontend as it is more compute intense operation to recalculate and draw
    - `/refs` will trigger everything back to step 1 and entire thing repeats

There is no way to test and prepare for all combinations.  We should stick with latest doc and latest API.

So, adopted to latest API and ignoring `rename` event.

### Prevent full gitlog history from server to client and load only what is needed

Upon page load, client will make and API call to get list of git nodes.  However, this call gets list of ALL nodes every time.  i.e. if a pages is showing 300 nodes and user scrolls down asking for 25 more nodes, backend will return 325 nodes, not just the 25 nodes needed.

This is now fixed and backend will return only the nodes client needs.

### Prevent redundant ref and node calculations per each `/gitlog` api call

One of the biggest performance hit for big repos was due to git refs calculations because ungit loads ALL refs, local and remote branches and tags of it's history.

This is needed to some degree to display them in the branch drop down and etc, but there were some inefficiencies on logic that thinking these refs had valid nodes and did unnecessary calculations.  

### todo
- [ ] fix tests

-----

https://github.com/git/git page load with old code: https://recordit.co/O1lzQO62Z2

with old code
-----

https://github.com/git/git page load with new code: https://recordit.co/bHmV8QiA0I

with new code

### DIFF
--- a/components/graph/git-node.js
+++ b/components/graph/git-node.js
@@ -12,13 +12,14 @@ const maxTagsToDisplay = ungit.config.numRefsToShow - maxBranchesToDisplay; // 2
 class GitNodeViewModel extends Animateable {
   constructor(graph, sha1) {
     super(graph);
+    this.hasBeenRenderedBefore = false;
     this.graph = graph;
     this.sha1 = sha1;
     this.isInited = false;
     this.title = ko.observable();
     this.parents = ko.observableArray();
     this.commitTime = undefined; // commit time in string
-    this.date = undefined;       // commit time in numeric format for sort
+    this.timestamp = undefined;  // commit time in numeric format for sort
     this.color = ko.observable();
     this.ideologicalBranch = ko.observable();
     this.remoteTags = ko.observableArray();
@@ -133,8 +134,14 @@ class GitNodeViewModel extends Animateable {
       }
 
       this.color(this.ideologicalBranch() ? this.ideologicalBranch().color : '#666');
+      if (!this.hasBeenRenderedBefore) {
+        // push this nodes into the graph's node list to be rendered if first time.
+        // if been pushed before, no need to add to nodes.
+        this.hasBeenRenderedBefore = true;
+        graph.nodes.push(this);
+      }
       this.animate();
-    }, 500, {leading: true})
+    }, 500, { leading: true })
   }
 
   getGraphAttr() {
@@ -146,11 +153,16 @@ class GitNodeViewModel extends Animateable {
     this.element().setAttribute('y', val[1] - 30);
   }
 
+  setParent(parent) {
+    this.aboveNode = parent;
+    if (parent) parent.belowNode = this;
+  }
+
   setData(logEntry) {
     this.title(logEntry.message.split('\n')[0]);
     this.parents(logEntry.parents || []);
     this.commitTime = logEntry.commitDate;
-    this.date = Date.parse(this.commitTime);
+    this.timestamp = logEntry.timestamp || Date.parse(this.commitTime);
     this.commitComponent.setData(logEntry);
     this.signatureMade(logEntry.signatureMade);
     this.signatureDate(logEntry.signatureDate);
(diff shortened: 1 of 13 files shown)


source/git-api.js
```diff
@@ -71,9 +71,8 @@ exports.registerApi = (env) => {
           });
         }
       }).then(() => {
-        const watcher = fs.watch(pathToWatch, options || {});
```
campersau: Please use the file watcher events which I added. See  https://github.com/FredrikNoren/ungit/pull/1263#discussion_r369016772

jung-kim: Ah.... why?  API documentation does not mention error event.

jung-kim: https://nodejs.org/api/fs.html#fs_fs_watch_filename_options_listener

ylecuyer:
Nice opportunities to speed up ungit for both repos 👍

Though could you split the PR in different branches so that they can be tested individually?

source/git-api.js
campersau: The returned `fs.FSWatcher` does have the events: https://nodejs.org/api/fs.html#fs_class_fs_fswatcher
And it is important to listen for the error event otherwise the process would crash if the event isn't handled. This was a source of clicktest failures because we deleted the temp directory while the directory was still openend in ungit.

jung-kim: You are right, I find this api design to be confusing...  multiple places to register events and when you register `change` event listener, it would multiple sub event types?  This is bit odd especially when `fswatcher{}` is not shared anywhere else.

Yeah I will fix it up later

jung-kim: Good catch!

jung-kim:
Yeah you are right, let me break this one out to two:

one with `ignore 'rename' filewatch event as it can cause constant update and refresh loop
` and `Prevent redundant ref and node calculations per each /gitlog api call`.  

But inevitably `Prevent full gitlog history from server to client and load only what is needed` will be disproportionally as it impacts many places.

components/graph/git-node.js
```diff
@@ -4,20 +4,22 @@ const components = require('ungit-components');
...
     super(graph);
+    this.hasBeenRenderedBefore = false;
     this.graph = graph;
     this.sha1 = sha1;
     this.isInited = false;
     this.title = ko.observable();
     this.parents = ko.observableArray();
     this.commitTime = undefined; // commit time in string
-    this.date = undefined;       // commit time in numeric format for sort
+    this.timestamp = undefined;  // commit time in numeric format for sort
```
Hirse: What is the "numeric format"?
Is that ISO or `yyyyMMddhhmm`?

ylecuyer:
I'm running the branch before the split for some days now and I often had this error : 

Screenshot_2020-02-12_10-42-12

Am I the only one with this glitch ?

I should try one branch and the other now that the PR has been split

jung-kim:
This PR definitely changes and challenges how things are being calculated.  

Definitely needs lots testing and incubations.  Will keep my eyes out for these kind of problems

PR #1277, opened 2020-02-11
performance optimizations for big repos

jung-kim:
- Subset of #1275 
- cc #1091

--- 

### Prevent full gitlog history from server to client and load only what is needed

Upon page load, client will make and API call to get list of git nodes. However, this call gets list of ALL nodes every time. i.e. if a pages is showing 300 nodes and user scrolls down asking for 25 more nodes, backend will return 325 nodes, not just the 25 nodes needed.

This is now fixed and backend will return only the nodes client needs.

Prevent redundant ref and node calculations per each /gitlog api call
One of the biggest performance hit for big repos was due to git refs calculations because ungit loads ALL 

### refs, local and remote branches and tags of it's history.

This is needed to some degree to display them in the branch drop down and etc, but there were some inefficiencies on logic that thinking these refs had valid nodes and did unnecessary calculations.


wmertens:
I think this can be closed, see #1461

PR #1282, opened 2020-02-21
Fix issue 1281, null forcedLaunchPath

martinmcclure:
Fixed by adding explicit handling of null forcedLaunchPath. Ideally there would be a test for null forcedLaunchPath, but I'm afraid I haven't yet figured out how to write a test for that. Manual testing of the feature works.

### DIFF
--- a/bin/ungit
+++ b/bin/ungit
@@ -31,11 +31,9 @@ const openUngitBrowser = (pathToNavigateTo) => {
 
 const navigate = () => {
   let url = config.urlBase + ':' + config.port;
-  if (config.forcedLaunchPath === null) {
-    url += '/#/';
-  } else if (config.forcedLaunchPath === undefined) {
+  if (config.forcedLaunchPath === undefined) {
     url += '/#/repository?path=' + encodeURIComponent(process.cwd());
-  } else if (config.forcedLaunchPath !== undefined && config.forcedLaunchPath !== '') {
+  } else if (config.forcedLaunchPath !== null && config.forcedLaunchPath !== '') {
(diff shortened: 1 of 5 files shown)


bin/ungit
```diff
@@ -31,7 +31,9 @@ const openUngitBrowser = (pathToNavigateTo) => {
 
 const navigate = () => {
   let url = config.urlBase + ':' + config.port;
-  if (config.forcedLaunchPath === undefined) {
+  if (config.forcedLaunchPath === null) {
+    url += '/#/';
+  } else if (config.forcedLaunchPath === undefined) {
     url += '/#/repository?path=' + encodeURIComponent(process.cwd());
   } else if (config.forcedLaunchPath !== undefined && config.forcedLaunchPath !== '') {
```
campersau: It is easier to just fix this line 😉 . The `undefined` check does not make much sense here since it is already checked above.
Please also make this change in this file
https://github.com/FredrikNoren/ungit/blob/5c01beb7a229f816eb0abcbdabb9cab8e7ee2e6a/public/main.js#L22-L26
which is used by eletron.

```suggestion
  } else if (config.forcedLaunchPath !== null && config.forcedLaunchPath !== '') {
```

---

The `null` check was accidentally removed in https://github.com/FredrikNoren/ungit/pull/1007 and I failed to fix it here https://github.com/FredrikNoren/ungit/pull/1260 because I just tested the command arguments and not the `.ungitrc`.

martinmcclure:
@campersau Thanks for the improvements! I believe I have applied them.

campersau:
Thanks! Merged manually because I messed up the changelog before...

PR #1283, opened 2020-02-23
bump dependencies

campersau:
- `mkdirp` is now using native promises
- `getmac` is now synchronous
- `node-cache`  is now synchronous and does not throw errors on missing keys
- `winston` changed transport api


package.json
```diff
@@ -23,54 +23,54 @@
...
     "snapsvg": "~0.5.1",
-    "socket.io": "~2.2.0",
-    "superagent": "~5.0.5",
-    "temp": "~0.9.0",
-    "winston": "2.4.0",
-    "yargs": "~13.2.2"
+    "socket.io": "~2.3.0",
+    "superagent": "~5.2.2",
+    "temp": "~0.9.1",
+    "winston": "3.2.1",
```
Hirse: Why is this the only pinned dependency?

source/sysinfo.js
```diff
@@ -12,15 +12,14 @@ exports.getUngitLatestVersion = () => {
...
-    });
-  });
+  let addr;
+  try {
+    addr = getMac();
+  } catch (err) {
+    winston.error("attempt to get mac addr failed, using fake mac.", err);
+    addr = "abcde";
+  }
+  return Bluebird.resolve(md5(addr));
```
Hirse: Since the function is no longer async, would it make sense to not return a promise?

package.json
campersau: Thanks for catching that. It shouldn't be pinned anymore because I adjusted the code to the new API.

source/utils/cache.js
```diff
@@ -11,15 +11,25 @@ const funcMap = {}; // Will there ever be a use case where this is a cache with
...
+      return;
+    }
+    getHardValue(result.func()) // func is found, resolve, set with TTL and return result
+      .then((r) => {
+        cache.set(key, r, funcMap[key].ttl)
+        resolve(r);
+      }).catch((err) => {
+        reject(err);
+      });
+  });
```
Hirse: I'm not quite sure what this function does, but I think you can prevent the new-promise-antipattern like this:
```js
  let result = cache.get(key);
  if (result !== undefined) {
    return Bluebird.resolve(result);
  }
  result = funcMap[key];
  if (result === undefined) {
    return Bluebird.reject(new Error(`Cache entry ${key} not found`));
  }
  return getHardValue(result.func()) // func is found, resolve, set with TTL and return result
    .then((r) => {
      cache.set(key, r, funcMap[key].ttl)
      return r;
    });
```

source/utils/winston.js
```diff
@@ -0,0 +1,24 @@
...
+          const splatStr = splat.map((arg) => JSON.stringify(arg)).join("\n");
+          return `${info.timestamp} - ${info.level}: ${info.message} ${splatStr}`;
+        }
+        return `${info.timestamp} - ${info.level}: ${info.message}`;
+      })
+    )
+  }));
+}
+
+module.exports = winston;
```
Hirse: It feels a bit weird to have this file be imported/required everywhere seemingly for its side effects.
Is there not a better way to ensure the logging is set up correctly?

source/utils/winston.js
```diff
@@ -0,0 +1,24 @@
...
+  return transport instanceof winston.transports.Console;
+});
+if (!consoleTransport) {
+  winston.add(new winston.transports.Console({
+    level: "error",
+    format: winston.format.combine(
+      winston.format.timestamp(),
+      winston.format.colorize(),
+      winston.format.printf((info) => {
+        const splat = info[Symbol.for("splat")];
```
Hirse: What is "splat"?
Is that coming from winston?

campersau: Yes, it does and it is an array of all the arguments after the first one like our error objects:
https://github.com/FredrikNoren/ungit/blob/dc60a9ddfa66b312985b2c7f1187582bfb77eb7d/source/sysinfo.js#L18

https://github.com/winstonjs/winston#streams-objectmode-and-info-objects

source/utils/cache.js
campersau: That's exactly what I had done first but then this test failed
https://github.com/FredrikNoren/ungit/blob/89a1d227935cb8772233e752a5e4c3be750e835a/test/spec.cache.js#L19
which throws an exception on `result.func()` which then isn't wrapped as a rejected promise...

source/utils/winston.js
campersau: I have changed it so it is only required by ungits entrypoints which are:
- `bin/credentials-helper`
- `bin/ungit`
- `nmclicktests/environment.js`
- `public/main.js`
- `source/server.js`

With winston 3 there is no default console transport that's why I did it this way. Also if no transport is specified winston will complain about that every time something is logged.

source/server.js
```diff
@@ -19,19 +19,28 @@ const Bluebird = require('bluebird');
 
 process.on('uncaughtException', (err) => {
   winston.error(err.stack ? err.stack.toString() : err.toString());
-  bugtracker.notify.bind(bugtracker, err, 'ungit-launcher');
+  bugtracker.notify(err, 'ungit-launcher');
```
campersau: @jung-kim @FredrikNoren Is the raven (sentry.io) bugtracking still active or can it be removed?

This got broken two years ago https://github.com/FredrikNoren/ungit/commit/543449e16ac83e1635d300c4e66b9121536e815e#diff-71e7385fb3d187340c7741bf6d4f82ffL25-R27

public/main.js
```diff
@@ -1,14 +1,15 @@
 var startLaunchTime = Date.now();
+var winston = require('../source/utils/winston');
```
Hirse: Is winston used in this file or just imported for side effects?

campersau: Currently only for the side effects.

source/bugtracker.js
```diff
@@ -12,15 +12,9 @@ class BugTracker {
...
     this.appVersion = 'unknown';
-    this.userHash = 'unkown';
+    this.userHash = sysinfo.getUserHash();
     this.appVersion = config.ungitDevVersion;
     winston.info(`BugTracker set version: ${this.appVersion}`);
-
-    sysinfo.getUserHash()
-      .then((userHash) => {
-        this.userHash = userHash;
-        winston.info('BugTracker set user hash');
```
Hirse: Do we not need this logging or is it done elsewhere?

source/git-promise.js
```diff
@@ -1,8 +1,8 @@
 const child_process = require('child_process');
 const gitParser = require('./git-parser');
 const path = require('path');
-const config = require('./config');
 const winston = require('winston');
+const config = require('./config');
```
Hirse: If you are ordering the imports, might as well order all of them. 😄

campersau: I have reverted it already but not pushed yet.

source/bugtracker.js
campersau: I think that this isn't needed.

source/server.js
jung-kim: I'm not sure anymore as either I cannot find my access nor I no longer have access.

We could potentially open a new free account and use it that way as it is very possible that account is no longer active.

PR #1284, opened 2020-02-23
Change how textdiff settings changes are handled

campersau:
Fixes performance issues with `signal` and should use less memory.
This is the PR for one of my proposals from https://github.com/FredrikNoren/ungit/issues/1091#issuecomment-586589079

### DIFF
--- a/components/textdiff/textdiff.js
+++ b/components/textdiff/textdiff.js
@@ -17,6 +17,7 @@ const loadLimit = 100;
 class WordWrap {
   constructor() {
     this.value = ko.observable(false);
+
     this.toggle = () => {
       this.value(!this.value());
     };
@@ -33,9 +34,6 @@ class Type {
     }
 
     this.value = ko.observable(ungit.config.diffType || textDiff);
-    this.value.subscribe(() => {
-      programEvents.dispatch({ event: 'invalidate-diff-and-render' });
-    });
 
     this.toggle = () => {
       this.value(this.value() === textDiff ? sideBySideDiff : textDiff);
@@ -48,9 +46,7 @@ class Type {
 class WhiteSpace {
   constructor() {
     this.value = ko.observable(ungit.config.ignoreWhiteSpaceDiff);
-    this.value.subscribe(() => {
-      programEvents.dispatch({ event: 'invalidate-diff-and-render' });
-    });
+
     this.toggle = () => {
       this.value(!this.value());
     };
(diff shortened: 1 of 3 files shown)


components/textdiff/textdiff.js
```diff
@@ -79,16 +75,15 @@ class TextDiffViewModel {
...
-        this.invalidateDiff();
-        if (this.isShowingDiffs()) this.render();
-      }
-    });
-
     this.isShowingDiffs.subscribe(newValue => {
       if (newValue) this.render();
     });
+    this.textDiffType.value.subscribe(() => {
+      if (this.isShowingDiffs()) this.render();
```
Hirse: Does a change in type no longer invalidate the diffs?

campersau: No, because it isn't needed.
The diff type setting and "load more" are client side features and don't require server requests.
The whitespace setting on the other hand is a git feature and requires a server request.

components/textdiff/textdiff.js
```diff
@@ -129,63 +124,70 @@ class TextDiffViewModel {
...
+          const remaining = this.loadCount - lineCount;
+          if (remaining > 0) {
+            if (remaining < length) {
+              if (!block.allLines) {
+                block.allLines = block.lines;
+              }
+              block.lines = block.allLines.slice(0, remaining);
+            } else if (block.allLines) {
+              block.lines = block.allLines;
+              block.allLines = undefined;
```
jung-kim: at block level, is there a reason to have `allLines` separately?

We could loosely follow the `line count` restrictions.  if limit is 10 and we first block has 12 lines of code, we could just display all 12 lines without doing this.  

Not sure if it is worth to follow that diff line limit strictly

campersau: It was there before (we implemented it like this 5 years ago https://github.com/FredrikNoren/ungit/commit/7bcd34c84e2516f5a54ff191879d22a10da433e7). I just changed it to enable "loading more" without invalidating the diff (reverting and really fixing https://github.com/FredrikNoren/ungit/pull/876).
But yes, beeing strict has also some bugs like this one where the whitespace changes are only visible after loading the hole block:

image

But one code block could also be huge like hundreds of lines and defeating the purpose of https://github.com/FredrikNoren/ungit/pull/376...

jung-kim: For sure there are some tech debts here.  If I remember correctly, the main cause of inefficiencies targeted in #376 was due to animation loop moving many elements and triggering recalculations that was blocked by animation loop due to large diff box opened up, which ultimately addressed in #630.  Performance part of the incremental load shouldn't be a consideration any more. 

However, if we are concerned by unexpectedly adding large block.  We could do conservative approach and load only that is below the limit.

campersau: I have changed the logic to always load complete blocks as long as the load limit is not reached.

PR #1296, opened 2020-03-18
GitHub Action to automatically bump dependencies and create pull request

campersau:
This will run our `bumpdependencies` grunt task every day and if the dependencies have changed automatically create a PR with the updated dependency versions.

Example: https://github.com/campersau/ungit/actions/runs/55682312

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -1,0 +1,35 @@
+name: Bump Dependencies
+
+on:
+  push:
+    branches:
+      - master
+  schedule:
+    - cron: '0 0 * * *'
+
+jobs:
+  bump:
+    if: github.repository == 'FredrikNoren/ungit'
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '*'
+
+      - run: npm ci
+      - run: ./node_modules/.bin/grunt bumpdependencies
+      - run: npm install
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v2
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: Bump Dependencies
+          title: Bump Dependencies
+          body: Bump Dependencies
+          labels: dependencies
+          branch: bumpdependencies


Hirse:
Sounds good, but what is the benefit of this over using a service like [Dependabot](https://dependabot.com) for all dependencies?

campersau:
Don't have the permission to do that and just wanted to play with GitHub Actions 😄

Hirse:
Sounds good. 
Is it possible to also use GitHub actions for the build?
I tried before, but couldn't get it to work with Nightmare. 😞

campersau:
Yes, I have done it here https://github.com/campersau/ungit/blob/e25633891ca92f1ed0a0c90f86e15ae2c12ed59b/.github/workflows/ci.yml
Sample run: https://github.com/campersau/ungit/actions/runs/55682534

Here are some random things about that:
- The setup-node action does [not print out the actual installed version](https://github.com/actions/setup-node/issues/90) which is why [this](https://github.com/campersau/ungit/blob/e25633891ca92f1ed0a0c90f86e15ae2c12ed59b/.github/workflows/ci.yml#L27) line exists.
- On windows because the username `runneradmin` is too long it gets truncated in some of our tests to use the stupid `RUNNER~1` syntax. That's what [this](https://github.com/campersau/ungit/blob/e25633891ca92f1ed0a0c90f86e15ae2c12ed59b/.github/workflows/ci.yml#L46-L47) line is for. See https://github.com/mvdan/github-actions-golang#known-bugs.
- Publishing artifacts is cool but since the upload-artifact action currently only supports single files or folders [this](https://github.com/campersau/ungit/blob/e25633891ca92f1ed0a0c90f86e15ae2c12ed59b/.github/workflows/ci.yml#L60-L125) is a bit repetitive. Appveyor [supports that](https://github.com/FredrikNoren/ungit/blob/670bef2bc918e174f6d864c958b07e7d055107e7/appveyor.yml#L34) which is nice. See [result](https://ci.appveyor.com/project/FredrikNoren/ungit/builds/31564567/job/wx45gxf50mq6ne6w/artifacts)
- Also I currently upload the zipped version it gets [zipped again](https://github.com/actions/upload-artifact/issues/39) by GitHub. I could also try to use the upload-artifact action to zip the folder for me but I haven't tested it yet. See  https://github.com/FredrikNoren/ungit/pull/1241#issuecomment-556998552 for possible issues.
- A good thing is that GitHub actions use the actual [powershell](https://github.blog/changelog/2019-10-17-github-actions-default-shell-on-windows-runners-is-changing-to-powershell/) on windows instead of travis which uses the [git for windows bash](https://docs.travis-ci.com/user/reference/windows/#git-bash) that's why [updating git](https://github.com/FredrikNoren/ungit/blob/670bef2bc918e174f6d864c958b07e7d055107e7/.travis.yml#L51) inside the git bash doesn't work there but in GitHub actions.
- One of the most annoying things is that GitHub actions don't trigger other GitHub actions which means that the PR created by this `bump.yml` workflow would not trigger my `ci.yml` workflow. See https://github.com/peter-evans/create-pull-request/issues/48 and an example PR https://github.com/campersau/ungit/pull/3. But it seems that third party CI runs see https://github.com/FredrikNoren/ungit/pull/1297.

Will do a PR later this week.

Hirse:
Looks pretty good. 
Thanks for all the details. Maybe you could include some of those as comments into the file (or link to a page that lists them all).

Issue #1302, opened 2020-03-26
Unable to launch ungit; git-parser error after update


netrbrunet:
Updated ungit and it will no longer launch. Subsequently updated GitBash on my Windows 10 in order to try and resolve, but the issue persisted. Full removal of the GitBash npm and npm_cache directories and re-install of ungit, but issue persists.
```
r*********t@LT-****** MINGW64 /c/users/r*********t/loader-code (master)
$ stderr: C:\Users\r*********t\AppData\Roaming\npm\node_modules\ungit\source\git-parser.js:47
const fileChangeRegex = /(?<additions>[\d-]+)\t(?<deletions>[\d-]+)\t((?<fileName>[^\x00]+?)\x00|\x00(?<oldFileName>[^\x00]+?)\x00(?<newFileName>[^\x00]+?)\x00)/g;
                        ^

SyntaxError: Invalid regular expression: /(?<additions>[\d-]+)\t(?<deletions>[\d-]+)\t((?<fileName>[^\x00]+?)\x00|\x00(?<oldFileName>[^\x00]+?)\x00(?<newFileName>[^\x00]+?)\x00)/: Invalid group
    at Object.<anonymous> (C:\Users\r*********t\AppData\Roaming\npm\node_modules\ungit\source\git-parser.js:47:25)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
    at Function.Module._load (module.js:497:3)
    at Module.require (module.js:596:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (C:\Users\r*********t\AppData\Roaming\npm\node_modules\ungit\source\git-api.js:3:19)
    at Module._compile (module.js:652:30)
Stopped keeping ungit alive

[1]+  Done                    ungit

```

campersau:
What version of node do you use? >= 10.18 is required. See also https://github.com/FredrikNoren/ungit/issues/1280#issuecomment-588084416 and https://github.com/FredrikNoren/ungit/pull/1256

netrbrunet:
Fantastic, that did it @campersau .
Just installed latest node.js and I'm off and running again.

Was there a way I should have seen that? Anything in the errors or a dependency note when re-installing ungit?

campersau:
Currently we only specify the version here
[...]

PR #1303, opened 2020-03-26
add .npmrc engine-strict=true

campersau:
might show warning during installation when node version is too low, not sure though.

see https://github.com/FredrikNoren/ungit/issues/1302


campersau:
Looks like this will only work during development. During install only a warning is shown...

PR #1311, opened 2020-04-16
Only display ref search button when there are hidden refs

jung-kim:
Old version that displays ref searching magnifying glasses button at each nodes.
Screen Shot 2020-04-15 at 22 24 25

Now only displays ref searching magnifying glasses button when a node has hidden refs
Screen Shot 2020-04-15 at 22 23 41


Hirse:
I like this change, though your commit contains a lot of whitespace change that makes it hard to see the exact diff.

jung-kim:
Yeah that one is bit annoying... I will revert irrelevant changes

PR #1314, opened 2020-04-22
Git: run detached

wmertens:
This prevents git hanging for asking ssh passphrases, like when you run it inside vscode with an empty ssh-agent.

Now when I run `SSH_AUTH_SOCK= npm start`, it immediately errors in the web UI instead of hanging.


ylecuyer:
Ahah nice I often have this error ;)
Though I think the `detached: false` was here on purpose ? I hope there is a test for it so we can see what this change breaks :p

campersau:
It was added in https://github.com/FredrikNoren/ungit/pull/1192 but I am not sure what issues @jung-kim had.
> fix issue with detached git processes on some OS and timeout not being enforced.

campersau:
I am not sure if `detached: true` is required when we also always close the `stdin`.

I have looked at vscode (they also use git via spawn) and they use [`stdio = ['ignore', null, null]`](https://github.com/microsoft/vscode/blob/888f89b40a4a1859b713de610ccd0ee912099521/extensions/git/src/git.ts#L537) to prevent redirecting `stdin` which I like a bit better.

wmertens:
I only added the `detached: true` when closing `stdin` wasn't sufficient. In VS Code I have the same problem, actually. I should open an issue for that 😅

jung-kim:
Hmm, I understand the issue but I don't think [`detached: true`](https://nodejs.org/api/child_process.html#child_process_options_detached) means that that hang process will continued to be hanged even after parents process has exited.

I don't think that is a good idea as there maybe just hung threads running without user's awareness, especially even after ungit process is dead.  

There was a PR or code changes I've done long time ago around doing timeouts on git processes but I think I've lost it by now.  I think something similar might be a better solution

wmertens:
Ok, so if it runs detached there's a chance it hangs for some reason other than waiting for a password and then it won't get killed? Doesn't the timeout work for that? (Didn't check)

jung-kim:
Yes it can cause orphaned threads.

I could depends on the implementations.

wmertens:
In the docs it says:

> By default, the parent will wait for the detached child to exit. To prevent the parent from waiting for a given subprocess to exit, use the subprocess.unref() method. Doing so will cause the parent's event loop to not include the child in its reference count, allowing the parent to exit independently of the child, unless there is an established IPC channel between the child and the parent.

Closing stdin is not enough, it will still prompt me for keys when I run `SSH_AUTH_SOCK= npm run start`

PR #1316, opened 2020-04-23
Use Prettier

wmertens:
It's very nice to not have to worry about formatting. Most of the time the result is very acceptable, and by formatting on save in editors that support it, you never have to spend time on getting the spacing right.

It even helps with spotting bugs, because the formatting will show nesting errors.


ylecuyer:
If this hasn't changed recently, the project has already jshint taking care of this.

wmertens:
Ah - I didn't notice.

In any case, as far as i know, prettier is the only formatter that formats the way it does, other formatters just follow some rigid rules, and prettier optimizes for legibility and clean diffs.

It also formats markdown, html etc

.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "useTabs": false,
+  "semi": true,
+  "singleQuote": true,
+  "trailingComma": "es5",
+  "bracketSpacing": true
```
Hirse, line 6: I'd prefer only specifying options that are not the default (https://prettier.io/docs/en/options.html).
```suggestion
  "singleQuote": true
```

CHANGELOG.md
```diff
@@ -1,27 +1,32 @@
 # Change Log
+
```
Hirse, line 2: Personally, I don't like this change, as it makes it harder to match headers to content.

In fact, I think it might be best to disable prettier for markdown.

.prettierrc
Hirse: I might also prefer to set the `printWidth` to 100 to make the code a bit more condensed.

.prettierignore
```diff
@@ -0,0 +1,15 @@
+.git/
+node_modules/
```
Hirse, line 2: As I understand, `node_modules` are always ignored:
https://prettier.io/docs/en/cli.html#--with-node-modules

```suggestion
```

wmertens:
changes applied. I decided against the pragma mode because it means adding a header to all files. If there's a problem, we can override it on a single directory.

Should the jshint config change?

wmertens:
oh, and do we want `(err) => ...` or `err => ...`? I've seen both styles.

Personally, I prefer tabs, no semicolumns, any chance of that happening? ;)

wmertens:
Ah, I see that jshint errors in the tests.

What would be the sentiment around switching to eslint? I believe it's currently the more maintained one (just an impression) and it integrates with prettier.

wmertens:
🤔 looks like prettier on Windows always fails, the other tests just time out sometimes.
Possibly line endings?

wmertens:
Ok that was it. Now the only test failures are random timeouts

package.json
```diff
@@ -6,8 +6,9 @@
   "ungitPluginApiVersion": "0.2.0",
   "scripts": {
     "start": "node ./bin/ungit",
-    "test": "grunt test",
+    "test": "prettier --check . && grunt test",
```
Hirse, line 9: Currently all tooling is run through grunt and while I think it might be good to switch to npm scripts in general, that might be a different change.

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "arrowParens": "avoid",
+  "endOfLine": "lf",
```
Hirse, line 3: This seems to be `lf` by default:
https://prettier.io/docs/en/options.html#end-of-line

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+  "arrowParens": "avoid",
```
Hirse, line 2: Personally, I would also go with the default here as I tend to agree with their rationale:
> At first glance, avoiding parentheses may look like a better choice because of less visual noise. However, when Prettier removes parentheses, it becomes harder to add type annotations, extra arguments or default values as well as making other changes. Consistent use of parentheses provides a better developer experience when editing real codebases, which justifies the default value for the option.
https://prettier.io/docs/en/options.html#arrow-function-parentheses

package.json
wmertens: my grunt is rusty, any quick hints for how to achieve this?

.prettierrc
wmertens: I needed to add this for the tests to pass on windows, looks like there's a bug in prettier

.prettierrc
wmertens: I think the main argument here is type annotations, which we don't have. I personally prefer less visual noise, but of course whatever is preferred in the project goes.

wmertens:
rebased, moved test into grunt, reverted arrowParens to default

package.json
jung-kim: I would recommend looking at `Gruntfile.js` and [grunt-prettier](https://www.npmjs.com/package/grunt-prettier) package

jung-kim:
Should prettier replace existing `js-hint`?

I admit, I've never done js professional so I feel I'm always behind on "new standard" but it seems that prettier is supersedes js-hint which superseded js-lint.

Hirse:
@jung-kim Not quite.
JSHint and JSLint do "linting", i.e. static analysis, finding of potential errors in the code (consider unused variables and parameters) as well as general best practices (consider issues like using `var` instead of `let`/`const`.)
Both are superseded by ESLint.

Prettier on the other hand is an automatic code formatter, focusing on code style (spaces around operators, etc).

Both (types of tools) can work well together when properly configured.

wmertens:
As for linting: I'll gladly do an eslint one later, and in the nodegit branch I have the configuration to make VS Code use Intellisense on the JS files, which is great for detecting typing errors and method/argument completion.

.prettierrc
campersau: So prettier fixes line endings as well?
Then we can remove [`lineending:production`](https://github.com/FredrikNoren/ungit/blob/6009c23e5e9cce78416f9d9ee55fcebdbb8a3920/Gruntfile.js#L424) and [`grunt.loadNpmTasks('grunt-lineending');`](https://github.com/FredrikNoren/ungit/blob/6009c23e5e9cce78416f9d9ee55fcebdbb8a3920/Gruntfile.js#L412) from the gruntfile.

.gitattributes
```diff
@@ -1,3 +1,3 @@
-* text=auto
+* text=auto eol=lf
```
campersau, line 1: For reference: https://github.com/prettier/prettier/issues/7853#issuecomment-603843422

package.json
```diff
@@ -6,7 +6,7 @@
   "ungitPluginApiVersion": "0.2.0",
   "scripts": {
     "start": "node ./bin/ungit",
-    "test": "prettier --check . && grunt test",
+    "test": "grunt test",
```
campersau, line 9: This change is currently in the [`prettier` commit](b8d381eab90991cea814badc4ebc7ef1030ce256). It would be better to move it into a separate commit instead, because the `prettier` commit is going to require some rebases / redos in the future.

Gruntfile.js
```diff
@@ -402,6 +402,21 @@ module.exports = (grunt) => {
     electronPackager(this.options()).then(() => { done(); }, done);
   });
 
+  grunt.registerTask('checkPrettier', 'Verify that all files are correctly formatted.', function() {
+    const done = this.async();
+    childProcess.exec('prettier -l .', (err, stdout, stderr) => {
+      const files = stdout.trim();
+      if (files) {
+        return done(new Error(`Files with incorrect formatting:\n${files}\n`))
```
campersau: Maybe we can also add a hint to this message like "Use `npm run format` to fix them automatically."

package.json
wmertens: Oops, good catch :sweat_smile:

wmertens:
Changes applied. If you prefer, I can leave off the `prettier` commit so it will be easier to merge.

Another option (reasonably nuclear) is to rewrite the whole repo history with prettier. If the forks do the same then the PRs can remain as-is, but formatted. It takes quite a while to run though:

```
git filter-branch --tree-filter 'git diff --cached --name-only --diff-filter=d -z | grep -z -E ".(js|jsx|ts|tsx|json)$" | \
    xargs -t -0 prettier --config '"$PWD"'/../prettierrc --ignore-path '"$PWD"'/../prettierignore --write || true' -- --all
```

package.json
wmertens: @jung-kim I ended up doing it in plain grunt, would the grunt-prettier package add anything?

package.json
```diff
@@ -8,6 +8,7 @@
     "start": "node ./bin/ungit",
     "test": "grunt test",
     "unittest": "grunt unittest",
+    "format": "prettier --write .",
```
campersau, line 11: I have tested it a little bit more and it seems that `./bin/credentials-helper` and `./bin/ungit` are not formatted.
Manually including them like this fixed it:

```suggestion
    "format": "prettier --write . ./bin/credentials-helper ./bin/ungit",
```

Gruntfile.js
```diff
@@ -427,7 +442,7 @@ module.exports = (grunt) => {
   // Run tests without compile (use watcher or manually build)
   grunt.registerTask('unittest', ['mochaTest:unit']);
   grunt.registerTask('clicktest', ['mochaTest:click']);
-  grunt.registerTask('test', ['unittest', 'clicktest']);
+  grunt.registerTask('test', ['checkPrettier', 'unittest', 'clicktest']);
```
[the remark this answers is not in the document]
campersau: Since this is now in the `default` task I think this can be remove from here. I don't like being interruppted by code style issue when modifying / debugging tests :)

wmertens:
@campersau 👍 done

wmertens:
Should I add a prettier commit so people can see the result?

campersau:
Yes, you can. Please run it twice!

package.json
Hirse: Is that because they are not `.js` files?
If so, what is the main reason for them to not have the `.js` ending?

.prettierignore
```diff
@@ -0,0 +1,12 @@
...
+assets/
+package-lock.json
+package.json
+
+public/css/
+public/js
+public/vendor/
+build/
+**/*.bundle.js
+**/*.md
```
Hirse, line 12: Should this also include `**/*.css`?

The components' less files get transformed to CSS, but we don't want to prettify those.

Gruntfile.js
```diff
@@ -188,6 +177,21 @@ module.exports = (grunt) => {
     }
   });
 
+  grunt.registerTask('checkPrettier', 'Verify that all files are correctly formatted.', function() {
```
Hirse, line 180: Nit: Current task naming convention seems to be `kebab-case`.

```suggestion
  grunt.registerTask('check-prettier', 'Verify that all files are correctly formatted.', function() {
```

Gruntfile.js
```diff
@@ -188,6 +177,21 @@ module.exports = (grunt) => {
...
+    const done = this.async();
+    childProcess.exec('prettier -l . bin/*', (err, stdout, stderr) => {
```
Hirse, line 182: I would prefer using the [prettier check API](https://prettier.io/docs/en/api.html#prettierchecksource--options) instead of childProcess.

Also, if we are using a command anyway, I do not see much value by adding it to Grunt.

Gruntfile.js
```diff
@@ -188,6 +177,21 @@ module.exports = (grunt) => {
...
+      const files = stdout.trim();
+      if (files) {
+        return done(new Error(`Files with incorrect formatting (run "npm run format" and consider a Prettier plugin for your editor):\n${files}\n`))
+      }
+      if (err) {
+        console.error(stderr)
```
Hirse, line 188: Use `grunt.log` instead of console.

Also, the line is missing a `;`. 

```suggestion
        grunt.log.error(stderr);
```

Hirse:
After this PR (in its current state) I can see several different pattern that we might want to align:
1. Grunt plugins
  less, watch, jshint
2. Custom Grunt tasks calling JS APIs 
  browserify (instead of grunt-browserify)
3. Custom Grunt tasks calling CLI APIs
  prettier check
4. Calling CLI APIs from npm scripts
  prettier format

What is the general idea for calling build steps/scripts?

Given the slowed down (or discontinued) maintenance of Grunt plugins, we might want to reduce their usage. Should we also migrate away from Grunt itself?

While this is not directly related to prettier, I wanted to mention it here, because this PR adds two more entries to my list above.

Gruntfile.js
campersau: We currently already have different naming conventions. e.g.
`clickParallel`
`travisnpmpublish`
`browserify-common`
So I don't mind because we might change that later.

Gruntfile.js
campersau: Yeah, I just tried to run `grunt` but it didn't work because I don't have `prettier` installed globally. Only `npm run build` works now.

wmertens: While I agree that that is the cleaner way of going about it, it is also a lot more work and I just want to get code formatted :)

On top of that, it's not clear that Grunt will stay in use. Personally, I havent used Grunt since 2014, and instead I rely on Webpack and nps for my entire workflow. Grunt is just so much more complex.

And even if Grunt will stay in use, eslint is imho better than jshint, and it has prettier support for checking and fixing.

I added `npx` to the grunt command so that it will use the installed prettier.

wmertens:
@Hirse So I think webpack is more powerful than browserify. I have a setup where the dev server automatically updates the server (hot-patching the endpoints, reloading if not possible) and the client side (uses React to do the same), all from running a single command.

The prod build converts only the files needed for the final bundle to the formats needed, thanks to require-based dependencies and pluggable loaders.

This removes 90% of what grunt does, and does it as fast or faster. The other 10% are maintenance scripts that are annoying to write in Grunt; I instead run them with https://github.com/sezna/nps

Hirse:
@wmertens I agree that webpack is powerful enough to handle most cases, but switching from browserify to webpack has lots of other consequences for the current plugin-infrastructure.

As a first step, replacing grunt with individual scripts (whether running them directly or through nps) would be good.

PR #1318, opened 2020-04-24
Fixing some accessibility issues

Hirse:
* Adding alt text to logo and gravatar
* Adding (invisible) labels to dropdown buttons
* Adding `lang="en"` to `<html>`
* Adding aria-labels for `input`s without label
* Removing placeholders when they are identical to label
* Adding `tabindex` to dialog to properly add tabbing behavior
* Use bootstrap alert to display login errors

The remaining big accessibility thing would be contrast between font-color and background, but that would introduce very visible changes.

### DIFF
--- a/components/dialogs/dialogs.js
+++ b/components/dialogs/dialogs.js
@@ -119,8 +119,8 @@ class YesNoDialogViewModel extends PromptDialogViewModel {
     this.result = ko.observable(false);
     const self = this;
     this.alternatives([
-      { label: 'Yes', style: 'primary', taId: 'yes', click() { self.result(true); self.close(); } },
-      { label: 'No', style: 'default', taId: 'no', click() { self.result(false); self.close(); } },
+      { label: 'Yes', primary: true, taId: 'yes', click() { self.result(true); self.close(); } },
+      { label: 'No', primary: false, taId: 'no', click() { self.result(false); self.close(); } },
     ]);
   }
 }
@@ -132,9 +132,9 @@ class YesNoMuteDialogViewModel extends PromptDialogViewModel {
     this.result = ko.observable(false);
     const self = this;
     this.alternatives([
-      { label: 'Yes', style: 'primary', taId: 'yes', click() { self.result(true); self.close(); } },
-      { label: 'Yes and mute for awhile', style: 'mute', click() { self.result("mute"); self.close() } },
-      { label: 'No', style: 'default', taId: 'no', click() { self.result(false); self.close(); } }
+      { label: 'Yes', primary: true, taId: 'yes', click() { self.result(true); self.close(); } },
+      { label: 'Yes and mute for awhile', primary: false, taId: 'mute', click() { self.result("mute"); self.close() } },
+      { label: 'No', primary: false, taId: 'no', click() { self.result(false); self.close(); } }
     ]);
   }
 }
@@ -146,8 +146,8 @@ class TooManyFilesDialogViewModel extends PromptDialogViewModel {
     this.result = ko.observable(false);
     const self = this;
     this.alternatives([
-      { label: "Don't load", style: 'primary', taId: 'noLoad', click() { self.result(false); self.close(); } },
-      { label: 'Load anyway', style: 'default', taId: 'loadAnyway', click() { self.result(true); self.close(); } },
+      { label: "Don't load", primary: true, taId: 'noLoad', click() { self.result(false); self.close(); } },
+      { label: 'Load anyway', primary: false, taId: 'loadAnyway', click() { self.result(true); self.close(); } },
     ]);
   }
 }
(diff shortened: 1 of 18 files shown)


components/login/login.html
```diff
@@ -1,26 +1,25 @@
...
     <!-- ko if: loginError -->
-    <div class="loginError" data-bind="text: loginError"></div>
+    <div class="alert alert-danger loginError" role="alert" data-bind="text: loginError"></div>
     <!-- /ko -->
-    <form data-bind="submit: login" role="form">
+    <form data-bind="submit: login">
 
       <div class="form-group">
         <label for="inputUsername">Username</label>
-        <input type="text" class="form-control" id="inputUsername" placeholder="Username" data-bind="value: username">
```
campersau, line 15: Removing these `type="text"` attributes breaks the clicktests. ([Example](https://github.com/FredrikNoren/ungit/blob/e6fa906db2a7b1c8a8bb4038294bb36f9fe94a75/nmclicktests/spec.authentication.js#L17-L31))
Was there a particular reason to remove them other then relying on default browser behavior?
I personally like to be explicit in these cases especially for `<button>` `type`s.

campersau:
Chrome Accessibility Audit:

Before:
image

After:
image

components/login/login.html
Hirse: Yes, mostly to reduce the number of attributes.

components/dialogs/formDialog.html
```diff
@@ -1,26 +1,24 @@
...
+        <div class="modal-body" data-bind="foreach: items">
+          <div class="form-group">
+            <label data-bind="text: name, attr: { for: name }" class="col-lg-2 control-label"></label>
+            <div class="col-lg-10">
+              <input class="form-control" data-bind="value: value, attr: { id: name, type: type, autofocus: autofocus }" />
+            </div>
+          </div>
+        </div>
+        <div class="modal-footer">
+          <button class="btn btn-primary" type="submit">Submit</button>
```
campersau, line 18: More test failures: https://github.com/FredrikNoren/ungit/pull/1318/checks?check_run_id=615722699#step:21:418

Hirse: Thanks. Updated.

components/login/login.html
Hirse: Reverted the removed `type`s.

components/dialogs/dialogs.js
```diff
@@ -132,9 +132,9 @@ class YesNoMuteDialogViewModel extends PromptDialogViewModel {
     this.result = ko.observable(false);
     const self = this;
     this.alternatives([
-      { label: 'Yes', primary: true, taId: 'yes', click() { self.result(true); self.close(); } },
-      { label: 'Yes and mute for awhile', primary: false, taId: 'mute', click() { self.result("mute"); self.close() } },
-      { label: 'No', primary: false, taId: 'no', click() { self.result(false); self.close(); } }
+      { label: 'Yes', style: 'primary', taId: 'yes', click() { self.result(true); self.close(); } },
+      { label: 'Yes and mute for awhile', style: 'mute', click() { self.result("mute"); self.close() } },
```
campersau, line 136: The default `btn btn-default` classes changed to only `btn` [here](https://github.com/FredrikNoren/ungit/pull/1318/files#diff-474b93f945d2a9ad29f677e433974d0eR12) which now looks like this:

Before:
image

After:
image

Was this intended?

Hirse: Good catch.

I didn't realize `btn-muted` was not a style class.
Changed it now to separate style classes from action/testing ids.

nmclicktests/spec.discard.js
```diff
@@ -8,20 +8,20 @@ const createAndDiscard = (env, testRepoPath, dialogButtonToClick) => {
...
+        return env.nm.ug.click('.modal-dialog [data-ta-action="no"]');
       } else {
-        return env.nm.visible('.modal-dialog .btn-primary')
+        return env.nm.visible('.modal-dialog [data-ta-action="yes"]')
           .then((isVisible) => { if (isVisible) throw new Error('Should not see yes button'); });
       }
     }).then(() => {
       if (dialogButtonToClick !== 'no') {
-        return env.nm.ug.waitForElementNotVisible('.files .file .btn-default');
+        return env.nm.ug.waitForElementNotVisible('.files .file [data-ta-action="no"]');
```
campersau, line 22: This was correct because it checked if the file isn't in the staging area anymore.

nmclicktests/spec.discard.js
```diff
@@ -8,20 +8,20 @@ const createAndDiscard = (env, testRepoPath, dialogButtonToClick) => {
...
       } else {
-        return env.nm.wait('.files .file .btn-default');
+        return env.nm.wait('.files .file [data-ta-action="no"]');
```
campersau, line 24: This was correct because it checked if the file is in the staging area.

PR #1328, opened 2020-04-27
Clean up arrow styling

Hirse:
The only visual change is that the arrows now use the exact color of the container instead of a very close one.


Hirse:
@campersau Would you mind reviewing this PR?

PR #1329, opened 2020-04-27
Restore missing new line in package.json

ylecuyer:
Since https://github.com/FredrikNoren/ungit/commit/613699d0224858908dab96ea8a5dac985de189ac I always have this diff as `npm install` always adds back this new line which has been removed


campersau:
Yes, I noticed it too. Would have also been automatically fixed by https://github.com/FredrikNoren/ungit/pull/1309

PR #1330, opened 2020-04-27
Fix for commit component with empty args object

bakkerthehacker:
Flatten total-lines-changed object to allow the commit object to not error when passed an empty object.

### DIFF
--- a/components/commit/commit.js
+++ b/components/commit/commit.js
@@ -55,8 +55,8 @@ class CommitViewModel {
     this.authorDateFromNow(this.authorDate().fromNow());
     this.authorName(args.authorName);
     this.authorEmail(args.authorEmail);
-    this.numberOfAddedLines(args.total.additions);
-    this.numberOfRemovedLines(args.total.deletions);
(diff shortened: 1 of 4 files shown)


campersau:
Under which circumstances an empty object would be passed in? How can I reproduce this?

bakkerthehacker:
I have had this issue occur a few times on a repo.  Not sure if theres something weird going on that repo or if my diff tools are interfering with ungit, but I have gotten crashes in commit.js with `total` being undefined.  

I am unable to determine exactly what causes this or how to reproduce it.

bakkerthehacker:
Thanks for fixing this, I did manage to get it to happen again:
Screenshot from 2020-05-04 11-39-39

campersau:
Which version of ungit do you use?
Can you inspect the network trace for /gitlog requests and paste the json here?
Or you can try logging it here manually via console.log:
https://github.com/FredrikNoren/ungit/blob/51bd6d957aba2e6a92003bbcf53b1190a1e5d0fe/source/git-promise.js#L449

PR #1331, opened 2020-04-28
Clean up "Load more"-button

Hirse:
* Consolidate styles in `textdiff.less`
* Switch button style from `warning` to `default`
* Use actual `<button>` instead of `<span>`


campersau:
Before / After
image

PR #1332, opened 2020-04-28
Organizes component styles

Hirse:
- Move styles specific to a component from `styles` and `generic` files to component `less` files
  - [app](https://github.com/FredrikNoren/ungit/commit/32a0ff5518627dd817fff96443329f4b51416397)
  - [crash](https://github.com/FredrikNoren/ungit/commit/8cce359dca116a458585839c35005795b822a2b9)
  - [graph](https://github.com/FredrikNoren/ungit/commit/6e873ac7a28d52bafb4b4023eb5ffb61542bb10c)
  - [repository](https://github.com/FredrikNoren/ungit/commit/ff7816ef71af34229239f8eeace5568696e9904d)
  - [staging](https://github.com/FredrikNoren/ungit/commit/cab2fa0d3a900ec02c665d7a295cb330aaffef6e)
- Remove unused style rules from `generic.less`
- Format `generic.less

### DIFF
--- a/public/less/generic.less
+++ b/public/less/generic.less
@@ -24,73 +24,41 @@
 }
 
 // -- Bootstrap styling ---
-
-.btn, .list-group-item {
-	position: relative;
-	& > .progress {
-		position: absolute;
-		left: 0px;
-		top: 0px;
-		width: 100%;
-		height: 100%;
-		background: transparent;
-		pointer-events:none;
-		.progress-bar {
-			pointer-events:none;
-		}
-	}
-}
-.btn-group {
-	.btn > .progress {
-		border-radius: 0px;
-		.progress-bar {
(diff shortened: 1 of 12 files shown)


public/less/generic.less
```diff
@@ -24,84 +24,41 @@
...
-}
-
 .panel {
-	border: 0px;
+  border: none;
 }
 
 .form-control {
-	border: 0px;
-	color: rgba(255, 255, 255, 0.8);
```
campersau, line 58: This breaks many input text colors:
image

image

Hirse: Good catch.

I added a line to `variables.less` to set the input color and forgot to include that file.
Should be fixed with the latest commit.

PR #1336, opened 2020-04-28
Migrate clicktests from nightmare to puppeteer

campersau:
Nighmare hasn't been updated for a while now and the chrome version is getting pretty old missing some features like `Promise.prototype.finally`.

Tests run fine on my machine but in CI they seem a bit flaky and need a bit more work. Might be because I removed a lot of `wait`s and puppeteer runs a lot faster and ungit sometimes can't keep up.

~~It currently does [not work in node 14](https://github.com/puppeteer/puppeteer/issues/5719) but that's because of [this node issue](https://github.com/nodejs/node/pull/32968) which should be [released in a few hours](https://github.com/nodejs/node/pull/33103).~~

Also using async / await!


.github/workflows/ci.yml
```diff
@@ -31,6 +31,8 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
       - run: sudo apt update
         if: matrix.os == 'ubuntu-latest'
+      - run: sudo apt install -y libgbm1
```
campersau, line 34: https://github.com/puppeteer/puppeteer/commit/5c839f5e48dffd57cbf740f4a980b154682a2d87

.travis.yml
```diff
@@ -27,6 +27,7 @@ addons:
   apt:
     packages:
       - xvfb
+      - libgbm1
```
campersau, line 30: https://github.com/puppeteer/puppeteer/commit/5c839f5e48dffd57cbf740f4a980b154682a2d87

Gruntfile.js
```diff
@@ -69,7 +69,7 @@ module.exports = (grunt) => {
         options: {
           reporter: 'spec',
           require: './source/utils/winston.js',
-          timeout: 15000,
+          timeout: 35000,
```
campersau, line 72: puppeteer default timeout is 30000ms.

Gruntfile.js
```diff
@@ -153,6 +153,7 @@ module.exports = (grunt) => {
       },
       mocha: {
         options: {
+          esversion: 8,
```
campersau, line 156: async / await

components/stash/stash.html
```diff
@@ -10,7 +10,7 @@ <h4 class="stash-toggle-text" data-bind="click: toggleShowStash">
     </h4>
     <div class="list-group" data-bind="foreach: stashedChanges">
       <div class="list-group-item">
-        <a href="#" class="stash-apply octicon-circled pull-left bootstrap-tooltip" data-bind="html: applyIcon, click: apply" data-toggle="tooltip" title="Apply this stash"></a>
```
campersau, line 13: The clicktest did fail on macOS because this used floats and could be clicked. It now works by using `display: inline-block` which is nicer.

nmclicktests/environment.js
```diff
@@ -1,189 +1,16 @@
...
-const Bluebird = require('bluebird');
-const Nightmare = require('nightmare');
+const puppeteer = require('puppeteer');
 const net = require('net');
 const request = require('superagent');
-const mkdirp = require("mkdirp");
-const rimraf = Bluebird.promisify(require("rimraf"));
+const mkdirp = require('mkdirp');
+const util = require('util');
+const rimraf = util.promisify(require('rimraf'));
```
campersau, line 9: Also getting rid of Bluebird and using nodes promisify.

nmclicktests/environment.js
```diff
@@ -1,189 +1,16 @@
...
 const portrange = 45032;
-let rootUrl;
```
campersau, line 11: No global state!

nmclicktests/environment.js
```diff
@@ -193,107 +20,54 @@ const prependLines = (pre, text) => {
...
-            .then(() => this.ensureStarted());
-        }
+  async init() {
+    try {
+      this.browser = await puppeteer.launch({
+        defaultViewport: {
+          width: this.config.viewWidth,
+          height: this.config.viewHeight
+        },
+        headless: true
```
campersau, line 62: Change this to `false` to see the browser!

nmclicktests/environment.js
```diff
@@ -314,38 +88,225 @@ class Environment {
...
+        } else {
+          winston.info(text);
+        }
+      });
+    }
 
-    return Bluebird.resolve();
+    await this.page.goto(url);
+
+    if (!options || options.waitForSocketIO) {
```
campersau, line 226: Puppeteer is so fast that we have to manually wait for socket io to connect, otherwise this can lead to problems when code relies on working directory changes.

nmclicktests/spec.branches.js
```diff
@@ -3,140 +3,129 @@ const environment = require('./environment')();
...
   });
 
-  // https://github.com/segmentio/nightmare/issues/932
-  it.skip('search for the hidden branch', () => {
-    return environment.nm.wait(5000) // sleep to avoid `git-directory-changed` event, which refreshes git nodes and closes search box
-      .nm.click('.showSearchForm')
-      .wait(200)
-      .type('input.name', '-4\u0028\u000d')
-      .wait('[data-ta-name="search-4"]')
+  it('search for the hidden branch', async () => {
```
campersau, line 35: Enabled this test!

source/git-api.js
```diff
@@ -692,23 +692,33 @@ exports.registerApi = (env) => {
     });
     app.post(`${exports.pathPrefix}/testing/createfile`, ensureAuthenticated, (req, res) => {
       const content = req.body.content ? req.body.content : (`test content\n${Math.random()}\n`);
-      fs.writeFile(req.body.file, content, () => res.json({}));
+      fs.writeFileAsync(req.body.file, content)
+        .then(() => res.json({}))
+        .then(emitWorkingTreeChanged.bind(null, req.body.path));
```
campersau, line 697: To make the tests more reliable trigger working tree changes manually.

source/git-api.js
```diff
@@ -692,23 +692,33 @@ exports.registerApi = (env) => {
...
     });
     app.post(`${exports.pathPrefix}/testing/removefile`, ensureAuthenticated, (req, res) => {
-      fs.unlink(req.body.file, () => res.json({}));
+      fs.unlinkAsync(req.body.file)
+        .then(() => res.json({}))
+        .then(emitWorkingTreeChanged.bind(null, req.body.path));
     });
     app.post(`${exports.pathPrefix}/testing/git`, ensureAuthenticated, (req, res) => {
-      jsonResultOrFailProm(res, gitPromise(req.body.command, req.body.repo));
+      jsonResultOrFailProm(res, gitPromise(req.body.command, req.body.path));
```
campersau, line 721: This was the only case where the repro path variable name was `repo`. I just changed it to `path` to be consistent with the rest.

nmclicktests/environment.js
Hirse: Would it makes sense to make that configurable for local runs?

source/git-api.js
Hirse: I don't particularly like the `bind`. Would this work too?

```suggestion
        .then(() => emitWorkingTreeChanged(req.body.path));
```

Gruntfile.js
Hirse: Would the default be enough?

Gruntfile.js
Hirse: Why does JSHint have `esversion: 6`?
Does it not support anything higher?

.github/workflows/ci.yml
Hirse: Maybe add a comment so we know where this requirement comes from.

Also are there nightmare dependencies that can be removed now?

campersau: According to the docs puppeteer also requires xvfb
https://github.com/puppeteer/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-on-travis-ci

Gruntfile.js
campersau: Yes, I also could change the timeout to 15000 which we used before, but opted for sticking with the puppeteer defaults, that's why I have increased it here.

Gruntfile.js
campersau: Because that is what `esnext` was before https://jshint.com/docs/options/#esnext.

source/git-api.js
campersau: Yes, it would but I just want to keep the existing codestyle for now.
In the future we will replace all of this with async await anyway.

Gruntfile.js
Hirse: Ah, this is the mocha timeout.

I thought this is setting the puppeteer timeout to 35s instead of 30s and was wondering whether not setting it would be fine.

Gruntfile.js
Hirse: Makes sense, but can we set it to `esversion: 8` for the test files (and node files)?

.github/workflows/ci.yml
campersau: Looks like puppeteer doesn't need xvfb https://github.com/FredrikNoren/ungit/pull/1336/commits/3a3af0d6fc1447ec9b150a0d8aeed9d4f6621005

- [GitHub Actions Linux](https://github.com/FredrikNoren/ungit/pull/1336/checks?check_run_id=629647416)
- [Travis Linux](https://travis-ci.org/github/FredrikNoren/ungit/jobs/680990378)

Gruntfile.js
campersau: Yes, we will but not as part of this PR.

nmclicktests/environment.js
ylecuyer: Yes it would be nice, I use to set a `GUI=1` env variable for this

components/stash/stash.less
```diff
@@ -2,12 +2,14 @@
...
   margin-left: 20px;
   margin-right: 20px;
+  margin-bottom: -15px;
   background: #55323C;
   h4 {
     margin-top: 0px;
   }
-  margin-bottom: -15px;
-
+  .toggle-show-commit-diffs {
```
ylecuyer, line 10: is this related to some test failing now?

Gruntfile.js
```diff
@@ -69,7 +69,7 @@ module.exports = (grunt) => {
         options: {
           reporter: 'spec',
           require: './source/utils/winston.js',
-          timeout: 15000,
+          timeout: 35000,
           bail: true
         },
         src: 'nmclicktests/spec.*.js'
```
ylecuyer, line 75: Could you also remove the `nm` references ?

nmclicktests/environment.js
```diff
@@ -193,107 +20,54 @@ const prependLines = (pre, text) => {
 // Environment provides
 class Environment {
   constructor(config) {
-    this.nm = Nightmare({ Promise: Bluebird, typeInterval: 500, show: false });
     this.config = config || {};
     this.config.rootPath = (typeof this.config.rootPath === 'string') ? this.config.rootPath : '';
-    this.config.serverTimeout = this.config.serverTimeout || 15000;
+    this.config.serverTimeout = this.config.serverTimeout || 35000;
     this.config.viewWidth = 2000;
```
ylecuyer, line 26: btw any idea why this viewport is so big and square ?

Maybe we can give it more sense and switch to a FullHD viewport which represents better the reality ?

nmclicktests/environment.js
campersau: Yes, we could make them configurable via environment variables and document them because otherwise they are as hidden as this boolean.

I usually don't use environment variables to configure stuff like this, is `GUI=1` unique enough? Would it clash with some other programs? What about `GUI=true`?

components/stash/stash.less
campersau: No, the stash tests did fail before on macOS which is fixed by this.
The current test failures (mainly on macOS) are different, but I don't have a mac available so they are pretty hard to debug for me.

Gruntfile.js
campersau: Yes, I will probably rename this folder to just `clicktests` but I haven't checked if git is able to recognize this as a rename because of the big amount of changes within these files. Which is why I haven't done it yet to keep the changes as "clean" as possible for now.

nmclicktests/environment.js
campersau: Not sure why this was done in https://github.com/FredrikNoren/ungit/commit/c7cdddca9083be7c03acdc37d64763f2b7688658in https://github.com/FredrikNoren/ungit/pull/908
But I will try out if it works in 1920 × 1080.

components/stash/stash.less
campersau: Just want to note that macOS tests sometimes succeed but yes, they are currently the most flaky.
https://github.com/campersau/ungit/runs/639350614

nmclicktests/environment.js
campersau: Yes, they work in full HD!

Gruntfile.js
campersau: I changed it and indeed there are two many changes so the overall diff does not recognize the renames.

campersau:
First green build after a long time!
macOS is still a bit flaky though don't expect that it will always be green.

PR #1340, opened 2020-04-30
Reduce jQuery UI imports and use Bootstrap tooltips

Hirse:
This PR has two changes that are not easily possible to split.

Currently, ungit is (unintentionally probably) using the jQuery UI tooltips instead of the bootstrap ones.

1. Only import the jQuery UI widgets that are used, i.e. `autocomplete`
   - Switch from `jquery-ui-bundle` to `jquery-ui` package
   - Add a new file that managest the jQuery UI import
   - ~Remove jQuery UI styles import as we use our own (see #1327)~
   - Replace jQuery UI styles import with only core styles, thanks for the hint @ylecuyer 

2. Add custom styling to bootstrap tooltips
3. Fix placement of tooltips in button groups

---

image image

### DIFF
--- a/components/commit/commit.html
+++ b/components/commit/commit.html
@@ -12,11 +12,11 @@
             <span class="title" data-bind="text: (title().length > 72 ? title().substring(0, 72) + '...' : title)"></span>
             <span class="text-muted">by <a data-bind="text: authorName, attr: { href: 'mailto:' + authorEmail() }"></a></span>
             <!-- ko if: pgpVerifiedString() -->
-              <span class="text-muted" data-bind="html: pgpIcon, attr: { title: pgpVerifiedString() }" class="bootstrap-tooltip" data-toggle="tooltip" data-placement="top"></span>
+              <span class="text-muted bootstrap-tooltip" data-bind="html: pgpIcon, attr: { title: pgpVerifiedString() }" data-toggle="tooltip"></span>
             <!-- /ko -->
           </div>
           <div class="text-muted nodeSummaryContainer">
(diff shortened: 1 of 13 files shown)


Hirse:
I just replaced the jQuery UI package in the dependencies, but was wondering if it makes more sense to move it to devDependencies, since it is only used when building the bundle file.

campersau:
When selecting a directory in the autocomplete a text gets displayed at the bottom.
image

There is also a general bug where tooltips are not initialized until a git directory is opened if you like to fix that too.
image

Not quite sure about moving `jquery-ui` to `devDependencies` yet.

ylecuyer:
> When selecting a directory in the autocomplete a text gets displayed at the bottom.
> image

It seems you need to include core.css and autocomplete.css (cf https://www.drupal.org/project/aloha/issues/1827626#comment-6672476) I tried and it works well

ylecuyer:
> There is also a general bug where tooltips are not initialized until a git directory is opened if you like to fix that too.
> image

I have made https://github.com/FredrikNoren/ungit/pull/1343

Hirse:
> 
> 
> > When selecting a directory in the autocomplete a text gets displayed at the bottom.
> 
> It seems you need to include core.css and autocomplete.css (cf https://www.drupal.org/project/aloha/issues/1827626#comment-6672476) (https://github.com/jquery/jquery-ui/tree/c6f1735249295bb37153861aa97e5b3e38d46b02/themes/base) I tried and it works well

I was trying to avoid having to add any CSS from jQuery-UI.
I'll see if we can include (or copy) just the rules we need.
(Generally, I would not want to copy anything, but in this case we don't need to worry about future maintenance as jQuery-UI is not getting any updates.)

Hirse:
@campersau, fixed the issue as @ylecuyer suggested by adding the jQuery-UI core.css (back) to our CSS/Less.

PR #1341, opened 2020-04-30
Remove unused color dependency

Hirse:
Even though the `color` package is required and exposed during browserify, I could not find any usage of it.


ylecuyer:
Indeed is has been added here https://github.com/FredrikNoren/ungit/commit/f94bf1891f58c3d90d82f33b8f14562b679ced5b#diff-c99e00307f003584fd1eee55a0492474 a long time ago and is not used since

PR #1343, opened 2020-05-01
Fix: init tooltips from the app start

ylecuyer:
**Before**
Base on @campersau comment here : https://github.com/FredrikNoren/ungit/pull/1340#issuecomment-622092861

Tooltip are not loaded until a repo is opened

80755823-c7cd4b80-8b31-11ea-970d-9e9c7e892f68

**Fix**

Instead of relying on some hand made event to init tooltips, this PR uses the event-delegation method provided by bootstrap tooltip (based on jquery.on) to init tooltips on existing elements and new elements.

https://getbootstrap.com/docs/4.0/components/tooltips/

Screenshot_2020-05-01_23-36-05

### DIFF
--- a/public/source/main.js
+++ b/public/source/main.js
@@ -16,7 +16,7 @@ var navigation = require('ungit-navigation');
 var storage = require('ungit-storage');
 var adBlocker = require('just-detect-adblock');
 
-// Request animation frame polyfill
+// Request animation frame polyfill and init tooltips
 (function() {
   var lastTime = 0;
   var vendors = ['ms', 'moz', 'webkit', 'o'];
@@ -41,12 +41,9 @@ var adBlocker = require('just-detect-adblock');
       clearTimeout(id);
     };
 
-  programEvents.add(function(event) {
-    if (event.event === 'init-tooltip') {
-      $('.bootstrap-tooltip').tooltip().removeClass('bootstrap-tooltip');
-    }
+  $(document).tooltip({
+    selector: '.bootstrap-tooltip'
   });
-
 }());
 
 ko.bindingHandlers.autocomplete = {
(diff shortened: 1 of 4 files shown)


public/source/main.js
```diff
@@ -41,12 +41,9 @@ var adBlocker = require('just-detect-adblock');
       clearTimeout(id);
     };
 
-  programEvents.add(function(event) {
-    if (event.event === 'init-tooltip') {
-      $('.bootstrap-tooltip').tooltip().removeClass('bootstrap-tooltip');
```
ylecuyer, line 46: No need to remove the `bootstrap-tooltip` class anymore as the init process of new elements is handled by bootstrap and won't call it twice one existing tooltips

jung-kim: Honestly, I think we can get rid of `bootstrap` as it's footprint has been reduced.

Hirse: @jung-kim I'm not sure what you mean. 
Bootstrap is still used all over the place with tooltips, modals, dropdowns, and general styling.

public/source/main.js
```diff
@@ -41,12 +41,9 @@ var adBlocker = require('just-detect-adblock');
...
-    }
+  $(document).tooltip({
+    selector: '.bootstrap-tooltip'
```
Hirse, line 45: Instead of the class, can/should we use `[data-toggle="tooltip"]`?

I think the class was only used (and added) so there was something to distinguish the initialized from the uninitialized tooltips.

Of course, that change could be a separate PR

campersau: Yes, following the bootstrap standard `[data-toggle="tooltip"]` selector would be good.

PR #1344, opened 2020-05-02
Remove unused tracker.js

Hirse:
Usage of that file seems to have stopped with 9a1ffbb5413df9efb57c4fcf72e794b30233186f in 2014.


campersau:
Ignoring the whole `public/js/` directory breaks the build because the directory `js` does not exist anymore when newly checking it out. (See gruntfile.js `browserify-common`)

So the existence of the unused `tracker.js` file is currently keeping the build running 🤦

PR #1345, opened 2020-05-03
Image cleanup

Hirse:
## Updated Logo
* Align all letters on same baseline (73px from top)
* Spread letters so spacing is equal (21px)
* Align size of `u`, `n` and top part of `g` (124&times;112 px)
* Align width of `i` and `t` (55px)
* Change "cut" of the `g` to be straight on top so it aligns with the bottom of `u` and `n`
* Change shadow down and right offset to be the same (16px)
* Change all values to whole pixel (i.e. `51` instead of `50.75961`) to guarantee straight lines
* Clean up SVG structure

*Old and New*:
old new

*Compare (old in red, new in green)*:
compare

## Remove `imagemin` task
The grunt imagemin task had two goals, compressing and copying the images from `assets` to `public`. With size savings of around 500 bytes (!), the compressions is not worth performing.
```
                 icon.png      logo.png
uncompressed:    1911 bytes    4939 bytes
compressed:      1346 bytes    4482 bytes
new (png):        793 bytes    2108 bytes
new (svg):        498 bytes    1151 bytes
```
With the new logo (and icon) introduced above, the size is more than halved even without compression.

Therefore, this PR removes the task, the grunt-contrib-imagemin` dependency (and its many dependencies), and moves the uncompressed images to the `public/images` folder directly.

## Remove old, unused SVG assets
Most existing SVG assets are no longer used, some representing icons that got replaced by Octicons, others are alternative logo designs. Almost all of them were only viewable in Inkscape. For cleaned up SVGs, see bd0c29a836627bb488d7c6b06cd37ce081ab3ea5, for exported PNGs see below:
- `branchIcon.svg`
  branchIcon
- `icon.svg`
  icon
- `logo.svg`
  logo
- `logo2.svg`
  logo2
- `logo3.svg`
  logo3
- `logo4.svg`
  logo4
- `remoteIcon.svg`
  remoteIcon
- `tagIcon.svg`
  tagIcon
- `throbber.svg`
  throbber

## Other changes
- Update Icon based on logo
- Update `favicon.ico` based on icon
- Remove `logoLarge` as new logo is high resolution


Hirse:
@campersau This PR only touches icons/images used by the regular web usage.
Could you tell me which icons are used for the Electron build?

campersau:
@Hirse electron uses the icons in the root directory
- icon.icns (macOS) (I have created this with an online tool in https://github.com/FredrikNoren/ungit/pull/1240.)
- icon.ico (windows)
- icon.png (currently unused but I will use it soon, so it shows up on linux)

Hirse:
Would it be possible to use the icons in `public/images` for electron?

Or move all electron related files into a separate directory?

campersau:
Yes, the images can be referenced from some other directory, just update this line:
https://github.com/FredrikNoren/ungit/blob/09232a774f8d53b3e28db68453d93fdadaa20429/Gruntfile.js#L193
e.g. `./public/images/icon` the extension will be automatically added (e.g. `./public/images/icon.ico` on windows) by electron-packager

campersau:
I have merged #1347 so this line also needs to change if you put the icon in a different location.
https://github.com/FredrikNoren/ungit/blob/51bd6d957aba2e6a92003bbcf53b1190a1e5d0fe/public/main.js#L148

Hirse:
@campersau Updated the two paths you mentioned above, added higher quality icons, and removed the old ones.

PR #1348, opened 2020-05-03
remove window information from package.json

campersau:
This was added a long time ago  when node-webkit was used instead of electron. https://github.com/FredrikNoren/ungit/commit/68cd534c0e8a60125fbd513660485e3c41f614b5

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,17 +94,5 @@
     "node": ">=10.18"
   },
   "license": "MIT",
-  "main": "public/main.js",
-  "window": {
-    "title": "Ungit",
-    "icon": "icon.png",
-    "toolbar": false,
-    "frame": false,
-    "show": false,
-    "width": 1000,
-    "height": 600,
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -94,17 +94,5 @@
...
-    "toolbar": false,
-    "frame": false,
-    "show": false,
-    "width": 1000,
-    "height": 600,
-    "position": "center",
-    "min_width": 400,
-    "min_height": 200
-  }
+  "main": "public/main.js"
```
Hirse, line 97: Is `main` used by electron?

If not, as this is the [entrypoint when doing `require('ungit')`](https://docs.npmjs.com/files/package.json#main), does it make sense to point to a UI script?

campersau: Yes, it is used by electron: https://www.electronjs.org/docs/tutorial/first-app#writing-your-first-electron-app

PR #1350, opened 2020-05-06
Remove bluebird

campersau:
> [Promises in Node.js 10 are significantly faster than before. Bluebird still includes a lot of features like cancellation, iteration methods and warnings that native promises don't. If you are using Bluebird for performance rather than for those - please consider giving native promises a shot and running the benchmarks yourself.](https://github.com/petkaantonov/bluebird#note)

Browsers have also good support for them now and because of #1336 we can use some new APIs on the client like [`Promise.prototype.finally`](https://caniuse.com/#feat=promise-finally).

Also using async fs APIs in more places.


Gruntfile.js
```diff
@@ -357,14 +354,14 @@ module.exports = (grunt) => {
...
     });
 
-    Bluebird.all(bumps).then(() => {
-      fs.writeFileSync('package.json', `${JSON.stringify(tempPackageJson, null, 2)}\n`);
-      grunt.log.writeln('Dependencies bumped, run npm install to install latest versions.');
-    }).then(() => { done(); }).catch(done);
+    Promise.all(bumps)
+      .then(() => fs.promises.writeFile('package.json', `${JSON.stringify(tempPackageJson, null, 2)}\n`))
+      .then(() => grunt.log.writeln('Dependencies bumped, run npm install to install latest versions.'))
+      .then(() => { done(); }).catch(done);
```
Hirse, line 364: If we don't care about the result, we should consider using `finally`.

```suggestion
      .finally(done);
```

source/server.js
```diff
@@ -95,8 +94,8 @@ if (config.allowedIPs) {
     const ip = req.ip || req.connection.remoteAddress || req.socket.remoteAddress || req.connection.socket.remoteAddress;
     if (config.allowedIPs.indexOf(ip) >= 0) next();
     else {
-      res.status(403).send(403, '<h3>This host is not authorized to connect</h3>' +
-      '<p>You are trying to connect to an Ungit instance from an unathorized host.</p>');
+      res.status(403).send('<h3>This host is not authorized to connect</h3>' +
+        '<p>You are trying to connect to an Ungit instance from an unathorized host.</p>');
```
Hirse, line 98:
```suggestion
        '<p>You are trying to connect to an Ungit instance from an unauthorized host.</p>');
```

source/server.js
```diff
@@ -179,8 +178,8 @@ if (config.authentication) {
 
 const indexHtmlCacheKey = cache.registerFunc(() => {
   return cache.resolveFunc(pluginsCacheKey).then((plugins) => {
-    return fs.readFileAsync(__dirname + '/../public/index.html').then((data) => {
-      return Bluebird.all(Object.keys(plugins).map((pluginName) => {
+    return fs.readFile(__dirname + '/../public/index.html').then((data) => {
+      return Promise.all(Object.keys(plugins).map((pluginName) => {
         return plugins[pluginName].compile();
       })).then((results) => {
```
Hirse, line 184: I know it's not strictly related, but can we use `Object.values` here, since we don't seem to care about the plugin name.

```suggestion
      return Promise.all(Object.values(plugins).map((plugin) => plugin.compile)).then((results) => {
```

Hirse:
This looks very nice.
I especially like `fs.access`. 👍

Gruntfile.js
campersau: We don't care about the result but about the potential error in `.catch`.

Hirse: As far as I can see, the done function only accepts `false` to signal that the task has failed.

> Note that passing false to the done() function tells Grunt that the task has failed.

The following task for example succeeds because the done function is passed a truthy value:
```js
  grunt.registerTask('fail-async', '', function() {
    const done = this.async();
    Promise.reject("Failing from async")
      .then(() => { done(); })
      .catch(done);
  });
```
image


To fail on promise reject, we should do something like this:
```js
  grunt.registerTask('fail-async', '', function() {
    const done = this.async();
    Promise.reject("Failing from async")
      .then(done)
      .catch((reason) => {
        grunt.log.error(reason);
        done(false);
      });
  });
```

image

campersau: Ah, yes of course grunt uses an other behaviour than node...
https://gruntjs.com/creating-tasks#why-doesn-t-my-asynchronous-task-complete
I will revisit all usages of `this.async()` then.

campersau: Looks like it works as expected when using actual `Error` objects.
```js
  grunt.registerTask('fail-async', '', function() {
    const done = this.async();
    Promise.reject(new Error("Failing from async"))
      .then(() => { done(); })
      .catch(done);
  });
```

image

https://gruntjs.com/api/inside-tasks#this.async

PR #1351, opened 2020-05-07
Clean up browserify

Hirse:
- Only expose dependencies used by the components
- skip parsing of large libraries without dependencies

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -209,6 +209,7 @@ module.exports = (grunt) => {
     b.require('bluebird', { expose: 'bluebird' });
     b.require('blueimp-md5', { expose: 'blueimp-md5' });
     b.require('diff2html', { expose: 'diff2html' });
+    b.require('jquery', { expose: 'jquery' });
     b.require('knockout', { expose: 'knockout' });
     b.require('lodash', { expose: 'lodash' });
     b.require('./node_modules/snapsvg/src/mina.js', { expose: 'mina' });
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -230,30 +223,16 @@ module.exports = (grunt) => {
...
-                'ungit-storage',
-                'ungit-main',
-                'ungit-address-parser',
-                'knockout',
-                'lodash',
-                'hasher',
-                'crossroads',
-                'async',
-                'moment',
-                'blueimp-md5']);
```
Hirse, line 255: With `bundleExternal: false`, all external dependencies are skipped, so we don't need this list.

components/graph/git-node.js
```diff
@@ -1,4 +1,3 @@
-const $ = require('jquery');
```
Hirse, line 1: Removing the `require` as jQuery adds `$` as global anyway and we don't need to expose `jquery` from the bundle this way.

campersau: I am not sure if that is correct. We also require jquery here:
https://github.com/FredrikNoren/ungit/blob/f866756b2a651a85dbd9bf17802c881ed6b65d9e/public/source/main.js#L2-L3

Currently the tests are failing because the ref search autocomplete in `git-node.js` is not working because of this removal.

Hirse: Thanks for catching that. Reverted this change.

I somehow thought jQuery is used in many components yet only required in one, but I must have mistaken other `$`.

Issue #1333, opened 2020-04-28
ungit not showing master (or origin master) and crashes when checking out branch from master


wkjid10t:
This happened recently. Master branches not appearing anymore. Literally. They don't show up in the gui. You can check them out in CLI. Then ungit still won't show them. The top right tools show that master is selected. When you attempt to check out another branch, ungit crashes. 

Screenshots of crash:
![Screenshot_20200428_010509](https://user-images.githubusercontent.com/2594108/80449191-75e7b400-88ec-11ea-8f2d-7eec231a82ec.png)
![Screenshot_20200428_010156](https://user-images.githubusercontent.com/2594108/80449192-75e7b400-88ec-11ea-91f1-f1f528900f7e.png)

ylecuyer:
Hello could you also add the logs from the terminal where you launched
ungit. And the repo if it is not private.

wkjid10t:
There are no logs from terminal.

wkjid10t:
Is there a flag to output verbose/debug?

I can't share the repo. It's private.

wkjid10t:
Master branch showed up again after manually pulling from terminal. But it only shows commits that have been made since the pull. All previous commits are not rendered. It only renders one long line... I haven't scrolled to the bottom to see where it starts.

wkjid10t:
This is the root of my repo. The very first commit. There is something wrong with the graph. Is there cache that I can delete?

![Screenshot_20200429_182831](https://user-images.githubusercontent.com/2594108/80653110-6630b280-8a47-11ea-9d94-40b8383de5fe.png)

jung-kim:
two settings you can set at `~/.ungitrc`. 
```
{
logGitCommands: true,
logLevel: 'info'
}
```

also, you might be at the [headless state](https://medium.com/north-arrow-research/git-tips-what-is-a-headless-state-daa9bc4e1d5a).  Can you try doing `git checkout master` first and opening up ungit?

wkjid10t:
Not in headless state. Tried anyway, but did not fix. 

It looks like the graph is struggling with some of the merges. We had a bunch of merges back-to-back and it looks like since then, the graph just freaked out. I've tried uninstalling ungit, I've cloned the repo clean.

wkjid10t:
This seems like a similar issue (https://github.com/FredrikNoren/ungit/issues/641)

[...]

PR #1353, opened 2020-05-07
Add dev and prod build options

Hirse:
This PR adds a flag to the grunt build to switch between `dev` and `prod` mode.
The build is changed to compress and optimize the less and js bundles, buts adds inline source maps to the files in dev mode.

File size comparison after including #1350, #1351, and #1352 comparing to latest release:
```
                 1.5.7     prod     dev   
styles.css        191kb    129kb     474kb
ungit.js         4882kb    500kb    3506kb
components        357kb     96kb     665kb
component js      344kb     85kb     624kb
component css      13kb     11kb      41kb
```
*Note, that css only have sourcemaps in the new dev build.*

I tried to create source maps as separate files to be loaded when required, but couldn't get it to work correctly with less. (Source map path has to adapted to future server serving location.)

Running the browserify optimizations (from the [`tinyify`](https://www.npmjs.com/package/tinyify) plugin) increases the time that task takes significantly. For development, we might want to look into better watch-options, for example [`watchify`](https://www.npmjs.com/package/watchify), which in turn might be difficult to integrate into the current grunt workflow, especially since we are not using any grunt-browserify. (Longer term, a shift from grunt to npm scripts might be the best way. It already seems to be the way the community is going.)

@campersau @jung-kim @wmertens Would love some feedback on this.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -10,6 +10,9 @@ const fs = require('./source/utils/fs-async');
 const maxConcurrency = 5;
 
 module.exports = (grunt) => {
+  const dev = Boolean(grunt.option('dev'));
+  grunt.log.writeln(`Running in ${dev ? 'dev' : 'prod'} mode`);
+
   const packageJson = grunt.file.readJSON('package.json');
   const lessFiles = {
     'public/css/styles.css': ['public/less/styles.less', 'public/vendor/css/animate.css', 'public/less/d2h.less']
@@ -20,7 +23,15 @@ module.exports = (grunt) => {
   grunt.initConfig({
     pkg: packageJson,
     less: {
-      production: { files: lessFiles }
+      default: {
+        files: lessFiles,
+        options: {
+          compress: true,
+          sourceMap: dev,
+          sourceMapFileInline: true,
+          outputSourceFiles: true
+        }
+      }
     },
     watch: {
       scripts: {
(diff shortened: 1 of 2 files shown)


campersau:
Hm I am not sure about removing source maps in PROD. At least for javascript they are quite helpful because they would include the original filename and such in the stack trace. See https://github.com/FredrikNoren/ungit/issues/1333 
https://github.com/FredrikNoren/ungit/pull/1330#issuecomment-623541110 for recent issues which included the client side log.

So I would only disable `tinyify` in DEV and enable it in PROD and leave source maps always on.

Hirse:
Having the source maps inlined adds 3mb (a 600% increase) to the `ungit.js` bundle. I don't know how much we care about that since most users run on `localhost` anyway, but I would consider that significant.

We could investigate further whether having external source maps (i.e. in `.map` files) is the best solution. I don't know if they improve the stack trace sufficiently.

The rationale for minifying (i.e. running `tinyify`) in DEV mode was to have more comparable builds and prevent errors from originating from the build process itself (like dependencies that get optimized out in PROD builds).

wmertens:
In prod mode I always replace source maps with links, and depending on the project I serve them or not. I also use an error reporting endpoint that maps the errors using the sourcemaps.

Sending minified files for prod use is very useful, it really cuts down on load time. Parsing 3MB of javascript takes a long time.

Can you configure tinify to replace semicolons with returns? Uglify can do this, and the resulting file is equally small but it has way shorter lines, which makes debugging the minified code much simpler.

campersau:
External sourcemaps are now working for both `js` and `css` files. See https://github.com/FredrikNoren/ungit/pull/1394

PR #1357, opened 2020-05-11
Clear git-promise timeout when git command was successful

campersau:
While trying to migrate from grunt to script files, I noticed that the mocha unittest weren't existing right away. I tracked the problem down with [`wtfnode`](https://github.com/myndzi/wtfnode) to our git-promise timeout (introduced in https://github.com/FredrikNoren/ungit/pull/1192) which wasn't cancelled if the git call was successful and didn't timed out.

### DIFF
--- a/source/git-promise.js
+++ b/source/git-promise.js
@@ -52,8 +52,6 @@ const gitExecutorProm = (args, retryCount) => {
         if (config.logGitCommands)
           winston.info(`git executing: ${args.repoPath} ${args.commands.join(' ')}`);
         let rejectedError = null;
-        let isCompleted = false;
-        let timeoutTimer = null;
         let stdout = '';
         let stderr = '';
         let env = JSON.parse(JSON.stringify(process.env));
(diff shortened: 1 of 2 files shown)


ylecuyer:
Nice finding. Maybe we can add a test for it? Just checking that clearTimeout is called wdyt?

campersau:
I could add a test like this but hacking global variables for that feels a bit dirty and may have sideeffets when other code also uses it.

```js
const expect = require('expect.js');
const git = require('../source/git-promise');

describe('git-promise', () => {
    it('should clear git timeout when successful', async () => {
        let clearTimeoutCallCount = 0;
        const originalClearTimeout = clearTimeout;
        clearTimeout = (arg) => {
            clearTimeoutCallCount++;
            originalClearTimeout(arg);
        }

        try {
            await git(['version']);

            expect(clearTimeoutCallCount).to.equal(1);
        } finally {
            clearTimeout = originalClearTimeout;
        }
    });
});
```

source/git-promise.js
```diff
@@ -65,9 +66,10 @@ const gitExecutorProm = (args, retryCount) => {
         };
         const gitProcess = child_process.spawn(gitBin, args.commands, procOpts);
         if (args.timeout) {
-          setTimeout(() => {
+          timeoutTimer = setTimeout(() => {
             if (isCompleted) return;
```
Hirse, line 70: As far as I can see `isCompleted` is only used to check if the timeout callback has been called or is still required, which seems like a very similar use case to the new `timeoutTimer`.

Can you replace checking `isCompleted` with checking if `timeoutTimer` is not `null`?

campersau: Yes, it looks like it can be changed as suggested because we always set a timeout:
https://github.com/FredrikNoren/ungit/blob/70eb8246421658358d3ef08cfbe6a1974bd9b3b9/source/git-promise.js#L159

Not sure why the `if (args.timeout)` is there.

PR #1358, opened 2020-05-11
Async await

wmertens:
The result of running the async-await codemod and some fixes

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -258,20 +258,19 @@ module.exports = (grunt) => {
     ).then(this.async());
   });
 
-  const bumpDependency = (packageJson, packageName) => {
+  const bumpDependency = async (packageJson, packageName) => {
     const dependencyType = packageJson['dependencies'][packageName]
       ? 'dependencies'
       : 'devDependencies';
     let currentVersion = packageJson[dependencyType][packageName];
     if (currentVersion[0] == '~' || currentVersion[0] == '^')
       currentVersion = currentVersion.slice(1);
-    return pkgVersions(packageName).then((versionSet) => {
-      const versions = Array.from(versionSet);
-      const latestVersion = semver.maxSatisfying(versions, '*');
-      if (semver.gt(latestVersion, currentVersion)) {
-        packageJson[dependencyType][packageName] = '~' + latestVersion;
-      }
-    });
+    const versionSet = await pkgVersions(packageName);
+    const versions = Array.from(versionSet);
+    const latestVersion = semver.maxSatisfying(versions, '*');
+    if (semver.gt(latestVersion, currentVersion)) {
+      packageJson[dependencyType][packageName] = '~' + latestVersion;
+    }
   };
 
   grunt.registerTask(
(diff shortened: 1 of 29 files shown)


Gruntfile.js
```diff
@@ -332,32 +331,33 @@ module.exports = (grunt) => {
       .then((tests) => {
         grunt.log.writeln('Running click tests in parallel... (this will take a while...)');
         return Promise.all(
-          tests.map((file) => {
+          tests.map(async (file) => {
```
Hirse, line 334: The whole `clickParallel` task still has a bunch of `then`. Why is only this part rewritten?

components/branches/branches.js
```diff
@@ -129,17 +129,21 @@ class BranchesViewModel {
...
             remote: branch.isRemote ? branch.remote : null,
             name: branch.refName,
-          })
-          .then(() => programEvents.dispatch({ event: 'working-tree-changed' }))
-          .catch((e) => this.server.unhandledRejection(e));
+          });
+
+          return programEvents.dispatch({ event: 'working-tree-changed' });
+        } catch (e) {
+          return this.server.unhandledRejection(e);
```
Hirse, line 145: As [`unhandledRejection`](https://github.com/FredrikNoren/ungit/blob/70eb8246421658358d3ef08cfbe6a1974bd9b3b9/public/source/server.js#L180) doesn't seem to return anything, I would remove the `return` here;
```suggestion
          this.server.unhandledRejection(e);
```

components/path/path.js
```diff
@@ -96,11 +95,13 @@ class PathViewModel {
...
-  createDir() {
+  async createDir() {
     this.showDirectoryCreatedAlert(true);
-    return this.server
+
+    await this.server
       .postPromise('/createDir', { dir: this.repoPath() })
-      .catch((e) => this.server.unhandledRejection(e))
-      .then(() => this.updateStatus());
+      .catch((e) => this.server.unhandledRejection(e));
```
Hirse, line 103: Shouldn't this also be in a try/catch?

Hirse:
I started to review a bit, but the diff is so large that it is difficult to check everything.

In general, the codemod seems good, but it doesn't seem to remove all promises.
I'm a bit confused when `.then` is converted to `await` and when it isn't.

Hirse:
Also, by changing the UI code to use async/await, we are dropping support for all IE versions.

@campersau, @jung-kim, which browsers are currently supported or supposed to be supported?

campersau:
@Hirse we had this discussion some months ago over here https://github.com/FredrikNoren/ungit/pull/1178#issuecomment-552916409 
And as you can see we have dropped IE support a long time ago 😃 
I addition to that I removed [Bluebird](https://github.com/FredrikNoren/ungit/pull/1350) which means browsers also have to support promises and even [`Promise.prototype.finally`](https://caniuse.com/#feat=promise-finally) which was added later than [`async / await`](https://caniuse.com/#feat=async-functions)!

---

<sub>I have currently a lot of work at work so I will probably reviewing PRs at the weekend.</sub>

Hirse:
> @Hirse we had this discussion some months ago over here [#1178 (comment)](https://github.com/FredrikNoren/ungit/pull/1178#issuecomment-552916409)

Oh, right. 😄 
Thanks for the reminder.

I still think we should mention the supported browsers (in the Readme).

test/spec.git-api.js
```diff
@@ -377,51 +362,49 @@ describe('git-api', () => {
...
-      .then(() => common.get(req, '/gitignore', { path: testDir }))
-      .then((res) => expect(res.content).to.be('abc'))
-      .then(() => common.put(req, '/gitignore', { path: testDir, data: '' }))
-      .then(() => common.get(req, '/gitignore', { path: testDir }))
-      .then((res) => expect(res.content).to.be(''));
+  it('test gitignore api endpoint', async () => {
+    await common.put(req, '/gitignore', { path: testDir, data: 'abc' });
+
+    const res2 = await common.get(req, '/gitignore', { path: testDir });
+    await expect(res2.content).to.be('abc');
```
campersau, line 404: This is wrong!

wmertens:
I'll wait to revisit this until #1461 is merged, since it will cause a bunch of conflicts

PR #1360, opened 2020-05-13
Switch from JSHint to ESLint

Hirse:
- Remove JSHint
- Add ESLint
  - Add [prettier plugin](https://github.com/prettier/eslint-plugin-prettier)
  - Add [node plugin](https://github.com/mysticatea/eslint-plugin-node)
  - Add [mocha plugin](https://github.com/lo1tuma/eslint-plugin-mocha)
- Configure linting rules specific to folder using `.eslintrc.json` files
  - `clicktest`: mocha
  - `components`: browser
  - `public/source`: browser
  - `source`: node
  - `test`: mocha

  Using files in the folders allows [editor plugins](https://github.com/microsoft/vscode-eslint) to get the correct config when a file is opened
- Use ESLint to also check formatting
  With the prettier plugin, we can check for linting issues and formatting with one command
  - Removed check-prettier-formatting grunt-task and added `npm run lint` script
  - Add lint step to CI builds explicitly
- Fix reported errors
- Use [recommended configuration](https://eslint.org/docs/user-guide/configuring#using-eslint-recommended) (of ESLint and the plugins) with the following exceptions
  - [`no-unused-vars`](https://eslint.org/docs/rules/no-unused-vars): Allow unused parameter as that seems to be current code style
  - [`no-process-exit`](https://eslint.org/docs/rules/no-process-exit): Currently used intentionally
  - [`no-control-regex`](https://eslint.org/docs/rules/no-control-regex): We have to use control characters as they appear in messages from git
  - [`no-prototype-builtins`](https://eslint.org/docs/rules/no-prototype-builtins): Only [one occurrence](https://github.com/FredrikNoren/ungit/blob/master/public/source/server.js#L59) that we can refactor later
  - [`mocha/no-mocha-arrows`](https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-mocha-arrows.md): Mocha [discourages](https://mochajs.org/#arrow-functions) the use of arrow functions, which are used in the code base
  - [`mocha/no-setup-in-describe`](https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-setup-in-describe.md): Can be moved to `beforeAll` blocks later
  - [`mocha/no-identical-title`](https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/docs/rules/no-identical-title.md): Consider whether current naming is optimal

  The current configuration is aimed at providing the most value without requiring big refactorings for now. Future rules to consider might be:
  - [`no-var`](https://eslint.org/docs/rules/no-var): Require `const` or `let`
  - [...](https://eslint.org/docs/rules)


campersau:
When I run `npm run build` I get this message:
```
Running "browserify-components" task
>> ./components/.eslintrc.json/.eslintrc.json.js does not exist. If this component is obsolete, please remove that directory or perform a clean build.
```
You can also [see it in the CI builds](https://github.com/FredrikNoren/ungit/pull/1360/checks?check_run_id=669674969#step:20:22).

This is because we use `readdir` and assume that inside the `components` folder are only other folders and no files. So we need to change the logic a little bit.

https://github.com/FredrikNoren/ungit/blob/2a09c95dbc2eab0f726b5f61bcc1d9255827df5c/Gruntfile.js#L154-L163

---

Before we did run `jshint` as part of `build` now we have to manually run `npm run lint`. So we should either:
- Run `npm run lint && grunt` on build or
- Document that `npm run lint` must be run manually in the CONTRIBUTING.md

---

Also we now have two options to fix formatting errors:
- `npm run format`
- `npm run lint -- --fix`

Do we need both?
If yes should be change `format` to run `eslint . bin/* --fix` instead? Should we rename it to e.g. `lint-fix`?

---

This will be the output when there are linting errors:
```
PS P:\ungit> npm run lint

> ungit@1.5.7 lint P:\ungit
> eslint . bin/*


P:\ungit\Gruntfile.js
  350:5  error  Insert `;`  prettier/prettier

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! ungit@1.5.7 lint: `eslint . bin/*`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the ungit@1.5.7 lint script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

Hirse:
@campersau thanks for the detailed review.

- Updated the grunt task to filter out files in the `components` directory.
- Added documentation to `CONTRIBUTING.md`
- Changed `npm run format` to run `npm run lint -- --fix`

I had removed the linting from the build step to speed up build time and to not stop the build when there are formatting issues.
The linting is only really required before commit. During development linting or formatting issues should not stop the build.

We could consider adding the linting as pre-commit hook or adding it to the build in a non-blocking way (i.e. only issue warnings).

package.json
```diff
@@ -84,13 +89,13 @@
...
     "pkg-versions": "~2.1.0",
-    "prettier": "^2.0.5",
+    "prettier": "~2.0.5",
     "puppeteer": "~3.0.4",
     "superagent": "~5.2.2",
     "supertest": "~4.0.2"
   },
   "engines": {
-    "node": ">=10.18"
+    "node": "^10.18.0 || >=11.14.0"
```
campersau, line 98: Something I forgot to ask: why is this needed?

Hirse: `fs.promises` was first added in `11.14.0` and later backported to `10.17.0`.
We are fine to use it if we support latest Node 10 and latest Node 11.
`>= 10.18` would technically include `11.0.0` which does not support `fs.promises` yet.

See [`eslint-plugin-node` rule](https://github.com/mysticatea/eslint-plugin-node/blob/4a348e7d24e0d806d7873f293c91a922d0316d14/lib/rules/no-unsupported-features/node-builtins.js#L136-L141).

PR #1361, opened 2020-05-13
Remove ansi-color devDependency

Hirse:
The package is only used in one grunt task (parallel clicktests) to color output.
Grunt can achieve similar coloring [itself](https://gruntjs.com/api/grunt.log#grunt.log.ok-grunt.verbose.ok).

Additionally, when looking at the [ansi-color npm](https://www.npmjs.com/package/ansi-color) and [GitHub](https://github.com/loopj/commonjs-ansi-color) page I noticed that the license is [**not clearly Open Source**](https://github.com/loopj/commonjs-ansi-color/issues/13), so we might be better off without this particular package even if we want something to color console output.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,6 +1,5 @@
 const browserify = require('browserify');
 const childProcess = require('child_process');
-const cliColor = require('ansi-color');
 const electronPackager = require('electron-packager');
 const path = require('path');
 const pkgVersions = require('pkg-versions');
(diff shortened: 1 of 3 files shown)


campersau:
I am planning to move to npm scripts so this will be refactored to some standalone script where we don't have access to `grunt.log`.
But since we only have two references to `ansi-color` we might just use the escape sequences directly:
- `\x1b[32m green \x1b[0m`
- `\x1b[31m red \x1b[0m`

https://stackoverflow.com/a/41407246

campersau:
Superseded by #1370

PR #1362, opened 2020-05-14
Better error message when files are not formated

ylecuyer:
I don't like error messages giving me orders. This is a rewording of the error displayed when files are not properly formated


Hirse:
LGTM.

@ylecuyer would you mind trying out #1360 as that would replace the check in question with ESLint's output?

campersau:
As @Hirse mentioned this is now superseded by #1360

Issue #895, opened 2017-04-01
Remove Grunt


FredrikNoren:
Grunt is pretty dead by now. We could probably refactor the build script into something much simpler in just plain JS

codingtwinky:
I like how mocha testings, bumping up dependencies, transpiling and other actions have a single point of entry.  Do you really mean to get rid of Grunt? or replace with other solutions such as gulp or maybe even a make file?

FredrikNoren:
I think most people just do plain .js build files these days (no gulp, no makefile). With async/await it's pretty easy to do multi-step asynchronous operations.

codingtwinky:
What about babel? should we step away from that?

FredrikNoren:
Bable is only used to make sure the backend code is compatible with old node versions right? If we removed it, the min version would be 6? Or 7?

codingtwinky:
I think our code base will be at node 6 and considering client side code isn't es6 I think it is more safer.  Honestly, babel is something that is seemingly becoming an industry standard that I don't personally have a strong affinity for so I don't really mind it but I would like for us to understand that it means we are limiting backward compatibility if we decided to keep es6 on server side.

FredrikNoren:
Personally, I'm perfectly fine with only supporting the "live" versions of node; 6 is the current LTS. Either way, removing grunt doesn't necessarily mean we have to remove Babel

codingtwinky:
I'd be okay with just supporting node 5+ as well.  Hopefully there are not many complaints...

PR #1363, opened 2020-05-15
Move grunt tests to npm scripts

campersau:
Part of https://github.com/FredrikNoren/ungit/issues/895


CONTRIBUTING.md
```diff
@@ -82,7 +81,7 @@ Ungit has two major parts; the server and the UI. The server exposes a REST inte
 
 ### Running tests
 
-`grunt test` will run both unit tests, REST-interface tests, and click tests. `grunt unittest` only runs the tests in the test/ folder, `grunt clicktest` runs only the tests in the clicktests/ folder. Install mocha (`npm install -g mocha`) to run specific tests in the test folder and get better stack traces: `mocha test/spec.git-api.js`.
+`npm test` will run both unit tests, REST-interface tests, and click tests. `npm run unittest` only runs the tests in the test/ folder, `npm run clicktest` runs only the tests in the clicktests/ folder. Install mocha (`npm install -g mocha`) to run specific tests in the test folder and get better stack traces: `mocha test/spec.git-api.js`.
```
Hirse, line 84:
```suggestion
`npm test` will run both unit tests, REST-interface tests, and click tests. `npm run unittest` only runs the tests in the `test/` folder, `npm run clicktest` runs only the tests in the `clicktests/` folder. Install Mocha (`npm install -g mocha`) to run specific tests in the test folder and get better stack traces: `mocha test/spec.git-api.js`.
```

Hirse: Instead of a global mocha, can we use
1. `npx mocha test/spec.git-api.js` or
2. `npm run unittest -- test/spec.git-api.js`

or something like that?

campersau: The second option already works but needs `--no-config` so mocha does not not merge the spec files from the config file. But you would lose the the other settings like `timeout` and `file`:
`npm run unittest -- test/spec.git-api.js --no-config`

campersau:
No, this just covers the tests.
Yes, you can take over migrating coverage to npm scripts if you want.
In the end I want to move everything there to fix #895 doing it step by step.

PR #1364, opened 2020-05-16
Split up grunt less task

Hirse:
Split up the grunt less task into `common` and `components` following the `browserify` tasks for the following benefits:
- Clearer structure
- Nicer output
- More granular watching

*Output before*
image

*Output now*
image

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -15,17 +15,23 @@ module.exports = (grunt) => {
     .filter((component) => component.isDirectory())
     .map((component) => component.name);
 
-  const lessFiles = {
-    'public/css/styles.css': ['public/less/styles.less'],
-  };
-  components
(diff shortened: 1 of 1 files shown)


campersau:
My plan is to get rid of grunt and move stuff like that into a separate `./scripts/*` file. Like `./scripts/build.js` which does `less` and `browserify`.
So I think I am going to skip this PR.
If you want you can start with that instead (I am going to bed shortly) otherwise I will do it tomorrow.

Hirse:
I understand the plan is to move away from Grunt.
I wasn't sure how quickly that would happen, so I opened this as a quick follow up from the change in the ESLint PR.

campersau:
superseded by https://github.com/FredrikNoren/ungit/pull/1374

PR #1368, opened 2020-05-17
prevent tests hangs in credentials-helper

campersau:
> The credentials-helper tests sometimes don't pass when I am having a crashed ungit instance opened in chrome because I stopped the server and the crashed ungit pages tries to reconnect by sending polling requests with socket.io which then the node server in the credentials-helper test picks up and throws inside of the request handler which means that the server does not a response and eventually times out. So to prevent this edge case we could add a timeout to the node response so it times out faster.

See https://github.com/FredrikNoren/ungit/pull/1365#issuecomment-629758116y

### DIFF
--- a/test/spec.credentials-helper.js
+++ b/test/spec.credentials-helper.js
@@ -22,15 +22,19 @@ describe('credentials-helper', () => {
       res.end(JSON.stringify(payload));
     });
 
+    // to avoid tests hanging after they finished
+    // set a low timeout so when an exception occurs inside of the request handler
+    // it times out faster because no response will be generated
+    server.timeout = 200;
     server.listen(config.port, '127.0.0.1');
 
     const command = `node bin/credentials-helper ${socketId} ${config.port} ${remote} get`;
     child_process.exec(command, (err, stdout, stderr) => {
+      server.close();
       expect(err).to.not.be.ok();
       const ss = stdout.split('\n');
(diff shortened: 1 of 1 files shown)


test/spec.credentials-helper.js
```diff
@@ -11,26 +11,33 @@ describe('credentials-helper', () => {
...
+        const params = querystring.parse(reqUrl.query);
+        expect(params['remote']).to.be(`${remote}`);
+        expect(params['socketId']).to.be(`${socketId}`);
 
-      res.writeHead(200, { 'Content-Type': 'application/json' });
-      res.end(JSON.stringify(payload));
+        res.writeHead(200, { 'Content-Type': 'application/json' });
+        res.end(JSON.stringify(payload));
+      } finally {
+        if (!res.finished) {
```
Hirse, line 25: It seems that `finished` is deprecated: https://nodejs.org/api/http.html#http_response_finished
Maybe this should use `writableEnded` or `writableFinished` instead?

campersau: These two APIs were added in `v12.9.0` and `v12.7.0` so we have to use `finished` for now...

Hirse: Oh, I didn't see that part, I was too distracted by the big, red "Deprecated" banner. 😄

PR #1371, opened 2020-05-17
move bumpdependencies to npm script

campersau:
using https://www.npmjs.com/package/npm-check-updates
part of https://github.com/FredrikNoren/ungit/issues/895

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -21,7 +21,7 @@ jobs:
           node-version: '*'
 
       - run: npm ci
-      - run: ./node_modules/.bin/grunt bumpdependencies
+      - run: npm run bumpdependencies
       - run: npm install
 
       - name: Create Pull Request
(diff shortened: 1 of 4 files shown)


Hirse:
Would it be possible to use the output of `npm-check-updates` for the PR body?

Since we are using version ranges, should we use `ncu -m, --minimal`?
> do not upgrade to newer versions that are already satisfied by the existing version range

Alternatively, with this CI setup we could consider removing the ranges altogether and just use pinned versions.

campersau:
> Would it be possible to use the output of npm-check-updates for the PR body?

That was a bit challenging but it now looks like this: https://github.com/campersau/ungit/pull/6

> Since we are using version ranges, should we use ncu -m, --minimal?

We were always bumping to the latest none prerelease version to also update the `package-lock.json`

> Alternatively, with this CI setup we could consider removing the ranges altogether and just use pinned versions.

Yes, we could but there are third party integrations which might be affected by that?

Hirse:
Output looks great. Thanks.

PR #1367, opened 2020-05-17
Only package relevant files for electron

Hirse:
Inspired by #1366, only add required files to the electron package.

PR #1373, opened 2020-05-17
move package to npm scripts

campersau:
also make `--all` optional
part of #895 
supersedes https://github.com/FredrikNoren/ungit/pull/1367


package.json
```diff
@@ -13,9 +13,9 @@
     "format": "npm run lint -- --fix",
     "build": "grunt",
     "watch": "grunt watch",
-    "package": "grunt package",
+    "package": "npm run build && node ./scripts/package.js",
     "travisnpmpublish": "grunt travisnpmpublish",
-    "electronpublish": "grunt electronpublish"
+    "electronpublish": "node ./scripts/electronpublish.js"
```
Hirse, line 18: I also don't particularly like the `electronpublish`-name since it doesn't actually publish anything as far as I can tell.

Maybe we could use `electronzip` or `electronarchive` instead.

.github/workflows/ci.yml
```diff
@@ -55,7 +55,7 @@ jobs:
 
       - run: npm ci
       - run: npm run lint
-      - run: npm run package
+      - run: npm run package -- --all
```
Hirse, line 58: Instead of testing all electron options on each run, would it make sense to restrict this to the host platform?
(I.e. only build win32 electron on Windows...)

package.json
```diff
@@ -13,9 +13,9 @@
     "format": "npm run lint -- --fix",
     "build": "grunt",
     "watch": "grunt watch",
-    "package": "grunt package",
+    "package": "npm run build && node ./scripts/package.js",
```
Hirse, line 16: Can we change this script's name to contain `electron`. 
As it stands, it can be confusing with `npm pack`, which creates the archive that is used by npm.

scripts/electronpublish.js
```diff
@@ -0,0 +1,37 @@
+const fs = require('fs');
+const path = require('path');
+const mkdirp = require('mkdirp');
+const util = require('util');
+const rimraf = util.promisify(require('rimraf'));
```
Hirse, line 5: Please add `mkdirp` and `rimraf` to devDependencies if you are using them here.

scripts/electronpublish.js
```diff
@@ -0,0 +1,37 @@
...
+const archiver = require('archiver');
+const baseDir = path.join(__dirname, '..');
+const buildDir = path.join(baseDir, 'build');
+const distDir = path.join(baseDir, 'dist');
+
+(async () => {
+  await rimraf(distDir);
+  await mkdirp(distDir);
```
Hirse, line 13: Is this actually necessary?
As far as I can tell, `fs.createWriteStream` would just overwrite existing files.

For cleaning out `dist` the following might be sufficient.
```js
if (fs.existsSync(distDir)) {
  (await fs.promises.readdir(buildDir)).forEach((oldZip) => {
    await fs.promises.unlink(oldZip);
  });
} else {
  await fs.promises.mkdir(distDir);
}

File: scripts/electronpublish.js [hunk 6]
```diff
@@ -0,0 +1,37 @@
...
+  const folders = await fs.promises.readdir(buildDir);
+  const promises = folders.map((folder) => {
+    const source = path.join(buildDir, folder);
+    const destination = path.join(distDir, folder + '.zip');
```
Hirse, line 17:
```suggestion
    const destination = path.join(distDir, `${folder}.zip`);
```

scripts/electronpublish.js
```diff
@@ -0,0 +1,37 @@
...
+  await new Promise((resolve, reject) => {
+    archive
+      .directory(source, path.basename(source))
+      .on('error', (err) => reject(err))
+      .pipe(stream);
+
+    stream.on('close', () => resolve());
+    archive.finalize();
+  });
+  console.log(`zip ${path.relative(baseDir, destination)}`);
```
Hirse, line 36: It would be nice to also have some logging when starting a zip, so if something breaks it's easy to see which one broke.

Final output like this might be nice:
```
Start Zipping electron packages
start zip win32-32
end zip win32-32
...
End Zipping electron packages
```

scripts/package.js
```diff
@@ -0,0 +1,18 @@
+const process = require('process');
+const path = require('path');
+const baseDir = path.join(__dirname, '..');
```
Hirse, line 3: Separate imports from rest.

```suggestion

const baseDir = path.join(__dirname, '..');
```

scripts/package.js
```diff
@@ -0,0 +1,18 @@
...
+
+require('electron-packager')({
```
Hirse, line 5: Personally, I wold prefer to have `electron-packager` explicitly imported at the top of the file.

.github/workflows/ci.yml
campersau: We could even just use `npm run build` and only for the node `10` build which currently publishes the electron artifacts run `npm run package` to speed things up.
When I first integrated it I wanted to make sure that this runs on all OSs for all packages but maybe we can remove that now.
This also sometimes hangs the macOS build.

scripts/electronpublish.js
campersau: I personally find that syntax harder to read when just concatenating two string.
But we already have that in the codebase so I will probably update it.

.github/workflows/ci.yml
Hirse: I would be in favor of speeding things up.

campersau:
`npm run electronpackage` is now only run when necessary in CI, also it now requires `npm run build` to be run manually before.
Successful macOS builds are now taking around 4min instead of > 10min.

scripts/electronzip.js
```diff
@@ -0,0 +1,46 @@
...
+  try {
+    distFolders = await fs.readdir(distDir);
+  } catch (e) {
+    await fs.mkdir(distDir);
+  }
+  for (const oldZip of distFolders) {
+    await fs.unlink(path.join(distDir, oldZip));
+  }
+
+  const folders = await fs.readdir(buildDir);
```
Hirse, line 21: This script would throw if being run before `electronpackage` as `build` wouldn't exist yet, right?
```suggestion
  try {
    const folders = await fs.readdir(buildDir);
  } catch (e) {
    console.error("Run "npm run electronpackage" before zipping');
    return;
  }
```

campersau: Right, this script was only intended to be run on CI so I left this check out.

scripts/electronpackage.js
```diff
@@ -0,0 +1,20 @@
...
+electronPackager({
+  dir: baseDir,
+  out: path.join(baseDir, 'build'),
+  icon: path.join(baseDir, 'public/images/icon'),
+  all: process.argv.includes('--all'),
+  asar: true,
+  overwrite: true,
+  ignore: [
+    /^\/(?:[^/]+?\/)*(?:\..+|.+\.less)$/, // dot-files and less files anywhere
+    /^\/(?:\..+|assets|clicktests|coverage|dist|test)\//, // folders in root
```
Hirse, line 16: Ignore `scripts`, see #1376.
```suggestion
    /^\/(?:\..+|assets|clicktests|coverage|dist|scripts|test)\//, // folders in root
```

PR #1372, opened 2020-05-17
move travisnpmpublish to npm scripts

campersau:
basically copy & paste the existing code in separate file.
part of https://github.com/FredrikNoren/ungit/issues/895

PR #1366, opened 2020-05-17
Clean up ignore files

Hirse:
This PR cleans up 4 of the existing ignore files (`.gitignore`, `.nodemonignore`, `.npmignore`, and `.prettierignore`). The most important changes are in which files are included in the npm package.

### Updated npm ignore
The configuration in this PR removes as many files from the npm package as possible (including for example screenshots and documentation). This is more intended as a starting point to pick which other files to include.

Removed files are
- Component Less and non-bundles JS files
- Common and vendor Less files
- Markdown docs (Plugins and Mergetool)
- Images from Readme

See below for the complete list of which files were and are now included.
In total 99 files, taking about 700kb, are now excluded.

### Other Changes
- `.gitignore`
  - Only include existing folders and files
  - Reorder and add comments
  - Add folders created by nyc in preparation for #1365
  - Add entries that might commonly be created by developers
- `.nodemonignore`
  - Delete since I couldn't find usage
- `.npmignore`
  - Reorder and add comments
  - Ignore all build ressources
  - Add folders created by nyc in preparation for #1365
  - Add entries that might commonly be created by developers
- `.prettierignore`
  - Reorder
  - Add folders created by nyc in preparation for #1365

### List of files in npm package
```diff
-.gitattributes
-.nodemonignore
 bin/credentials-helper
 bin/ungit
 CHANGELOG.md
 components/app/app.bundle.js
 components/app/app.css
 components/app/app.html
-components/app/app.js
-components/app/app.less
 components/app/ungit-plugin.json
 components/branches/branches.bundle.js
 components/branches/branches.css
 components/branches/branches.html
-components/branches/branches.js
-components/branches/branches.less
 components/branches/ungit-plugin.json
 components/commit/commit.bundle.js
 components/commit/commit.css
 components/commit/commit.html
-components/commit/commit.js
-components/commit/commit.less
 components/commit/ungit-plugin.json
 components/commitdiff/commitdiff.bundle.js
 components/commitdiff/commitdiff.css
 components/commitdiff/commitdiff.html
-components/commitdiff/commitdiff.js
-components/commitdiff/commitdiff.less
-components/commitdiff/commitlinediff.js
 components/commitdiff/ungit-plugin.json
 components/crash/crash.bundle.js
 components/crash/crash.css
 components/crash/crash.html
-components/crash/crash.js
-components/crash/crash.less
 components/crash/ungit-plugin.json
 components/dialogs/dialogs.bundle.js
-components/dialogs/dialogs.js
[...]

PR #1376, opened 2020-05-18
Add scripts to npmignore

Hirse:
Since #1366 and #1372 were developed in parallel, the new `script` folder was missed during the ignore cleanup.

This PR adds that folder and moves the existing `teststabilitytester.js` script there.


campersau:
Should we also copy the `source/.eslintrc.json` in the `scripts` folder?

campersau:
eslint does not like `#!/usr/bin/env node` in `scripts\teststabilitytester.js`

PR #1378, opened 2020-05-27
Encode URI paths with slashes

wmertens:
It's fine for query parameters to have visible `/` in them, and it makes it easier to work with the ungit URLs


wmertens:
@Hirse changes applied :)

campersau (approved):
Thanks for the update!

Issue #1014, opened 2018-01-19
Errors when working offline


jpschewe:
Today I was doing some work without being connected to the internet and ungit kept giving me errors. It seems that it doesn't like not being able to talk to the upstream repository.

```
2017-12-10T18:16:54.754Z - warn: Responding with ERROR:  {"isGitError":true,"errorCode":"offline","command":"-c color.ui=false -c core.quotepath=false -c core.pager=cat -c credential.helper=/home/jpschewe/src/mungit/bin/credentials-helper 2 8448 fetch origin --prune","workingDirectory":"/home/jpschewe/projects/fll-sw/working-dir","error":"ssh: Could not resolve hostname github.com: Name or service not known\r\nfatal: Could not read from remote repository.\n\nPlease make sure you have the correct access rights\nand the repository exists.\n","message":"ssh: Could not resolve hostname github.com: Name or service not known\r","stderr":"ssh: Could not resolve hostname github.com: Name or service not known\r\nfatal: Could not read from remote repository.\n\nPlease make sure you have the correct access rights\nand the repository exists.\n","stdout":""}
2017-12-10T18:16:54.754Z - warn: Sending exception to Sentry.io
raven@2.2.1 alert: failed to send exception to sentry: getaddrinfo ENOTFOUND sentry.io sentry.io:443
raven@2.2.1 alert: failed to send exception to sentry: getaddrinfo ENOTFOUND sentry.io sentry.io:443
```

Originally posted to mungit at https://github.com/exsilium/mungit/issues/9

jung-kim:
To me, ungit's goal is to "git made easy".  To me, this means it does a lot of things in the background automatically for you or bundles some of frequent the operations.  And this result us in some what more frequent querying of the remote repos, and communication.

long time ago I've tried to do some what more offline tolerating version but had some odd same origin policy.  I will try to resume this effort again

jpschewe:
I think it's great that ungit does these things in the background. However I works expect it to fail gracefully and not being up an error reporting screen when this particular issue comes up.

jpschewe:
It seems that this problem is back. I'm getting a block at the top of my screen "Unhandled git error!" every couple of minutes and it keeps stacking more and more up until I close the browser window or connect to the network with the git server. I then need to close each one individually or reload the page for them all to go away.

jung-kim:
Hmm, code for this shouldn't have changed. Been crazy busy and will be crazy busy for awhile but will check it when I get a some time.

[...]

PR #1381, opened 2020-06-05
When autoFetch=false don't make remote repo calls automatically

### LINKED ISSUE #1380: autoFetch: False still tries to fetch tags and erroring out if not successful
If you are disconnected from a remote repo or say don't have your private key in to access a remote repo there is no option to avoid error.    `authFetch: false` is not enough to disable ungit from trying to connect to the remote to fetch the tags.  
The tags will be fetched on focus/mouse move ( causing `working-tree-changed` event to fire).    There is also `git-directory-changed` both which will cause branches.js and remotes.js to fire remote calls (and remotes.js will throw errors on screen).

This also probably leads to more battery drain in an offline/disconnect situation.

Will submit a PR I believe doesn't break anything but does allow offline/non remote support.

mitchcapper:
closes #1380

Note the discardchanges test does fail with this due to:
` Error: Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (ungit\test\spec.git-api.discardchanges.js)`
however I didn't see the specific reason is failing and it fails in master on my machine so ignored that.

An alternative would be to add an offline option (with a config setting to default start in this) that takes a step further to not hit remote repos and allow local work.

Review and let me know whats best.

### DIFF
--- a/components/branches/branches.js
+++ b/components/branches/branches.js
@@ -27,6 +27,7 @@ class BranchesViewModel {
       this.updateRefs();
       return value;
     };
+    this.shouldAutoFetch = ungit.config.autoFetch;
     this.isShowRemote.subscribe(setLocalStorageAndUpdate.bind(null, showRemote));
     this.isShowBranch.subscribe(setLocalStorageAndUpdate.bind(null, showBranch));
     this.isShowTag.subscribe(setLocalStorageAndUpdate.bind(null, showTag));
@@ -44,7 +45,7 @@ class BranchesViewModel {
     ko.renderTemplate('branches', this, {}, parentElement);
   }
   clickFetch() {
-    this.updateRefs();
+    this.updateRefs(true);
   }
   onProgramEvent(event) {
     if (
@@ -56,7 +57,9 @@ class BranchesViewModel {
       this.updateRefsDebounced();
     }
   }
-  updateRefs() {
+  updateRefs(forceRemoteFetch) {
+    forceRemoteFetch = forceRemoteFetch || this.shouldAutoFetch || '';
+
     const currentBranchProm = this.server
       .getPromise('/branches', { path: this.repoPath() })
       .then((branches) =>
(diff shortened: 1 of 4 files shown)


campersau:
I like the idea because our current [`detectReActivity`](https://github.com/FredrikNoren/ungit/blob/7b034138b8ce963dec031dc1ddbf90ab07bc95dc/public/source/main.js#L115-L127) logic makes `autoFetch: false` pretty much useless since it will fetch anyway after moving the mouse. 

But instead of disabling the loading of the branches completely I would prefer to just disable the `fetch` before `show-ref` on the server so you can still see the locally cached data.
https://github.com/FredrikNoren/ungit/blob/7b034138b8ce963dec031dc1ddbf90ab07bc95dc/source/git-api.js#L447-L458
And add a new parameter to force a `fetch` if the users clicks on the refresh button manually.

Also the `autoFetch` config description currently states that it just "prevents" the initial fetch and should be updated accordingly.
https://github.com/FredrikNoren/ungit/blob/7b034138b8ce963dec031dc1ddbf90ab07bc95dc/source/config.js#L237


Something like this was also proposed here https://github.com/FredrikNoren/ungit/issues/1014#issuecomment-477943414

mitchcapper:
Sounds good shall I make those changes and update the docs?

campersau:
@mitchcapper Yes, that would be great!

mitchcapper:
Ok updated, let me know if this is what you were thinking.  It seems to perform as desired.

mitchcapper:
Done.

components/branches/branches.js
```diff
@@ -24,9 +24,10 @@ class BranchesViewModel {
     this.graph = graph;
     const setLocalStorageAndUpdate = (localStorageKey, value) => {
       storage.setItem(localStorageKey, value);
-      this.updateRefs();
+      this.updateRefs(this.shouldAutoFetch);
```
campersau, line 27: Revert
```suggestion
      this.updateRefs();
```

PR #1383, opened 2020-06-13
Update warning message to be gramatically correct

nathanwoodcodes:
I'm unsure what the specific intent of this message is. Is it specifically saying "you may be reversing history, which could loose commits", or "you may be rewriting history, which is bad practice", both, or something else? I ask, because it might be helpful to add a link to some git tutorial page which explains why it's bad or what could go wrong, or to make the message more specific.


campersau:
It is more like `you may be reversing history, which could lose commits` but I personally think the message is fine since it does not reverse anything, you can still go back to the future if there was another ref. It is just moving a ref to another commit.

Issue #315, opened 2014-02-21
Resizing commit body box doesn't trigger repositioning of files changed


jwir3:
![screen shot 2014-02-21 at 5 33 37 pm](https://f.cloud.github.com/assets/997106/2235892/cb9e2b80-9b50-11e3-8b74-2e86702a0dc4.png)

See above screenshot. Once the commit-body box is resized via the lower right corner handle, it seems to be displayed behind the 'files to commit' checklist. This results in two problems - 1) The text of the checklist and the text of the commit body message is interleaved, making it difficult to read, and 2) I can't resize the commit body message box any longer, since it seems to be laid out behind the checklist of items.

I think a responsive design, where the checklist is moved further to the right as the commit body message box is resized wider, would be most effective here.

PR #1390, opened 2020-06-28
Prevent commit message <textarea> from resizing horizontally

simonwiles:
> Resizing horizontally (typically by accident when trying to resize vertically only, in my case) screws up the staging panel layout.  This change prevents that.

This is an adequate solution (to my mind at least) to #315, which has been bothering me too.

~If the change is acceptable, would you like me to bump the patch version and update CHANGELOG.md?  The instructions at [CONTRIBUTING.md](https://github.com/FredrikNoren/ungit/blob/master/CONTRIBUTING.md#pull-requests) seem to be out-of-date w/r/t the current CHANGELOG.md style.  Perhaps just include a line under "Fixed" in the "Unreleased" section?~

Note: the Windows test which is failing below with a 404 fetching resources is currently passing on my fork.

### DIFF
--- a/components/staging/staging.less
+++ b/components/staging/staging.less
@@ -13,7 +13,7 @@
     }
   }
 
-  textarea.form-control {
+  textarea.commit-body {
     resize: vertical;
   }
 
--- a/components/staging/staging.less
+++ b/components/staging/staging.less
@@ -13,6 +13,10 @@
     }
   }
 
+  textarea.form-control {
+    resize: vertical;
+  }
+
   .arrow {
     border-top-color: #643a44;
     left: (@log-width-small + 45px);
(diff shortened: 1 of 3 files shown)


public/less/bootstrap.less
```diff
@@ -50,3 +50,7 @@
 .panel {
   border: none;
 }
+
+.staging textarea.form-control& {
```
campersau, line 54: Can you please move that into [`staging.less`](https://github.com/FredrikNoren/ungit/blob/1b6ccefab995893f3cc3a353e976217a258c867a/components/staging/staging.less)?

simonwiles: No worries, sorry -- don't know how I missed that.

components/staging/staging.less
```diff
@@ -13,6 +13,10 @@
     }
   }
 
+  textarea.form-control {
```
Hirse, line 16: Would this change all textareas?

Would it be better to add a `class` like `commit-body` and use that?

simonwiles: Well, it would change all textareas with the class `form-control` in the `.staging` area...  I don't think there are others?

PR #1393, opened 2020-07-04
Fix git ignore settings

Hirse:
Version 1.5.8 has the `public/js/ungit.js` and `public/css/style.css` files missing on npm.
It seems that all `.gitignore` files get transformed to `.npmignore` (unless there is an actual `.npmignore` file), even when they are in subfolders. 

This PR removes the `.gitignore` files that are not in the root folder and creates the required folders in the build step.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@ build/
 coverage/
 dist/
 node_modules/
+public/css/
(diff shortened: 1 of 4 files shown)


campersau:
Ups.. thanks!

PR #1395, opened 2020-07-04
Import Bootstrap from npm and upgrade to latest 3.x

Hirse:
Most changes in bootstrap's minor version upgrades are fixes and security patches.
I only found the following changes required for us:
* Replace dropdown `.pull-right` with `.dropdown-menu-right`
* Remove `white-space` from `.dropdown-header`
* Use new `.container-fluid` class for wide container

A further upgrade to 4.x is more complicated as it contains lots of breaking changes and switches from LESS to SCSS.


campersau:
The tests are failing because that apparently wasn't the best idea:
https://github.com/FredrikNoren/ungit/blob/a65cbc80595e79d3ac837dcfd23a2e7f8eb82dbc/components/graph/git-node.js#L178

You could replace that with for example this
```js
const textBox = event.currentTarget.parentElement.querySelector('input[type="search"]');
```
to fix it.

PR #1399, opened 2020-07-09
Update glyphicons after bootstrap update

campersau:
also including `woff2` which was missing and resulted in [404 errors](https://github.com/FredrikNoren/ungit/runs/847067026?check_suite_focus=true#step:17:324) after bootstrap update in #1395.

We could also add them to the copy step in the build script.
https://github.com/FredrikNoren/ungit/blob/9fc2903d252d1caf0f6c1f3e9b025a2af4ae3346/scripts/build.js#L94

cc @Hirse


Hirse:
Thanks. 
I might look into copying them during build.

PR #1401, opened 2020-07-12
Copy glyphicons from node_modules

Hirse:
Follow up from #1395 and #1399.

@campersau I'm also looking into copying *Open Sans* and *Source Code Pro* from `node_modules`.


scripts/build.js
```diff
@@ -89,14 +89,24 @@ const baseDir = path.join(__dirname, '..');
   );
 
   // copy
-  console.log('copy');
+  console.log('copy bootstrap fonts');
+  await Promise.all(
+    [
+      'node_modules/bootstrap/fonts/glyphicons-halflings-regular.eot',
+      'node_modules/bootstrap/fonts/glyphicons-halflings-regular.svg',
+      'node_modules/bootstrap/fonts/glyphicons-halflings-regular.ttf',
```
Hirse, line 97: I'm not sure if these are necessary since `woff` seems to be [supported by all browsers](https://caniuse.com/#feat=woff) (and we only include woff for the other fonts).

PR #1420, opened 2020-09-04
add `ssh-add -K`

jung-kim:


README.md
```diff
@@ -59,8 +59,17 @@ Example of `~/.ungitrc` configuration file to change default port and enable bug
 }
 ```
 
+
+Github SSH Authorization
```
campersau, line 63: Does this also work for none github repositories? If so I would remove it from here.

```suggestion
SSH Authorization
```

README.md
```diff
@@ -59,8 +59,17 @@ Example of `~/.ungitrc` configuration file to change default port and enable bug
 }
 ```
 
+
+SSH Authorization
+-----------------
+Ungit does run git commands in background and authentication will be asked continously when working with private repos.  Easiest way to prevent this is to set up ssh authorization and cache it.
+
+1. Follow github's [instruction](https://docs.github.com/en/github/authenticating-to-github/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent)
```
Hirse, line 67: 1. Correctly capitalize GitHub
2. Remove language from URL to default to user language

```suggestion
1. Follow GitHub's [instructions](https://docs.github.com/github/authenticating-to-github/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent)
```

README.md
```diff
@@ -59,8 +59,17 @@ Example of `~/.ungitrc` configuration file to change default port and enable bug
...
+2. Ensure `ssh-add -K ~/.ssh/id_rsa` or with whatever ssh key file name is executed.
```
Hirse, line 68: Is this not covered by the instructions linked above?

Additionally, it seems that the `-K` option only applies to (most) macs:
image

README.md
```diff
@@ -59,8 +59,17 @@ Example of `~/.ungitrc` configuration file to change default port and enable bug
 }
 ```
 
+
+SSH Authorization
```
Hirse, line 63: The instructions from GitHub that are linked below speak of "Authentication" instead of "Authorization" which I think is the correct term in this case.

```suggestion
SSH Authentication
```

README.md
jung-kim: It is covered, but not always followed, which is why I would like to ensure.

As far as `-K` being only for mac only I didn't know about. Maybe it be better to change to more targeted ssh-add

README.md
```diff
@@ -59,8 +59,17 @@ Example of `~/.ungitrc` configuration file to change default port and enable bug
 }
 ```
 
+
+SSH Authentication
+-----------------
+Ungit does run git commands in background and authentication will be asked continously when working with private repos.  Easiest way to prevent this is to set up ssh authorization and cache it.
```
Hirse, line 65: 1. Fix typo.
2. "authorization" -> "authentication" (see above
```suggestion
Ungit does run git commands in background and authentication will be asked continuously when working with private repos.  Easiest way to prevent this is to set up ssh authentication and cache it.
```

Issue #1421, opened 2020-09-12
Click & Drag Selection on Patch Changes


Sogolumbo:
Current behaviour:
When selecting `Patch` all lines are checked. The only way to change the selection is to click on all checkboxes of the lines that should be deselected.
To me this feels very inefficient and a little bit annoying, especially with long commits (>100 lines) which need patching the most.

Feature proposal:
Make it possible to select or deselect multiple lines at once by clicking and holding on the first checkbox and releasing the mouse on the last line.
To make this feature easier to understand and find it would be nice if the state of the checkboxes would change while moving the mouse around.

More ideas:
It would be nice if the clickable area of the checkboxes would be a little bit bigger (maybe a square which fills the whole height of the line).
A button which selects/deselects all checkboxes would also be very useful.

wmertens:
Ideally, you'd simply select a bunch of lines and then there's a popup to select/deselect. This works easily with touch, too.

Also, a toggle all lines button would be great.

Sogolumbo:
> Ideally, you'd simply select a bunch of lines and then there's a popup to select/deselect.
What you're wishing for could be done with a 'invert selection' button, right?

What I'm talking about is something different: being able to select/deselect many specific lines at once. This is useful when you have hundreds of changed lines and roughly half of them need to go into one commit, the rest somewhere else.

Sogolumbo:
About the touch surfaces: The feature might work best with double tap & drag on touch surfaces.

wmertens:
How about a select strip: when you use the mouse you need to click and drag to toggle lines. For touch, you simly need to touch and drag. The first state you toggle is the one that is repeated across the drag. That seems very fast to me.

It needs to be a strip because otherwise you can't scroll any more on touch surfaces.

simonwiles:
I've actually written a simple ~15 line change which allows clicking (or tapping) any part of the line to toggle the checkbox.  I've was gonna tidy it up and offer it as a PR but life is coming at us all rather fast at the minute, huh...  It doesn't have any dragging support, but it's a really simple change that improves usability considerably in the kind of scenario @Sogolumbo is describing (at least, for me imo).  It doesn't implement the fancy UI described here (which sounds cool btw), but I reckon it's one of those 80/20 things -- it eases a lot of the pain with a minimally invasive intervention.  If no ones actually currently working on a more involved solution to this issue, it might be a worthwhile halfway house in the interim?

wmertens:
[...]

PR #1429, opened 2020-09-23
 Allow toggling patch checkbox by clicking on lines

simonwiles:
This PR changes the line-select UI in patch mode to allow toggling lines by clicking anywhere on the line.  It doesn't address all the desired improvements described in #1421 (and so won't close it), but I've been using this modification for a while and it does make things somewhat easier.

If it passes an initial review I can update the CHANGELOG.md and bump the version number.

Peek 2020-09-23 08-31


wmertens:
Would it be possible as a quick win to add a button for toggling all? Just an idea, not worth blocking this over.

wmertens (approved):
Knockout novice but LGTM

simonwiles:
> Would it be possible as a quick win to add a button for toggling all? Just an idea, not worth blocking this over.

Yeah, I like the idea of an "invert selection" button, or something like that.  I'm not going to be able to do it right now, but I'll make a note.  I might look at addressing actual drag selecting if I can find time too, but I'd rather see this closed off first, because with the best will in the world I don't know if/when I'll have time!

> Knockout novice but LGTM

Likewise, but tbh this is more about getting the regex find/replace right :)

Is the advice at https://github.com/FredrikNoren/ungit/blob/master/CONTRIBUTING.md#pull-requests still valid?  I notice the release process seems to have changed a bit since this was written.  I've updated the changelog under "unreleased" and not touched the version number in `package.json` -- hope that's okay.

components/textdiff/textdiff.js
```diff
@@ -169,13 +169,23 @@ class TextDiffViewModel {
...
+              this.patchLineList()[index] = true;
+            }
+            content = content.replace(
+              /<span class="d2h-code-line-[a-z]+">(\+|-)<\/span>/g,
+              (match, capture) => {
+                return this.getPatchCheckBox(capture, index, this.patchLineList()[index++]);
+              }
+            );
+            return `<td class="${tdClass}" ${
+              this.editState() === 'patched' ? 'style="cursor:pointer"' : ''
```
Hirse, line 185: Could you add a class instead and move this style to CSS?

components/textdiff/textdiff.js
```diff
@@ -198,7 +208,7 @@ class TextDiffViewModel {
     }
     return `<div class="d2h-code-line-prefix"><span data-bind="visible: editState() !== 'patched'">${symbol}</span><input ${
       isActive ? 'checked' : ''
-    } type="checkbox" data-bind="visible: editState() === 'patched', click: togglePatchLine.bind($data, ${index})"></input></div>`;
+    } type="checkbox" data-bind="visible: editState() === 'patched', click: togglePatchLine.bind($data, ${index}), clickBubble: false"></input></div>`;
```
Hirse, line 211: Since the checkbox is inside the row, this click event on the checkbox is not really needed, right?

components/textdiff/textdiff.js
simonwiles: Done!

components/textdiff/textdiff.js
simonwiles: Well, it's needed in the sense that it's invoked here:
https://github.com/FredrikNoren/ungit/blob/406a4b0652a9592d2899895019625b7100c3fdf2/components/textdiff/textdiff.js#L228-L231
and simply removing it means the functionality no longer works :)

But yes, I agree it feels like it should be redundant (I originally made these changes as a quick-and-dirty hack for my own use some months ago, but the conversation at #1421 prompted my to PR them).

The alternative, it seems to me, is to bind the "checked" state of the checkbox to the index in `TextDiffViewModel.patchLineList` and binding `togglePatchLine(index)` to the click event on the `<td/>` -- so I've added a commit that does that.

There are two downsides to this: 1) is that, as the comment at https://github.com/FredrikNoren/ungit/blob/406a4b0652a9592d2899895019625b7100c3fdf2/components/textdiff/textdiff.js#L168-L170 mentions, it is necessary to force a re-render; and 2) is that the `data-bind="checked: ...` binding is two-way, so the normal behaviour of the checkbox has to be inhibited otherwise a) the checkbox is changed and the immediately re-changed, and b) `togglePatchLine(index)` is only called once, and the interface ends up in an inconsistent state.  I've addressed that by simply applying `point-events: none` to the checkbox itself (making the `checked` binding effectively one-way), but there may be a better way?

All in all, I'm not 100% convinced this is better than before -- what do you think?

components/textdiff/textdiff.less
```diff
@@ -3,4 +3,12 @@
     padding: 10px 0;
     text-align: center;
   }
+
+  td.patched {
+    cursor: pointer;
+  }
+
+  .d2h-code-line-prefix > input[type='checkbox'] {
```
campersau, line 11: Could this be moved to here so everything regarding the checkbox is in one place?
https://github.com/FredrikNoren/ungit/blob/3c80532c102be4b7227812faeb6bc3a330a2e8eb/components/staging/staging.less#L101-L105

components/textdiff/textdiff.js
```diff
@@ -198,7 +206,7 @@ class TextDiffViewModel {
     }
     return `<div class="d2h-code-line-prefix"><span data-bind="visible: editState() !== 'patched'">${symbol}</span><input ${
       isActive ? 'checked' : ''
```
campersau, line 208: This can be removed.

Also the [`isActive` check](https://github.com/FredrikNoren/ungit/pull/1429/files#diff-21bc09960eb0842e91cc5dfecb8efa75R204) could go [here](https://github.com/FredrikNoren/ungit/pull/1429/files#diff-21bc09960eb0842e91cc5dfecb8efa75R178) so `getPatchCheckbox` has no sideeffects.

components/textdiff/textdiff.less
```diff
@@ -3,4 +3,12 @@
     padding: 10px 0;
     text-align: center;
   }
+
+  td.patched {
```
campersau, line 7: Maybe move this as well to `staging.less` since it is only used there.

components/textdiff/textdiff.less
simonwiles: Sure, no problem (see 34f036a6dca58fc4f546f006a50751c13bad2ccc).

components/textdiff/textdiff.less
simonwiles: Sure, no problem (see 34f036a6dca58fc4f546f006a50751c13bad2ccc).

components/textdiff/textdiff.js
simonwiles: Okay, this is done.  See 74b679f and 051d956.

simonwiles:
The `numberOfSelectedPatchLines` variable was only used to reset the edit state when it evaluated to zero, so I dropped it and replace the check with a test again `this.patchLineList().filter(Boolean).length` (see 825e120).

simonwiles:
### Important:
When investigating this, I noticed a bug in the way that ungit operates in the present release version (`1.5.11`).  See below for a quick demonstration:

Peek 2020-10-01 22-56

The reason this occurs is that after all the checkboxes are cleared, the edit state is set to `none`, and they are hidden; when the patch button is clicked again, the array returned by `patchLineList()` is reinitialized to all `true`, but the UI is not re-rendered, so the boxes all appear empty.  Clicking one inverts it's state, and so what gets committed as the patch is the inverse of the UI state.  Note too that not moving the mouse for 3 seconds and then moving it again causes the UI to re-render (because of https://github.com/FredrikNoren/ungit/blob/9374f6c0f6dafc6dd8f9240eb506a087e1fa39ce/public/source/main.js#L115-L125):

Peek 2020-10-01 23-03

This became much more obvious with the earlier commits in this PR, because the diff (and checkboxes) are re-rendered on click:

Peek 2020-10-01 22-14


My choice to address this was to make the diff re-render on each render (i.e. not checking for a difference in the html string, as this check is inadequate to determine if the state of the checkboxes has changed anyway).  This doesn't generate more re-renders or cause any thrashing, and fixes the issue.

components/textdiff/textdiff.js
```diff
@@ -162,28 +161,30 @@ class TextDiffViewModel {
...
+            );
+            return `<td class="${tdClass}" data-bind="click: (editState() === 'patched') ? togglePatchLine.bind($data, ${index++}) : null, css: { patched: editState() === 'patched' }">${content}</td>`;
           }
-
-          return this.getPatchCheckBox(capture, index, this.patchLineList()[index++]);
-        });
+        );
       }
 
-      if (html !== this.htmlSrc) {
```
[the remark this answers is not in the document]
simonwiles: > Unfortunatley this makes selecting lines a bit hard. Because the selection gets lost after the `detectReActivity` mouse move delay...

Yeah, that's an additional problem I hadn't noticed, thanks (sigh).

> Maybe we will need the `checked` attribute on the checkbox again to force the HTML to be different in your described scenario?

This won't work, as demonstrated by the fact that the bug I described in my last post is present in the v1.5.11 release where this attribute is still written.  The `html == this.htmlSrc` check compares the html the last time the render function ran against the newly generated html -- in the scenario above, the `editState` is set to `'none'` when all the boxes are unchecked, and the `patchLineList` array is emptied (not set to all `false`, for example), and so the `TextDiffViewModel.render` method recreates it will all values set to true.  The result is the html string contains all `checked` attributes on all checkboxes, and is identical to the last time the component rendered.

I'll give it some more thought; the whole way the `patchLineList` is connected with the `editState` and instantiated in `textdiff.js` and invalidated in `staging.js` could stand a rethink, and I'm starting to get far more familiar with the architecture here and also knockout.js than I ever intended.

 Fwiw none of these problems, except the bug already in production, exist with my original PR as submitted, and I'll be continuing to use it that way locally, as I have been for months now.

simonwiles:
Okay, this last commit addresses the bug in production, I think (although it's a workaround, not a fix).

wmertens:
I wish that you could click and drag (or tap + hold and drag) to select/unselect lines, and that there was a toggle select button like the toggle files button, but this PR is already a great improvement and I'd rather not wait for those improvements.

@campersau I tried it out, it merged cleanly and LGTM, would you consider merging?

simonwiles:
I'm just getting around to rebasing all my local changes on top of the most recent release, and I realized this PR is still open.  Might there be any interest in revisiting this?

PR #1430, opened 2020-10-03
Prevent selection of leading space in diffs when copying and pasting

simonwiles:
I ([and, it seems, at least some others](https://github.com/FredrikNoren/ungit/issues/1421#issuecomment-696916968)!) often find myself wanting to copy text from diffs in the commit history or in the staging area.  In fact, I had a PR merged last year that prevented selection of line number to make this process easier (#1215, thanks!).

This PR just adds a little to that idea, by preventing the selection of the leading space that still gets copied when selecting multiple lines (this is not generally a problem when using JS and Prettier on paste, for example, but for copying Python code, for example, it's a pain).

Hopefully this proves to be more straightforward than my other recent PR!  Thanks!

### DIFF
--- a/public/less/d2h.less
+++ b/public/less/d2h.less
@@ -55,6 +55,14 @@
   color: #000000;
 }
 
+.d2h-code-line {
+  user-select: none;
+}
+
+.d2h-code-line-ctn {
+  user-select: text;
+}
+
 .d2h-code-line del,
 .d2h-code-side-line del {
   background-color: rgba(255, 255, 255, 0.2);
--- a/public/less/d2h.less
+++ b/public/less/d2h.less
@@ -58,10 +58,13 @@
 .d2h-code-line,
 .d2h-code-side-line {
   user-select: none;
+  width: 100%;
 }
 
 .d2h-code-line-ctn {
+  display: inline-block;
   user-select: text;
+  width: 100%;
 }
 
 .d2h-code-line del,
--- a/public/less/d2h.less
+++ b/public/less/d2h.less
@@ -55,7 +55,8 @@
   color: #000000;
 }
 
-.d2h-code-line {
+.d2h-code-line,
+.d2h-code-side-line {
   user-select: none;
 }
 


public/less/d2h.less
```diff
@@ -55,6 +55,18 @@
   color: #000000;
 }
 
+.d2h-code-line,
+.d2h-code-side-line {
+  user-select: none;
+  width: 100%;
```
campersau, line 61: Looks like this will make selecting text easier so you can start dragging the mouse anywhere in a line and it does not need to be within text right?

campersau (commented):
Thanks! Looks good, but I think it would be better to fix it in [`diff2html`](https://github.com/rtfpessoa/diff2html) directly. We currently take the `diff2html` CSS and mostly override some colors in `d2h.less` so it fits nicer within ungits UI.

Earlier this year @rtfpessoa and I were working on some fixes regarding text selection see
https://github.com/rtfpessoa/diff2html/pull/304
https://github.com/rtfpessoa/diff2html/pull/306
https://github.com/rtfpessoa/diff2html/pull/307

public/less/d2h.less
simonwiles: Yeah -- it also requires this bit: https://github.com/simonwiles/ungit/blob/ef7543b466c902ddadf4f241da45bb32de28352e/public/less/d2h.less#L64-L68
```css
.d2h-code-line-ctn {
  display: inline-block;
  user-select: text;
  width: 100%;
}
```

so that the text-line container's parent has width too.

simonwiles:
> Earlier this year @rtfpessoa and I were working on some fixes regarding text selection see
> [rtfpessoa/diff2html#304](https://github.com/rtfpessoa/diff2html/pull/304)
> [rtfpessoa/diff2html#306](https://github.com/rtfpessoa/diff2html/pull/306)
> [rtfpessoa/diff2html#307](https://github.com/rtfpessoa/diff2html/pull/307)

Cool -- yes, the first of these is a migration of my earlier PR that was merged into this repo.  I did notice that the lines I'd added were no longer necessary in ungit because they'd made their way upstream, but I hadn't realized you'd been involved in that :)

Do you think it would be more appropriate to close this, then, and raise an issue and/or PR against `diff2html`?

campersau:
@simonwiles Yes, please raise an issue / PR in the [`diff2html`](https://github.com/rtfpessoa/diff2html) repo and after these changes are release we can just update our `diff2html` dependency. Thanks!

campersau:
Superseded by https://github.com/FredrikNoren/ungit/pull/1438

PR #1457, opened 2021-01-24
Update dependencies once a month

ylecuyer:
I'm tired of receiving an update dependencies almost all days. Once a month should be good enough.

### DIFF
--- a/.github/workflows/bump.yml
+++ b/.github/workflows/bump.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - master
   schedule:
-    - cron: '0 0 * * *'
+    - cron: '0 0 1 * *'
   workflow_dispatch:
 
 jobs:


ylecuyer:
@campersau could you merge this please :3

campersau:
I personally find it quite useful as it makes spotting breaking changes in dependencies, including git, early. (Like https://github.com/FredrikNoren/ungit/commit/c106484b8cda94e9a51e51cd72b2ffae45c20543, https://github.com/FredrikNoren/ungit/commit/cdc0e9d65d5ea6f42af2b495ea02988acef71796, https://github.com/FredrikNoren/ungit/commit/1d8b9037a6967a59f807166cdaabb2702e82c378, https://github.com/FredrikNoren/ungit/commit/c642d039b4edc942f75ad4d2c813fd5df74cd784)
Maybe it could be changed to only create / update a PR if tests fail, but since we support different OSs with different git versions it isn't easy to do.

ylecuyer:
Or maybe we can change the level to security bump only ? I hope we won't have a notification each day with such setup

Issue #108, opened 2013-08-21
Add a way to copy the SHA-1 of a commit to the clipboard


Ajedi32:
Ungit is great, but there are still many common features of git that it doesn't support. When I need to use one of those features, it would be nice if I could copy SHA-1's from commits in ungit so I can use them with command-line git.

PR #1462, opened 2021-02-06
Add clipboard button on commit

wmertens:
Right now, it shows when you hover or click the commit. How can I make it also show when I just hover over the commit hash?

Should this have a test? What should it test?


campersau:
Please note that `navigator.clipboard.writeText` will only work on `localhost` and `https` sites. Since ungit can bind to other hostnames as well (or all `0.0.0.0`) this would not work in these cases.

> [Note that the API only works when served over secured domains (https) or localhost and when the page is the browser's currently active tab.](https://alligator.io/js/async-clipboard-api/)

https://web.dev/async-clipboard/#security-and-permissions

See also https://stackoverflow.com/a/30810322 for possible workarounds.
You can test this scenario by setting `ungitBindIp` to `0.0.0.0` and then load ungit with `http://[yourcomputername]:8448`

---

ungit uses octicon icons everywhere else so I think it would be better to use them here too. E.g. https://primer.style/octicons/clippy-16

---

What do you think about making the button smaller like `btn-xs`?

wmertens:
@campersau actually I'm a complete n00b at KO and I'm working on the nodegit branch, this was a quick hack to get a feel for it. If you were to implement your suggestions I'd be really grateful :sweat_smile: (if not, I totally understand, I'm just saying that I wouldn't feel offended in the least if you implemented this properly, on the contrary)

campersau:
@wmertens  How do you like that?
image

Ref: https://github.com/FredrikNoren/ungit/issues/108

wmertens:
@campersau :100: :heart_eyes:

wmertens:
I just wonder how to show the button when you hover over the commit message, not the commit node. That way you can quickly copy an oid from the page.

campersau:
Or just keep the button always visible regardless of mouseover?

wmertens:
Hmm right, I was worrying that it adds too much visual noise

campersau:
@wmertens this is how you would only show the button on `mouseenter` / `mouseleave` but since the button is higher than the text it moves the files a bit down when it gets rendered. So for simplicity I am going to always show the button.
https://github.com/FredrikNoren/ungit/commit/4530d032839c868ff009abb8b603f670eb0e6879

PR #1463, opened 2021-02-06
prettier: add homer0 jsdoc plugin

wmertens:
This formats JSDoc comments. This is great for readability, consistency, writability and spotting errors in the definition.


wmertens:
Closed in favor of #1466

PR #1466, opened 2021-02-21
Types: add support for VSCode IntelliSense

wmertens:
this adds some non-commital TypeScript support, by adding JSDoc comment parsing to Prettier (allows checking for mistakes) and enabling error highlighting in VS Code and any others that might run TypeScript checking.

Note that no compiling is needed, this just adds static configuration to help out VSCode.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,5 +1,7 @@
 {
   "endOfLine": "lf",
+  "jsdocUseInlineCommentForASingleTagBlock": true,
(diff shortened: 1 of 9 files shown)


campersau:
This would supersede https://github.com/FredrikNoren/ungit/pull/1463 right?

wmertens:
Ah yes indeed, forgot I already made that one :)

wmertens:
Any chance of getting this merged? It only improves dev experience withoout impacting runtime. It doesn't have to be perfect from the start either...

Issue #1460, opened 2021-01-29
simple git flow breaks ungit


Console32:
I was trying to create a graphical representation of our git workflow. I used the following commands to create the history

```
git init                                                              
git commit --allow-empty -m 'Initial commit'                  
git checkout -b dev                                           
git checkout -b features/A                                     
git commit --allow-empty -m 'implement feature A'         
git commit --allow-empty -m 'implement feature A'         
git checkout dev                                          
git checkout -b features/B                                     
git commit --allow-empty -m 'implement feature B'         
git commit --allow-empty -m 'implement feature B'         
git checkout dev                                          
git merge --no-ff features/A                               
git merge --no-ff features/B                                   
```

The resulting graph looks like this in my CLI:
![image](https://user-images.githubusercontent.com/4870099/106240844-46603200-6205-11eb-8cb7-095ef9b26561.png)

But ungit shows:
![image](https://user-images.githubusercontent.com/4870099/106240912-65f75a80-6205-11eb-992e-bf329e0eed5f.png)

I tried a fresh start - same result.

antibunching:
I am facing the same issue, with ver1.5.11 on windows 10, and ver1.5.13 on Debian 10.

wmertens:
This is what it looks like with #1315 : 
![image](https://user-images.githubusercontent.com/54934/109050132-59b5ce80-76d9-11eb-80eb-d1735d5afedd.png)

PR #1480, opened 2021-05-22
Fix: More than one NULL to separate commits. #1460

ylecuyer:
We use the `-z` flag within git log to separate commits with NULL chars instead of new lines

In this case for example there is no one but two NULL chars before the next commits thus failing with the previous code

### DIFF
--- a/source/git-parser.js
+++ b/source/git-parser.js
@@ -162,7 +162,7 @@ exports.parseGitLog = (data) => {
       parser = parseFileChanges;
       return;
     }
-    if (rows[index + 1] && (rows[index + 1].indexOf('\x00commit ') == 0 || rows[index + 1].indexOf('\x00\x00commit ') == 0)) {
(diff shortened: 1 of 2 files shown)


source/git-parser.js
```diff
@@ -162,7 +162,7 @@ exports.parseGitLog = (data) => {
       parser = parseFileChanges;
       return;
     }
-    if (rows[index + 1] && rows[index + 1].indexOf('\x00commit ') == 0) {
+    if (rows[index + 1] && rows[index + 1].match(/^\u0000+commit/) != null) {
```
campersau, line 165: Would `test` instead of `match` also work here?

ylecuyer: Indeed it will be better, I'll update the PR

campersau (approved):
Looks good thanks!

Just curious, what is #1560 referring to?

ylecuyer:
> Looks good thanks!
> 
> Just curious, what is #1560 referring to?

It is a mistake, it should read #1460 I have udpated the title :)

PR #1499, opened 2022-01-08
Add Tab Size Configuration

ZimGil:
This PR will allow the user to specify `tab-size` for the diff view.

Using tabs looks awful without this configuration (specially in `side-by-side` view) as it cause the text to be over indented as the default `tab-size` value is 8. 

Before (Tab size default to 8):
before

After (Tab size set to 2):
after

### DIFF
--- a/components/repository/repository.html
+++ b/components/repository/repository.html
@@ -1,4 +1,4 @@
-<div class="repository-view animated fadeInLeft" data-bind="attr: { style: 'tab-size: ' + ungit.config.diffTabSize }">
+<div class="repository-view animated fadeInLeft" data-bind="attr: { style: 'tab-size: ' + ungit.config.tabSize }">
   <!-- ko component: gitErrors --><!-- /ko -->
 
   <!-- ko if: isSubmodule -->
--- a/components/repository/repository.html
+++ b/components/repository/repository.html
@@ -1,4 +1,4 @@
-<div class="repository-view animated fadeInLeft" data-bind="style: { tabSize: ungit.config.diffTabSize }">
+<div class="repository-view animated fadeInLeft" data-bind="attr: { style: 'tab-size: ' + ungit.config.diffTabSize }">
   <!-- ko component: gitErrors --><!-- /ko -->
 
   <!-- ko if: isSubmodule -->
--- a/components/repository/repository.html
+++ b/components/repository/repository.html
@@ -1,4 +1,4 @@
-<div class="repository-view animated fadeInLeft">
+<div class="repository-view animated fadeInLeft" data-bind="style: { tabSize: ungit.config.diffTabSize }">
   <!-- ko component: gitErrors --><!-- /ko -->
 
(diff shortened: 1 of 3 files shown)


campersau:
Thanks! Since `tab-size` gets inherited it could also be set just once in e.g. the `repository.html`. In that case it would also affect commit messages. WDYT?

ZimGil:
Yeah Let's do it...
I'll push a change a bit later :)

ZimGil:
@campersau Would you say this should just go here?:
https://github.com/FredrikNoren/ungit/blob/628301f0de022b070efa3a574212c8f43656d351/components/repository/repository.html#L1

campersau:
Yes, I think that works fine.

source/config.js
```diff
@@ -299,6 +302,7 @@ const argv = yargs
     'numRefsToShow',
     'Number of refs to show on git commit bubbles to limit too many refs to appear.'
   )
+  .describe('diffTabSize', 'Specify tab size for diff')
```
ylecuyer, line 305:
```suggestion
  .describe('diffTabSize', 'Specify tab size for diff in css units')
```

I would just add it would be nice to say it is css units otherwise I'd have though it is tab size in spaces

ZimGil: Based on the [docs](https://www.w3schools.com/cssref/css3_pr_tab-size.asp) it actually work with number of space characters

ZimGil: Based on the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/tab-size) it actually work with number of space characters.

Length values are not supported in some browsers (Safari)

ylecuyer: oh good to know thanks,t hen you can discard my comment

ZimGil: We could mention that the value should be number of spaces to be clear

components/repository/repository.html
```diff
@@ -1,4 +1,4 @@
-<div class="repository-view animated fadeInLeft">
+<div class="repository-view animated fadeInLeft" data-bind="style: { tabSize: ungit.config.diffTabSize }">
```
campersau, line 1:
```suggestion
<div class="repository-view animated fadeInLeft" data-bind="attr: { style: 'tab-size: ' + ungit.config.diffTabSize }">
```
knockout has a "nice" feature to automatically append `px` to numeric values. But in this case we don't want it, so we need to use `attr` instead.
https://github.com/knockout/knockout/blob/45e86bf0347cf742ca359bab051ea82fdfae7a5d/src/binding/defaultBindings/style.js#L25-L27

ZimGil:
@campersau I believe it's ready now

components/repository/repository.html
ZimGil: Is there any plan or willing to move the UI to a more modern framework? (React / Vue.js)

campersau (approved):

PR #1509, opened 2022-03-17
Add `--` before potential user provided input to make it disambiguating

campersau:
> When writing a script that is expected to handle random user-input, it is a good practice to make it explicit which arguments are which by placing disambiguating -- at appropriate places.

https://git-scm.com/docs/gitcli/2.25.0#_description


campersau:
Closing in favor of #1510

PR #1527, opened 2022-07-16
Improvements from nodegit branch

wmertens:
I moved some generic fixes down on my branch and I think these can be considered for inclusion in master. I added the more contentious ones near the top so I can just move the HEAD :-)


wmertens:
FYI the tests fail because gravatar returns 404

PR #1553, opened 2023-01-14
Kill server when ungit gets killed

### LINKED ISSUE #1552: Server process keeps running when parent gets killed
When killing the parent process, aka `ungit`, the child process, `server.js`, keeps running. This constitutes an issue in the context of editor extensions for instance, where closing the editor does not kill ungit's server, so running ungit then keeps saying "ungit already running".

~~It seems this might be related to #1537. I feel like their pull request uses a bit of a hacky workaround to the arguments issue, which is as far as I know, more of a `yargs` issue.~~

~~If it fixes the issue, I'd suggest coming back to using `spawn` and find another workaround for the former issue instead, if possible.~~

Edit: after testing, it seems Hirse's PR is not the issue

victorbnl:
This makes sure the child process `server.js` is killed when the parent `ungit` gets killed

Fixes #1552

### DIFF
--- a/bin/ungit
+++ b/bin/ungit
@@ -16,8 +16,9 @@ const net = require('net');
 const server = net.createServer();
 let child;
 const cleanExit = () => {
-  if (child != undefined)
+  if (child) {
     child.kill('SIGINT');
+  }
   process.exit();
 };
 
--- a/bin/ungit
+++ b/bin/ungit
@@ -16,7 +16,8 @@ const net = require('net');
 const server = net.createServer();
 let child;
 const cleanExit = () => {
-  child.kill('SIGINT');
+  if (child != undefined)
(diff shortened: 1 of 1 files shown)


bin/ungit
```diff
@@ -14,7 +14,11 @@ const bugtracker = new BugTracker('launcher');
 // Fastest way to find out if a port is used or not/i.e. if ungit is running
 const net = require('net');
 const server = net.createServer();
-const cleanExit = () => process.exit();
+let child;
+const cleanExit = () => {
+  child.kill('SIGINT');
```
campersau, line 19: Please add a check if child is not undefined as it might be possible to kill the process before the child is spawned.

victorbnl: Right

PR #1564, opened 2023-07-28
Fallback for font if bundled font dont support glyphs

bes-internal:
Example of problem:
Browser select to render local font because bundled OpenSans.woff but from but from the wrong font style because it wasn't specified.
image

Fixed version:
image


campersau (approved):
Looks good. Thanks!

PR #1597, opened 2024-08-18
Better: change watcher to properly filter ignored directories

ylecuyer:
Lately I started to work on a node project with a lot of dependencies, when opening this project in ungit I started to get this error:

```
 2024-08-18T18:21:39.758Z - error: Error: ENOSPC: System limit for number of file watchers reached, watch '/home/ylecuyer/Projects/project/.git/index.lock'
```

So far I was able to get ungit running by changing the limit of inotify nodes doing

```
sudo sysctl fs.inotify.max_user_watches=524288 && sudo sysctl -p
```

And when checking the number of used inotify nodes, It is really big (263k):

```
ylecuyer@inwin:~/Projects/inotify-info$ ./_release/inotify-info 
------------------------------------------------------------------------------
INotify Limits:
  max_queued_events    16,384
  max_user_instances   128
  max_user_watches     524,288
------------------------------------------------------------------------------
       Pid Uid        App                     Watches  Instances
     14596 1000       node                    263,061          1
```

After investigating, I noticed that since node 20, native support for recursive has been added to fs.watch and the current watcher library used in ungit uses the native watcher when available (https://github.com/yuanchuan/node-watch/blob/main/lib/has-native-recursive.js#L6). The problem is that the filtering (to remove .gitignore files and node_modules files) is done only when the non native version is used.

I couldn't find a way to improve the current library to get it to work with native, so I just changed it for another one, there are a lot so I took one of them without benchmarking.

With the new watcher library the number of inotify nodes for the same project is now normal:

```
ylecuyer@inwin:~/Projects/inotify-info$ ./_release/inotify-info 
------------------------------------------------------------------------------
INotify Limits:
  max_queued_events    16,384
  max_user_instances   128
  max_user_watches     524,288
------------------------------------------------------------------------------
       Pid Uid        App                     Watches  Instances
     15510 1000       node                     13,494          1
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -32,7 +32,6 @@
         "mkdirp": "~3.0.1",
         "moment": "~2.30.1",
         "node-cache": "~5.1.2",
-        "node-watch": "~0.7.4",
         "nprogress": "~0.2.0",
         "open": "~10.1.0",
         "p-limit": "~5.0.0",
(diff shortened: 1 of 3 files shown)


campersau:
Looks good, lets try this one out, thanks!

